PR #2, opened 2018-04-21
Allow custom aliases

michaelmior:
This doesn't change the normal behaviour of the plugin at all.
It simply allows the user to override the default aliases with something else via environment variables.


wfxr:
Thanks for your contribution. I've tested it and it works perfectly!

PR #5, opened 2018-08-02
Bash support

idoa01:
Translates the zsh code to bash code. keeps the same behaviour.


wfxr:
Merged.
That would be great if you can maintain the bash part of the plugin. :)
Anyway, thanks for your contribution!

PR #7, opened 2018-08-07
zsh: fix glo not displaying output

sohkai:
Might also need to be done in the bash version?

I'm not aware of the rationale in using `eval`, but the [current `eval`ed line](https://github.com/wfxr/forgit/blob/master/forgit.plugin.zsh#L59) does not display any logs for me.

Changing it to also just be `git log --graph --color=always --format='%C(auto)%h%d %s %C(black)%C(bold)%cr' "$@" "$forgit_emojify"` also had issues with git (I've realiased it to `fglo`):

screen shot 2018-08-07 at 10 17 17 am

I do have `emojify` available in my path, and removing the `$forgit_emojify` from that line makes everything work :).

### DIFF
--- a/forgit.plugin.zsh
+++ b/forgit.plugin.zsh
@@ -56,7 +56,7 @@ forgit::inside_work_tree() {
 forgit::log() {
     forgit::inside_work_tree || return 1
     local cmd="echo {} |grep -o '[a-f0-9]\{7\}' |head -1 |xargs -I% git show --color=always % $forgit_emojify $forgit_fancy"
-    eval "git log --graph --color=always --format='%C(auto)%h%d %s %C(black)%C(bold)%cr' $@ $forgit_emojify" |
+    git log --graph --color=always --format='%C(auto)%h%d %s %C(black)%C(bold)%cr' "$@" |
         forgit::fzf +s +m --tiebreak=index \
             --bind="enter:execute($cmd |LESS='-R' less)" \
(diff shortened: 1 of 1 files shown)


wfxr:
Sorry but I can't reproduce the problem you described above.

Eval here is to evaluate variable `$forgit_emojify` first and then execute the whole string as a command. Otherwise `$forgit_emojify` will be regarded as a normal argument of `git log`

Equivalent command without eval:
```
git log --graph --color=always --format='%C(auto)%h%d %s %C(black)%C(bold)%cr' "$@" "|emojify" 
```

Equivalent command with eval:
```
git log --graph --color=always --format='%C(auto)%h%d %s %C(black)%C(bold)%cr' "$@" |emojify
```

They look similar but has different behavior.

I guess the `emojify` you installed do not support pipe. Try the following command and see if it works as expected:
```
echo :apple: :bug: :rocket: | emojify
```

sohkai:
Odd...

`echo :apple: :bug: :rocket: | emojify` -> works for me

`git log --graph --color=always --format='%C(auto)%h%d %s %C(black)%C(bold)%cr' |emojify` -> prints the log fine

`git log --graph --color=always --format='%C(auto)%h%d %s %C(black)%C(bold)%cr' "|emojify" ` -> the same as the error I was getting before

It looks like there's something odd with the way my eval is working on macOS?

PR #9, opened 2018-10-12
Add `git stash` viewer for bash

mars90226:
I've copied your `git stash` viewer implementation for zsh and make it works in bash.


wfxr:
Thants for your contribution!

PR #12, opened 2018-11-05
Support various OSs and tmux in flog ctrl-Y

doronbehar:


wfxr:
Perhaps you misunderstood what I meant.

What you only need is to add an command named `pbcopy`. For me on linux, I have a script in `PATH` with following lines:

``` bash
#!/usr/bin/env bash
xclip -selection clipboard "$@"
```

There should not be any problem using it with git alias.

If you want to make it support various OSs, put the multi-os clipboard code here then everywhere else can use it include `forgit`. You don't want to copy & paste the code again and again in every script, don't you?

Maybe not everyone like the hard-coded name `pbcopy`. I will add an option to make it configurable.

PR #14, opened 2019-01-17
Feature: use gitignore templates from local repo directly

laggardkernel:
The gitignore templates used here is from [dvcs/gitignore](https://github.com/dvcs/gitignore), which is
the same one used by [gitignore.io](https://www.gitignore.io/).

Environment variable `FORGIT_GI_USE_LOCAL` is used as a switch to
enable this feature, otherwise the original method is used by default.

`forgit::ignore::update` and `forgit::ignore::get` are rewritten
to download the repo, to get templates from the local repo.
Local variable `preview` in `forgit:ignore` is also rewritten to
get the preview from there.

All these implementations are put into an `if` conditional statement.
If you wanna accept this feature, you'd better reindent the files after merge.
Not only because I didn't indent the content for a better diff view
on purpose, but also the indentations in `forgit.plugin.sh`, the one for BASH,
are **mixed with using two spaces and four spaces**.

### DIFF
--- a/forgit.plugin.sh
+++ b/forgit.plugin.sh
@@ -127,6 +127,8 @@ __forgit_stash_show() {
 # git ignore generator
 export FORGIT_GI_CACHE=~/.gicache
 export FORGIT_GI_INDEX=${FORGIT_GI_CACHE}/.index
+
+if [[ -z "$FORGIT_GI_USE_LOCAL" ]]; then
 __forgit_ignore() {
     [ -f $FORGIT_GI_INDEX ] || __forgit_ignore_update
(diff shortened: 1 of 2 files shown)


laggardkernel:
The second commit is trying to make `forgit::ignore::get` work on case-sensitive filesystems. 

The first commit works with the single `gi` command, but not command like this `gi c`. Because the parameters passed to `forgit::ignore::get` by single `gi` command is revised by fzf. What you type in the fuzzy finder will be mapped to a word from `$FORGIT_GI_CACHE`, the words within which reflect the template filenames exactly.

While command like `gi c` passes the parameter `c` directly to `forgit::ignore::get`. `find -iname 'c.gitignore` is needed to match the real file `C.gitignore`.

BTW, I force-pushed the second commit once cause I made a typo in the log message. Sorry about that.

wfxr:
Good idea! Will be merged later.
And I found it painful to maintain both versions at the same time.
I am doing some work on it to share the code between zsh & bash.

PR #22, opened 2019-05-28
Enable configuring preview command for every cmd

### LINKED ISSUE #21: Support for options to pass to `git diff`
Hello and thank you for `forgit`. I think I am starting to like it.

But I didn't find `diff-so-fancy` (which you suggest in the README) to my particular liking so I removed it.

My `git df` alias uses `-w` to suppress white space changes to increase the content-wise SNR, and `--word-diff=color` to avoid excess lines in my diffs. I would like `ga` to show the same. Likewise others may have other diff preferences.

`ga` needn't do anything specific to support each option, but if you provide some mechanism to pass our desired options to `git diff`, such as from an environment variable `FORGIT_DIFF_OPTIONS` (similar to existing `FORGIT_FZF_DEFAULT_OPTS`) then it would be useful.

Thanks!

doronbehar:
May fix #21 as well.

I noticed this was especially needed in `forgit::add` because it's default preview command shows nothing for completely untracked files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,17 @@ FORGIT_FZF_DEFAULT_OPTS="
 --height '80%'
 "
 ```
+
+You can also customize options fzf will use for each command individually:
+
+- `ga`: `FORGIT_ADD_FZF_OPTS`
+- `glo`: `FORGIT_LOG_FZF_OPTS`
+- `gi`: `FORGIT_IGNORE_FZF_OPTS`
+- `gd`: `FORGIT_DIFF_FZF_OPTS`
+- `gcf`: `FORGIT_CHECKOUT_FZF_OPTS`
+- `gss`: `FORGIT_STASH_FZF_OPTS`
+- `gclean`: `FORGIT_CLEAN_FZF_OPTS`
+
 ## Optional
 
 - [`diff-so-fancy`](https://github.com/so-fancy/diff-so-fancy): Improve the `git diff` output.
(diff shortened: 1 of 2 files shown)


wfxr:
@doronbehar Thanks for your patch!
Inspired by your PR, how about exposing `FORGIT_ADD_FZF_OPTS`, `FORGIT_DIFF_FZF_OPTS`, `FORGIT_LOG_FZF_OPTS`, etc directly to obtain the maximum customization capability (eg, custom keybind on specific command)?

```bash
# git commit viewer
forgit::log() {
    local cmd="echo {} |grep -o '[a-f0-9]\{7\}' |head -1 |xargs -I% git show --color=always % $* $forgit_emojify $forgit_fancy"
    eval "git log --graph --color=always --format='%C(auto)%h%d %s %C(black)%C(bold)%cr' $* $forgit_emojify" |
        forgit::fzf +s +m --tiebreak=index \
            --bind="enter:execute($cmd |LESS='-R' less)" \
            --bind="ctrl-y:execute-silent(echo {} |grep -o '[a-f0-9]\{7\}' |${FORGIT_COPY_CMD:-pbcopy})" \
            --preview="$cmd" \
            "$FORGIT_LOG_FZF_OPTS"
}
```


One can configure them like this:

```bash
FORGIT_LOG_FZF_OPTS='
--preview=whatever
--bind=whatever
'
```

doronbehar:
Thankfully, `fzf` uses only the last `--preview` so this is a much better idea. I'm updating the PR.

wfxr:
Thanks for you greet work!

PR #24, opened 2019-07-05
Add `grh` command for `git reset HEAD` (unstage)

attomos:
I added `grh` (git reset HEAD) command to unstage the staged files. This is the opposite of `git add`.

Could you please review and add screenshot to README?

Hope you find this PR useful 🙂


wfxr:
@attomos Sorry for the late reply. I like this enhancement and tested it works well.
Thanks for your contribution!

Issue #37, opened 2019-10-02
Feature request: add icdiff support


stephencheng:
First, this script is so cool.

It would be great it can optionally support icdiff as a side by side viewer.

Essentially the code change would be some thing like: 
   # cmd="git diff --color=always $commit -- {} $forgit_fancy"
to
  # cmd="git icdiff --color=always $commit -- {} 

Thanks

wfxr:
@stephencheng 
Thanks for your suggestion but `icdiff` is not the standard sub command for git. Implement a function which is not in the scope of the standard git is not the project's goal.

If your prefer `icdiff` you can override the `forgit::diff` in your bashrc/zshrc.

PR #38, opened 2019-10-02
add icdiff support and fix FORGIT_FZF_DEFAULT_OPTS being populated multiple times

stephencheng:
Creating this for: https://github.com/wfxr/forgit/issues/37

Two things:

1. Not sure if it's a good thing to duplicated the code but think it's simpler to manage the switches. The icdiff view isn't rendered well in the not maximised preview window, hence hidden

2. Fir the bug that the FORGIT_FZF_DEFAULT_OPTS are populated a few times when my dotfile is loaded multiple times, maybe you intend to do it so.

### DIFF
--- a/forgit.plugin.zsh
+++ b/forgit.plugin.zsh
@@ -66,7 +66,6 @@ forgit::icdiff() {
         $FORGIT_DIFF_FZF_OPTS
     "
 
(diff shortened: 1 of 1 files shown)


wfxr:
See #37

PR #46, opened 2019-11-05
Fixed readme and added fzf check to zsh to be the same as fish version

cjappl:
As promised, here is the README update, feel free to check my fork for how it looks now, basically added the fish sections!

I also added a check for fzf as well as a helpful warning message if it's missing. This is something I did for the fish version and figured it'd be nice if they matched!

### DIFF
--- a/forgit.plugin.zsh
+++ b/forgit.plugin.zsh
@@ -1,4 +1,5 @@
 # MIT (c) Wenxuan Zhang
+hash fzf > /dev/null 2>&1 || {echo "FZF not found and is requried for forgit" && return 1}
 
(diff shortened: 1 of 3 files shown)


forgit.plugin.zsh
```diff
@@ -1,4 +1,5 @@
 # MIT (c) Wenxuan Zhang
+hash fzf > /dev/null 2>&1 || {echo "FZF not found and is requried for forgit" && return 1}
```
wfxr: Is it better to put this line below the 6th line? Then we can use `forgit::warn` to print the warning message:

``` bash
hash fzf &>/dev/null || {forgit::warn "FZF not found and is requried for forgit" && return 1}
```

wfxr:
LGTM.

PR #47, opened 2019-11-06
fix shell error in forgit.plugin.zsh

laur89:
caused by https://github.com/wfxr/forgit/commit/325e98cd03c71d7e280b7259ff7793cec0afd726#diff-b3f00b3beaf9e81cc8bee82da7fc86faR6


cjappl:
I'm not sure I see the error here, when I run it unchanged:

```
CJAPPL-MBP% source forgit.plugin.zsh
[Warn] FZF not found and is requried for forgit
CJAPPL-MBP% echo $?
1
```

Can you describe what error you're seeing? These commands look equivalent

laur89:
Did you try it in bash?

cjappl:
Ah! Yes, definitely see it! Thanks for the clarity. This LGTM, I'll merge it now because it's core to the plugin actually working @wfxr

cjappl:
Thanks for the help @laur89, my inexperience with bash is showing ;)

cjappl:
Tested and confirmed this fix on my machine with both zsh and sh

wfxr:
@cjappl @laur89 Thanks. :lgtm:

PR #53, opened 2019-11-18
Add option to Interact with selection

AhmedAbdulrahman:
I would like to thank you for this awesome plugin :). 

I have extended stash functionality to add options to interact with selections using `pop`, `apply` or even`checkout` to selection. 

bindings:
- `CTRL + b` checks out the stash as a branch and removes the stash `
- `CTRL + y` will apply the stash`
- `CTRL + x` will drop the stash`

I would appreciate if you reivew it

Kapture 2019-11-18 at 21 45 19

### DIFF
--- a/forgit.plugin.zsh
+++ b/forgit.plugin.zsh
@@ -105,14 +105,47 @@ forgit::restore() {
 # git stash viewer
 forgit::stash::show() {
     forgit::inside_work_tree || return 1
-    local cmd opts
-    cmd="git stash show \$(echo {}| cut -d: -f1) --color=always --ext-diff | $forgit_pager"
+    local cmd opts out
+    cmd="git stash show --patience --stat --pretty=oneline -p \$(echo {} | cut -d: -f1) --color=always --ext-diff $forgit_fancy"
     opts="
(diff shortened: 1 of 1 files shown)


wfxr:
Hi @AhmedAbdulrahman Thanks for your contribution!

This feature looks great. But it doesn't seem to work properly under bash:

`bash: ${(f)out}: bad substitution`

Since the `forgit.plugin.sh` is a soft link to `forgit.plugin.zsh` we should ensure the compatibility.

wfxr:
The latest `fzf` add support for reloading input without exiting: https://github.com/junegunn/fzf/issues/1750

Now to implement this feature you only need to add a bind to `FORGIT_STASH_FZF_OPTS` like this:
```
FORGIT_STASH_FZF_OPTS='
--bind="ctrl-x:reload(git stash drop $(cut -d: -f1 <<<{}) 1>/dev/null && git stash list)"
'
```

Considering the existing interfaces already support adding keybind I decided to put it as an example in the readme instead of adding a lot of new commands.

https://github.com/wfxr/forgit/blob/master/README.md#152

Issue #39, opened 2019-10-23
xargs: realpath: No such file or directory


skylarmb:
Just installed forgit using zplug. When running `gd` i get the following error. How can I debug this? other forgit commands seem to work fine

```
xargs: realpath: No such file or directory
```

wfxr:
It seems the `realpath` command does not exist in your system. What's your OS ?

wfxr:
@skylarmb `gd` use `realpath` to convert absolute path to relative path which is shorter and easier to browse. `realpath` is a gnu standard command but some os like mac does not install it by default.
I will add some code to avoid this convert when `realpath` was not installed. But if you want the relative paths you can install the `realpath` command. (most gnu commands include `realpath` exists in homebrew package named `coreutils`)

skylarmb:
👏 thanks!!

PR #58, opened 2019-11-22
Fix path related bugs.

wfxr:
# Description

Fix some path related bugs. Add status indicator to `forgit::diff` command BTW.

## Type of change

Please delete options that are not relevant.

- [x] Bug fix
- [x] (Minor) Feature : Add status indicator to `forgit::diff` command BTW.
- [x] (Minor) Breaking change : Paths listed by forgit::diff will now use repo's root as the start point.

## Shells this fix should effect: 
- [x] bash
- [x] zsh 
- [ ] fish 

## Shells this fix has been tested in: 
- [x] bash
- [x] zsh 
- [ ] fish 

**Test Configuration**:
* OS: Arch Linux x86_64

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas

# Related issue(s):

#39 : This pr will remove the dependency on `realpath`.

Please help sync the fish version. @cjappl

### DIFF
--- a/forgit.plugin.zsh
+++ b/forgit.plugin.zsh
@@ -37,14 +37,14 @@ forgit::diff() {
     }
 
     repo="$(git rev-parse --show-toplevel)"
-    target="\$(echo {} | sed 's/.*]   //')"
+    target="\$(echo {} | sed 's/.*]  //')"
     cmd="git diff --color=always $commit -- $repo/$target | $forgit_pager"
     opts="
         $FORGIT_FZF_DEFAULT_OPTS
         +m -0 --preview=\"$cmd\" --bind=\"enter:execute($cmd |LESS='-R' less)\"
         $FORGIT_DIFF_FZF_OPTS
     "
-    eval "git diff --name-status $commit -- ${files[*]} | awk '{printf \"[%2s]  \", \$1; \$1=\"\"; print \$0}'" |
+    eval "git diff --name-status $commit -- ${files[*]} | sed 's/^\(.\)[[:space:]]\+\(.*\)$/[\1]  \2/'" |
         FZF_DEFAULT_OPTS="$opts" fzf
 }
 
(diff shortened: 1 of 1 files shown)


cjappl:
Cool! I see this. I will get a commit in as soon as I can. I am on vacation so it may be a few days :)

wfxr:
@cjappl No rush. Take your time :)

forgit.plugin.zsh
```diff
@@ -63,7 +64,7 @@ forgit::add() {
     "
     files=$(git -c color.status=always -c status.relativePaths=true status --short |
         grep -F -e "$changed" -e "$unmerged" -e "$untracked" |
-        awk '{printf "[%10s]  ", $1; $1=""; print $0}' |
+        sed 's/^\(..[^[:space:]]*\)[[:space:]]\+\(.*\)$/[\1]  \2/' |
```
cjappl: Hi @wfxr, working on this now, but want to be 100% sure I'm understanding what this is before I put it in!

Can you explain what this sed command, and the sed command up top do? I'm not super well versed on sed

wfxr: Here regular expressions are used for capturing and replacing.

If you remove the escape character it will look a bit clearer:
`sed 's/^(..[^[:space:]]*)[[:space:]]+(.*)$/[\1]  \2/'`. 

What this line does is to capture the status indicator at the beginning of the line, surround it with square brackets, then append the full path. Completely explaining sed's regex rules is not easy. But you can test it to see how it works:

``` bash
# go to `forgit`'s local repo
touch 'a          b   c'
echo 'hello world!' >> README.md

# buggy
git status --short | awk '{printf "[%10s]  ", $1; $1=""; print $0}'

# fixed
git status --short | sed 's/^\(..[^[:space:]]*\)[[:space:]]\+\(.*\)$/[\1]  \2/'
```

PR #62, opened 2019-11-26
Fix path fish

cjappl:
## Check list

## Description

Getting up to parity with zsh version #58 

Also fixed a bunch of issues the fish side of things had with whitespace! so glad I checked this out, I was not handling things correctly.

## Type of change

- [x] Bug fix
- [x] New feature
- [x] Refactor

## Test environment

- Shell
    - [ ] bash
    - [ ] zsh 
    - [x] fish 
- OS
    - [ ] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/forgit.plugin.fish
+++ b/forgit.plugin.fish
@@ -16,11 +16,11 @@ end
 set core_pager (git config core.pager)
 
 if test -n $core_pager
-    set forgit_pager (echo "| $core_pager")
+    set forgit_pager "$core_pager"
 else if type -q diff-so-fancy >/dev/null 2>&1
-    set forgit_pager "| diff-so-fancy | less --tabs 4 -RFX"
+    set forgit_pager "diff-so-fancy | less --tabs 4 -RFX"
 else 
-    set forgit_pager "| cat"
+    set forgit_pager "cat"
 end
 
 # https://github.com/wfxr/emoji-cli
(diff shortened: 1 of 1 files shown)


cjappl:
@wfxr  tagging you to take a look, but merging and closing. Thanks for the example!

In the future, if it's easy, providing an example like that is super helpful to make sure I'm implementing the right thing! That would be very appreciated :) Your example helped me catch a bunch of bugs in my original implementation I didn't realize.

Cheers!

wfxr:
@cjappl OK. I will try to add more descriptions when I change the code. I notice that you fixed more bugs about white space. I will sync later.

cjappl:
That would be great :)

Yeah I noticed there were a few more things going on.

Maybe at some point it would make sense to set up a "test repo" and a few
scripts to get files in certain states so we can make sure everything is
working right! May be good as a kind of "unit test"

wfxr:
@cjappl Agree with you. But I don't have any good ideas on doing unit testing based on fzf. Maybe we have to use fzf's sdk to do this, just like fzf's own tests. That would be more troublesome. I will try it when I have enough time.

cjappl:
Yeah, it's hard to say what the best way would be. I'll also investigate
FZF's tests and see what they have!

PR #70, opened 2020-01-13
Add files as arguments

carlosedp:
## Check list

## Description

Add case where "ga" command can accept files as arguments
that will be passed directly to "git add" bypassing fzf.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [x] zsh 
    - [ ] fish 
- OS
    - [x] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/forgit.plugin.zsh
+++ b/forgit.plugin.zsh
@@ -52,6 +52,12 @@ forgit::diff() {
 forgit::add() {
     forgit::inside_work_tree || return 1
     local changed unmerged untracked files opts
+    # Add files if passed as arguments
+    [[ $# -ne 0 ]] && {
+        git add $@
+        return
+    }
+
     changed=$(git config --get-color color.status.changed red)
     unmerged=$(git config --get-color color.status.unmerged red)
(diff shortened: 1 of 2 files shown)


forgit.plugin.zsh
```diff
@@ -52,6 +52,12 @@ forgit::diff() {
 forgit::add() {
     forgit::inside_work_tree || return 1
     local changed unmerged untracked files opts
+    # Add files if passed as arguments
+    [[ $# -ne 0 ]] && {
+        eval "git add $@"
```
wfxr: Hi @carlosedp I don't see why `eval` is needed here, can you explain it? Or we just use `git add "$@" instead?

carlosedp: Sure, I just wanted to keep consistent with other functions. Will push the change.

wfxr: LGTM. Thanks for your contribution!

PR #73, opened 2020-02-22
Diff staged

hoshsadiq:
## Check list

## Description
This adds a new alias `gds` that acts the same as `gd` but it shows the staged changes. The alias `gd` has an argument `$1` which can be `commit`, this has been removed since this is invalid when reviewing staged commits.

In addition, the `eval` command has been removed for the function `forgit::diff`, and isn't used in `forgit::diff::staged`. This should make forgit that bit safer.

Note that I do not have fish or do I plan on installing it, so while, I did my best to provide the same functionality for fish, this has unfortunately not been tested. I'm happy to back out this change if necessary.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [x] bash
    - [x] zsh 
    - [ ] fish 
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,7 @@ You can change the default aliases by defining these variables below.
 # Define them before sourcing the plugin if you don't use any plugin manager.
 forgit_log=glo
 forgit_diff=gd
+forgit_diff_staged=gds
 forgit_add=ga
 forgit_reset_head=grh
 forgit_ignore=gi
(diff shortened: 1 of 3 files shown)


wfxr:
Hi @hoshsadiq Thanks for your contribution!
If you want to show diff between index and latest commit you can use `gd <current_branch>` or more generally `gd HEAD`.

hoshsadiq:
Didn't realise, but after trying it out, it seems `gd HEAD` shows the unstaged changes when nothing is staged, which in my opinion can cause confusion.

wfxr:
How about `gd --staged` then? I don't want to add duplicate code for a less-used feature. And `gd <branch>`, `gd HEAD`, or` gd --staged` basically does what you want (even if they are not the exactly same, 90% of the cases are covered). I think it is no necessary to reimplement a `gds` function using basically the same logic.

PR #89, opened 2020-05-16
Make gss accepts git stash list options

attomos:
## Check list

## Description
This change allows `gss` to accept `git stash list` options.

For example:
```bash
$ gss --date=human
$ gss --date=local
$ gss --date=relative
$ gss --date=short
```
This can be useful if you want to filter stashes by their creation time.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [x] zsh 
    - [ ] fish 
- OS
    - [ ] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Others:


forgit.plugin.zsh
```diff
@@ -128,7 +128,12 @@ forgit::stash::show() {
         +s +m -0 --tiebreak=index --bind=\"enter:execute($cmd | LESS='-R' less)\"
         $FORGIT_STASH_FZF_OPTS
     "
-    git stash list | FZF_DEFAULT_OPTS="$opts" fzf --preview="$cmd"
+    if [ "$#" -eq 0 ]; then
+      git stash list | FZF_DEFAULT_OPTS="$opts" fzf --preview="$cmd"
+    else
+      # Prepend git stash list output with stash@{0}:..stash@{N-1}: to make the `cut` above works.
+      git stash list $* | awk '{print "stash@{" NR-1 "}:" $0}' | FZF_DEFAULT_OPTS="$opts" fzf --preview="$cmd"
+    fi
```
wfxr, line 136: Thanks for your work and the good idea to improve `gss` @attomos.
 
Support all options may easily break the output format pattern we depends to show preview.

Since other options may not very useful for `gss` we can improve the output format by pre-set the `--format` option.
```suggestion
    git stash list --color=always --format='%C(cyan)%gd%C(reset): %s %C(black bold)%cr%C(reset)' |
        FZF_DEFAULT_OPTS="$opts" fzf --preview="$cmd"
```

In this way we don't need to manually prepend stash ids. It shoud be more robust. Relative to flexibility, I prefer stability. How do you think about it? @attomos

wfxr: Or can we have a better way to get the stash id instead of manually calculating which I think is not very stable? So that can safely support any options for `stash list`.

attomos: Thanks for the feedback!

I will let you know if I find a way to make it more robust.

Issue #75, opened 2020-03-04
ga - patch mode


loonies:
## Check list

- [x] I have read through the [README](https://github.com/wfxr/forgit/blob/master/README.md)
- [x] I have the latest version of forgit
- [x] I have searched through the existing issues

## Environment info

Irrelevant

## Feature

Add a new command, or extand `ga` for patch mode (`git add --patch`).

wfxr:
I think to completely solve the similar problems we shoud support pasing arguments to the final `git add` command. But there are some troubble because `ga` already take it's arguments as filters to only list files start with those paths.

PR #90, opened 2020-05-17
Allow files as arguments to `forgit:reset:head`

PSalant726:
## Check list

## Description
Adds a case where `grh` (`forgit:reset:head`) can accept files as arguments that will be passed directly to `git reset -q HEAD ...`, bypassing `fzf`.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [x] zsh 
    - [ ] fish 
- OS
    - [ ] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/forgit.plugin.fish
+++ b/forgit.plugin.fish
@@ -127,6 +127,9 @@ end
 ## git reset HEAD (unstage) selector
 function forgit::reset::head
     forgit::inside_work_tree || return 1
+    # Reset HEAD of specific files, if passed as arguments
+    count $argv >/dev/null && git reset -q HEAD "$argv" && git status --short && return
+
     set cmd "git diff --cached --color=always -- {} | $forgit_pager"
     set opts "
         $FORGIT_FZF_DEFAULT_OPTS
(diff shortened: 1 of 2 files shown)


wfxr:
Thank you @PSalant726 . I don't want to add the additional logic to bypass `fzf` since it's `forgit`'s core function and I don't think this is a very common or necessary operation. 

You can implement it in a non-invasive way. One way is to add a wrap function in your `rc` file such as `forgit :: reset :: head :: wrap`. In the function you can bypass the call to `forgit :: reset: head` according to your needs. Then just alias ` gss` to `forgit :: reset :: head :: wrap`.

Anyway thanks for your contribution!

PSalant726:
Seemed very similar to https://github.com/wfxr/forgit/pull/70 ¯\_(ツ)_/¯ 

Anyway thanks for the feedback, and the hard work on this plugin.

wfxr:
@PSalant726 Yes it's similar to #70 . But its solution conflicts with #75 which is still open :(

That's also why I am more cautious when facing these non-essential features now.

PR #95, opened 2020-06-13
fixed gcf, and a bunch of other things. Added git status before returns

### LINKED ISSUE #94: FISH: gcf doesn't respect tab to multi select, will only checkout first file
## Check list

## Environment info

- OS
    - [x] Mac OS X

- Shell
    - [ ] bash
    - [ ] zsh
    - [x] fish

When using `gcf` on fish with tab to multi-select, it will only select the first file, not multiple.

Multi select is working in `ga`, so just need to investigate why it's not working with the other

@wfxr when you get a chance, would you double check zsh and make sure this problem isn't there too?

cjappl:
fixes #94 

## Check list

## Type of change

- [x] Bug fix
- [x] New feature
- [x] Refactor

## Test environment

- Shell
    - [x] bash
    - [x] zsh 
    - [x] fish 
- OS
    - [x] Mac OS X

### DIFF
--- a/forgit.plugin.fish
+++ b/forgit.plugin.fish
@@ -83,7 +83,7 @@ end
 function forgit::add
     forgit::inside_work_tree || return 1
     # Add files if passed as arguments
-    count $argv >/dev/null && git add "$argv" && return
+    count $argv >/dev/null && git add "$argv" && git status --short && return
 
     set changed (git config --get-color color.status.changed red)
     set unmerged (git config --get-color color.status.unmerged red)
@@ -119,6 +119,7 @@ function forgit::add
         for file in $files
             echo $file | tr '\n' '\0' | xargs -I{} -0 git add {}
         end
+        git status --short
         return
     end
     echo 'Nothing to add.'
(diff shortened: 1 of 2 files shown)


forgit.plugin.fish
```diff
@@ -136,14 +137,16 @@ function forgit::reset::head
...
         end
+        git status --short
+        return
     end
     echo 'Nothing to unstage.'
 end
 
 # git checkout-restore selector
-function forgit::restore
+function forgit::checkout_file
```
cjappl, line 149: @wfxr How would you feel about changing all the "restore" variables to "checkout_file" just to match the alias?

for example, these functions would be renamed, but then we would also rename $forgit_restore, and change the comment to say "nothing to checkout". 

I think it may be a bit clearer as to what's going on!

forgit.plugin.zsh
```diff
@@ -142,7 +142,7 @@ forgit::clean() {
     "
     # Note: Postfix '/' in directory path should be removed. Otherwise the directory itself will not be removed.
     files=$(git clean -xdfn "$@"| sed 's/^Would remove //' | FZF_DEFAULT_OPTS="$opts" fzf |sed 's#/$##')
-    [[ -n "$files" ]] && echo "$files" | tr '\n' '\0' | xargs -0 -I% git clean -xdf '%' && return
+    [[ -n "$files" ]] && echo "$files" | tr '\n' '\0' | xargs -0 -I% git clean -xdf '%' && git status --short && return
```
cjappl, line 145: just added these to have the same functionality in all our calls! basically the idea is:

"If forgit has changed your git tree, always put a git status --short before returning"

cjappl:
@wfxr I'd like your feedback on my comments before I merge, let me know what you think!

wfxr:
@cjappl These improvements look good. I am glad to merge them into code base.

PR #105, opened 2020-08-20
Configurable pager

jilleJr:
## Check list

## Description

I have a use case where I need to override the pager config, so I thought I'd write a PR on the case.

My use case is that I'm using WSL 2 where they still have super slow file access across the distros. To conquer that I'm using a small script in my WSL instead of git that goes something like this:

```bash
#! /bin/bash

LINUX_GIT='/usr/bin/git'
WINDOWS_GIT='git.exe'

WORKING_DIR="$(pwd)"

if [[ "${WORKING_DIR:0:3}" == '/c/' ]]; then
	$WINDOWS_GIT "$@"
else
	$LINUX_GIT "$@"
fi
```

Using the interopability to my advantage to use either git inside my WSL or inside my Windows host machine depending on which repo I'm working on. Making it blazing fast in comparison to only using the WSL git.

Problem then arises with the pager. Since it's pointing to my `git.exe` when targeting repos on my Windows host machine then it fetches the paging config from my host machine, which in my case depends on different programs and different syntax (ex WSL core.pager: `diff-so-fancy`, ex Windows core.pager: `diff-so-fancy.cmd`).

Instead of having to do some weird workarounds of a `core.pager` config that works on both windows and linux, I would find it being configurable to be much easier to use. With these changes, whenever I'm using the `forgit::log` function from my WSL I can override it to use a WSL based pager and ignore all my windows options.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [x] zsh 
    - [ ] fish 
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/forgit.plugin.zsh
+++ b/forgit.plugin.zsh
@@ -59,7 +59,7 @@ forgit::diff() {
 forgit::add() {
     forgit::inside_work_tree || return 1
     # Add files if passed as arguments
-    [[ $# -ne 0 ]] && git add "$@" && $git status -su && return
+    [[ $# -ne 0 ]] && git add "$@" && git status -su && return
 
     local changed unmerged untracked files opts preview extract
     changed=$(git config --get-color color.status.changed red)
@@ -88,7 +88,7 @@ forgit::add() {
         sed -E 's/^(..[^[:space:]]*)[[:space:]]+(.*)$/[\1]  \2/' |
         FZF_DEFAULT_OPTS="$opts" fzf --preview="$preview" |
         sh -c "$extract")
-    [[ -n "$files" ]] && echo "$files"| tr '\n' '\0' |xargs -0 -I% git add % && $git status -su && return
+    [[ -n "$files" ]] && echo "$files"| tr '\n' '\0' |xargs -0 -I% git add % && git status -su && return
     echo 'Nothing to add.'
 }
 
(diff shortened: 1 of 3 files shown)


forgit.plugin.zsh
```diff
@@ -38,20 +35,20 @@ forgit::diff() {
     forgit::inside_work_tree || return 1
     local cmd files opts commit repo
     [[ $# -ne 0 ]] && {
-        if $forgit_git_no_pager rev-parse "$1" -- &>/dev/null ; then
+        if git --no-pager rev-parse "$1" -- &>/dev/null ; then
```
wfxr, line 38: does it necessary to add `--no-pager` here?

jilleJr: Yea no actually. It's my inexperience with tty/pty showing through. Thanks for that quora link, great info!

I've redacted my `--no-pager` changes.

jilleJr:
I've only manually tested this in zsh, but not in fish nor bash explicitly. I haven't looked at your CI pipeline but I hope they ensure it's working as expected 😅 Haven't coded with fish before now

wfxr:
@jilleJr I tested both `bash` and `fish` work fine. Thanks for your work!

PR #107, opened 2020-09-02
Add missing parameters to fix gclean

jsfaint:
Without the parameters some file will not be cleaned

## Check list

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [x] bash
    - [x] zsh 
    - [ ] fish 
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/forgit.plugin.zsh
+++ b/forgit.plugin.zsh
@@ -142,8 +142,8 @@ forgit::clean() {
         $FORGIT_CLEAN_FZF_OPTS
     "
     # Note: Postfix '/' in directory path should be removed. Otherwise the directory itself will not be removed.
-    files=$(git clean -xdfn "$@"| sed 's/^Would remove //' | FZF_DEFAULT_OPTS="$opts" fzf |sed 's#/$##')
-    [[ -n "$files" ]] && echo "$files" | tr '\n' '\0' | xargs -0 -I% git clean -xdf '%' && git status --short && return
+    files=$(git clean -xdffn "$@"| sed 's/^Would remove //' | FZF_DEFAULT_OPTS="$opts" fzf |sed 's#/$##')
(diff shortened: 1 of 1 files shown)


wfxr:
@jsfaint Thank you but I don't fully understand what you mean. Can you show a minimal example to describe what the problem this pr try to solve?

jsfaint:
I have a repo which will generate some temporary files.
When I run `gclean` in that repo, I will get this message
image

Then I use `gclean -f`, that folder will not be skipped in the list
image

But parameter `-f` is not passed to `git clean -xdf`, so the folder is still skipped
image

jsfaint:
Reproduce step as below:
1. `git init a`
2. `git init a/b/b`
3. `cd a`
4. `gclean`
5. `gclean -f`

wfxr:
Clear explanation! I have not considered the case of repo nesting. Arguments for `gclean` is used to filter the files, eg: `gclean targets`, passing them all to `git clean` would make the filter not work. Before there is a good and simple way to separate the parameters (like `-f`) and files (like `targets`) I would prefer appending `-f` option by default:
```bash
    files=$(git clean -xdffn "$@"| sed 's/^Would remove //' | FZF_DEFAULT_OPTS="$opts" fzf |sed 's#/$##')
    [[ -n "$files" ]] && echo "$files" | tr '\n' '\0' | xargs -0 -I% git clean -xdff '%' && git status --short && return
```
How do you think about it @jsfaint ?

jsfaint:
@wfxr Got it! I'm misusing this feature.
Appending `-f` option by default is good choice to me.

jsfaint:
I updated the PR, add `-f` for both filtering and removing

wfxr:
Merged. Thank you!

jsfaint:
@wfxr Thanks for your time :)

PR #114, opened 2020-09-10
fixed fish if .forgit wasn't already existing

mirsella:
## Check list

## Description

when using forgit for the first time, the .forgit folder is not
created. forgit::ignore::update, which git glone the repo, was executed
only if the folder existed, so now it execute only if the .forgit don't
exist.

also, there some weird things in the logic with forgit::ignore::update,
some part of the code it not running in any case, since the if test -d
in the fonction is called in a fonction which is called only if the
folder don't exist, or i miss something ?

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [ ] zsh 
    - [x] fish 
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/forgit.plugin.fish
+++ b/forgit.plugin.fish
@@ -235,7 +235,7 @@ if test -z "$FORGIT_GI_TEMPLATES"
 end
 
 function forgit::ignore
-    if test -d "$FORGIT_GI_REPO_LOCAL"
+    if ! test -d "$FORGIT_GI_REPO_LOCAL"
         forgit::ignore::update
     end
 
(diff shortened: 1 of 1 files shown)


cjappl:
Nice find! definitely a translation issue when I first created forgit for fish. Thanks for catching this :) 

Changed "!" to "not", as that's better fish syntax, but this was definitely a bug.

Issue #116, opened 2020-10-09
Compatible with a fish plugin manager


FollieHiyuki:
## Check list

- [x] I have read through the [README](https://github.com/wfxr/forgit/blob/master/README.md)
- [x] I have the latest version of forgit
- [x] I have searched through the existing issues

## Environment info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:
- Shell
    - [ ] bash
    - [ ] zsh
    - [x] fish

## Problem / Steps to reproduce

Hi. Thanks for a nice shell plugin.
I have some suggestions: maybe moving forgit.plugin.fish into a directory named conf.d so that it can be easily installed,updated and autoloaded with a plugin manager like fisher or oh my fish. And the file doesn't need to be executable.

cjappl:
Hi @AoHiyuki. This is a great idea.

Unfortunately I don't have a lot of experience using or developing fish plugins for these systems. Do you have any experience? If so, please feel free to do a PR on this!

Otherwise it may be a little bit while I figure out how to make it work :)

AoHiyuki:
Hi @cjappl. I made it working with Fisher. Fairly simple as I described.
As I'm not familiar with Oh My Fish, it may take a while for me to figure out.
1 question though: do you mind if I separate the plugin into autoload and functions, since issue #110 ?

wfxr:
@AoHiyuki The main reason I object to splitting `forgit.plugin.zsh` is that it makes it difficult for zsh and bash to share code, or you have to do a lot of dirty work. This however does not make the plugin easier to use or maintain.

But for the fish version, it is not compatible with zsh or bash from the beginning (As you can see, it is a translation of the bash/zsh version, not a symbol link). Therefore, I am not opposed to splitting it. @cjappl is responsible for maintaining the fish version, so you can freely discuss solutions with him. @AoHiyuki

wfxr:
`fisher` support has been implemented. I think this issue can be closed now @cjappl

cjappl:
Sounds good.

Thanks for the good work @wfxr  and @AoHiyuki

AoHiyuki:
> I just caught a minor issue. Didn't check things out thoroughly :(
> forgit.plugin.fish appears in 2 places (even if it is just a symlink). So when installed with Fisher it will be copied to both conf.d and functions (the file in conf.d to ~/.config/fish/conf.d and the main file to ~/.config/fish/functions)
[...]

PR #117, opened 2020-10-10
Attempt to make this a fish plugin

AoHiyuki:
## Check list

## Description

Related to issue #116.
I've only moved things around. It works flawlessly with Fisher but has some problems with OhMyFish. Local variables like ```set forgit_pager "$FORGIT_PAGER"``` aren't sourced by OhMyFish.
We can work around this by making them global, or create a symlink in install hook. I will address this in OhMyFish, since I don't know whether it is a bug.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [ ] zsh 
    - [x] fish 
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:


cjappl:
Looking good so far. Definitely willing to give this a shot!

Will you please update the CI job to run the new "main" file instead of the old one? That is why the checks are failing.

That file is ".github/workflows/ci.yaml", change the line that looks like this:

` - name: Test fish
      run: fish forgit.plugin.fish`

to run the file in conf.d and make sure the checks are passing, then I'll merge

cjappl:
and it's great to start with one, we can worry about OhMyFish next, this is a valuable step for now :)

cjappl:
Oh, one more request. Please update the readme so that new users can "try out" the plugin without downloading the source. That line just needs a file location update. Also, please add a section for how to install this plugin using Fisher

cjappl:
I don't see the readme update in this latest commit

AoHiyuki:
Sorry @cjappl I was kinda slow. Does it look alright?

cjappl:
Awesome, let's give this a try. 

Thanks for your help on this issue. Much appreciated : )

AoHiyuki:
Ah shoot I forgot we need to copy (or symlink) the functions directory in the manual installation or it will not work.

And my change to the "try out" section seems wrong. I have no idea about the Github shortlink tbh.

cjappl:
Reverted it for now while we sort out these issues! thanks for catching them

cjappl:
@wfxr how did you create those short links to begin with, and how would we update git.io/forgit.fish to point to the new file once this is submitted again?

AoHiyuki:
I can't think of an easier way to install the plugin manually @cjappl .

Same goes for the shortlink. It has to curl all the files.

cjappl:
Damn, that's a good point.

I wonder how many of our users use that shortcut versus would use it if it were available on fisher/OhMyFish?

Do you have a plugin that you like to use that you've installed with fisher/OhMyFish that we could look at their repo? maybe they have some hints for us

AoHiyuki:
I use some that are presented in [this repo](https://github.com/jorgebucaran/awesome.fish).

And here are the official [OhMyFish plugins](https://github.com/oh-my-fish). Fisher is compatible with OhMyFish plugins too.

wfxr:
@cjappl 
> how did you create those short links to begin with, and how would we update git.io/forgit.fish to point to the new file once this is submitted again?

It's a shorten URL created by [git.io](https://git.io/).

@AoHiyuki 
Is there a way to add support for fisher/omf without splitting file?

AoHiyuki:
> Is there a way to add support for fisher/omf without splitting file?

@wfxr there is. We can combine all the files into 1 file like before, and put it inside conf.d directory. Fisher will work, but the same problem remains with OhMyFish.

cjappl:
Let's give that a try for the first step, seems like low impact change that can get us functionality on one of the managers at least!

AoHiyuki:
Ok. Then we'll go with that path.

Do you want me to make the change, and commit it in a new pull request? Since all we need to do is few readme changes and rearrange the file, it will be quick.

wfxr:
@AoHiyuki Yes. You can just submit a new pr. I will help update the readme.

PR #126, opened 2020-12-03
Fix forgit::diff's fzf execute because of undefined $cmd

frediz:
## Check list

## Description
This pull request aims to fix a bug I had with fish and using forgit::diff
$cmd was not defined at the time opts get defined resulting in the following error : 
```
fish: Expected a command, but instead found a pipe
 |env LESS='-R' less
 ^
```
The fix is just about moving cmd definition before opts and after repo.
I'm using the fix and seems fine.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [ ] zsh 
    - [x] fish 
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/conf.d/forgit.plugin.fish
+++ b/conf.d/forgit.plugin.fish
@@ -69,13 +69,12 @@ function forgit::diff -d "git diff viewer"
     end
 
     set repo (git rev-parse --show-toplevel)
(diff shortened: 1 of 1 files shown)


cjappl:
d'oh!

Thanks for this great fix. Very much appreciated :)

PR #140, opened 2021-03-08
Use argument to create new branch on gcb

carlosedp:
Previously, forgit alias `gcb` worked to create new branches but later updates broke this.

Here if one pass an argument to gcb, a new branch is created.

## Check list

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [x] zsh 
    - [ ] fish 
- OS
    - [x] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Others:


forgit.plugin.zsh
```diff
@@ -184,6 +184,11 @@ forgit::rebase() {
 
 forgit::checkout::branch() {
     forgit::inside_work_tree || return 1
+    [[ $# -ne 0 ]] && {
+        cmd="git checkout -b $*"
+        eval "$cmd"
+        return
```
wfxr, line 190: `eval` seems not necessary here? And `cmd` should be declared as local otherwise it will be exported to user's environment.

Can we can simplified these into one line?
```bash
[[ $# -ne 0 ]] && git checkout -b "$*" && return
```

wfxr (changes requested):

wfxr:
@carlosedp Merged. Thanks for your contribution!

PR #146, opened 2021-03-25
Add fixup function

ArikDavidov:
Signed-off-by: Ariel Davidov <arik.davidov@gmail.com>

## Check list

## Description

A combination of fixup and autosquash is useful to fix any past commit.
However, using it manually is inconvenient (see: https://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html).
Fixing the last commit is easy: `git commit --amend --no-edit`.
The new function makes fixing any commit as convenient as fixing the last commit.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [x] bash
    - [x] zsh 
    - [ ] fish 
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/forgit.plugin.zsh
+++ b/forgit.plugin.zsh
@@ -187,7 +187,7 @@ forgit::rebase() {
 forgit::fixup() {
     forgit::inside_work_tree || return 1
     git diff --cached --quiet && echo 'Nothing to fixup: there are no staged changes.' && return 1
-    local cmd preview opts graph files commit
+    local cmd preview opts graph files target_commit prev_commit
     graph=--graph
     [[ $FORGIT_LOG_GRAPH_ENABLE == false ]] && graph=
     cmd="git log $graph --color=always --format='$forgit_log_format' $* $forgit_emojify"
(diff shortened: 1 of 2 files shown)


wfxr:
@ArikDavidov Thanks for your contribution and also thanks for letting me known this elegant way to do the fixup! I have always used `git rebase` before.

I made some minor changes based on your pr. Please let me known if there is any misunderstanding.

cjappl:
Hi @ArikDavidov 

Can you elaborate a bit on what a good test case for this would be? I am trying to implement this in fish, but I haven't used this functionality of git before.

If you can provide me an easy to try test case, it would be super valuable for me!

Thanks

ArikDavidov:
Sure @cjappl .

echo "hell world!" > my_file
git add my_file
git commit -m"First commit"
echo "Some other text" > another_file
git add another_file
git commit -m"Second commit"
# At this point you realize you have a typo in the second to last commit ("First commit")
echo "hello world!" > my_file
git add my_file
gfu
# Select "First commit"

# The expected result is that "First commit" is now fixed and there is no additional commits

cjappl:
Thank you @ArikDavidov 

I think I have mine up to what the zsh version is doing, but when I run your steps I am getting additional new commits that say "fixup! First commit", as you can see in my commit history.

I replicated this on zsh as well, so if it is a bug I think it exists in both zsh and fish versions of forgit right now.

Any idea if this is an issue, and how it could be fixed?

Screen Shot 2021-04-05 at 1 58 04 PM

forgit.plugin.zsh
```diff
@@ -199,12 +199,19 @@ forgit::fixup() {
...
-    commit=$(eval "$cmd" | FZF_DEFAULT_OPTS="$opts" fzf --preview="$preview" |
+    target_commit=$(eval "$cmd" | FZF_DEFAULT_OPTS="$opts" fzf --preview="$preview" |
         grep -Eo '[a-f0-9]+' | head -1)
-    [[ -n "$commit" ]] && git commit --fixup "$commit" && \
+    if [[ -n "$target_commit" ]] && git commit --fixup "$target_commit"; then
+        # "$commit~" is invalid when the commit is the first commit, but we can use "--root" instead
+        if [[ "$(git rev-parse "$target_commit")" == "$(git rev-list --max-parents=0 HEAD)" ]]; then
+            prev_commit="--root"
+        else
+            prev_commit="$commit~"
```
ArikDavidov, line 209: We should also rename it here: "$commit~" -> "$target_commit~" (plus in the comment on line 205)

ArikDavidov:
@cjappl  It seems renaming "$commit~" -> "$target_commit~" fixes this  bug.

forgit.plugin.zsh
cjappl: perfect, I'm going to fix this as a part of my checkin, should be done by the end of the day!

cjappl:
Closed with fish #149. Thanks for the help @ArikDavidov

forgit.plugin.zsh
wfxr: @ArikDavidov Thank you for quickly catching this!

PR #154, opened 2021-06-10
Bugfix: gd(git diff) for fish does not work with arguments

ray-x:
## Check list

- [] I have commented my code in hard-to-understand areas
- [] I have made corresponding changes to the documentation

## Description

fish does not use $1 and $2 for arguments, instead it using $argv which is a array/list and to index $1, $argv[1] should be used.
I am not sure why this issue was not reported before. But I believe it has been there for a long time.

To test:

`gd branch_name `

It should works for both zsh and fish. But it only works for zsh ATM. Run command in fish, an error will be shown

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [x] zsh 
    - [x] fish 
- OS
    - [ ] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/conf.d/forgit.plugin.fish
+++ b/conf.d/forgit.plugin.fish
@@ -59,11 +59,11 @@ function forgit::log -d "git commit viewer"
 end
 
 ## git diff viewer
-function forgit::diff -d "git diff viewer" 
+function forgit::diff -d "git diff viewer"
     forgit::inside_work_tree || return 1
     if count $argv > /dev/null
-        if git rev-parse "$1" > /dev/null 2>&1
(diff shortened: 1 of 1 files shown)


cjappl:
Looks great @ray-x. Thanks for the fix!

I'm sure this has been in this for a while, the initial version of this I had to copy and convert a lot of code, so not surprised there are bugs lurking.

Thanks for being proactive and submitting this, much appreciated :)

PR #159, opened 2021-07-19
Fix `gcb` alias in Fish

Cellane:
## Check list

## Description

When defining a custom alias for the `gcb` (Git checkout branch) in alias, the custom alias is not registered; rather, `fish` will omit the following warning on start: `alias: Body cannot be empty`.

I think this is caused due to a typo between the variable that’s being tested for presence and the variable that’s later used for setting the alias.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [ ] zsh 
    - [x] fish 
- OS
    - [ ] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/conf.d/forgit.plugin.fish
+++ b/conf.d/forgit.plugin.fish
@@ -499,7 +499,7 @@ if test -z "$FORGIT_NO_ALIASES"
     end
 
     if test -n "$forgit_checkout_branch"
-        alias $forgit_branch 'forgit::checkout::branch'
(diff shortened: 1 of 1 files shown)


cjappl:
Thanks so much for catching this, and submitting a fix! :)

PR #193, opened 2022-03-27
Fixed: grc fails when commit is preceded by anything else than '* '

sandr01d:
## Check list

## Description

Before this commit, the regex for extracting the commit hash when using `grc` assumed that each line starts with '* '. This is not the case when your commit history contains merges that were not squashed and looks more like this 

```
*   7bf6560 Merge branch 'test-branch' 4 days ago
|\  
| * c753f1c (test-branch) Test commit 5 4 days ago
| * d3c7c6d Test commit 4 4 days ago
* | 1869317 Test commit 6 4 days ago
|/  
* b344261 Test commit 3 3 weeks ago
```
In the example above, selecting the line `| * c753f1c (test-branch) Test commit 5 4 days ago` would fail to revert the commit. This is now fixed.
Note: The fish implementation does not seem to have this issue.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [x] bash
    - [x] zsh
    - [x] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/forgit.plugin.zsh
+++ b/forgit.plugin.zsh
@@ -300,7 +300,7 @@ forgit::checkout::commit() {
 forgit::revert::commit() {
     forgit::inside_work_tree || return 1
     [[ $# -ne 0 ]] && { git revert "$@"; return $?; }
-    local cmd opts files preview commits
+    local cmd opts files preview commits IFS
     cmd="git log --graph --color=always --format='$forgit_log_format' $* $forgit_emojify"
     opts="
         $FORGIT_FZF_DEFAULT_OPTS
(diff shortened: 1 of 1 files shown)


sandr01d:
`git revert "$commits"` would not work for multiple commits because git will fail to recognize that we want to revert separate commits e.g. '3f649be' and 'b344261' instead of one commit '3f649be b344261' (containing a space). `git revert $(echo "$commits")` would work though. I think this is better than using `eval`, as this makes the intent clearer IMO. Should I update the PR accordingly?

wfxr:
> `git revert $(echo "$commits")` would work though. I think this is better than using `eval`, as this makes the intent clearer IMO.

I think the same as you.

sandr01d:
So, shellcheck failed on the solution we've discussed and I took the opportunity for a bit of refactoring. The commit hashes are now passed into an array and I revert each commit one by one. This also solves git logging an error and aborting when a commit is selected that already has been reverted.﻿

wfxr (approved):
LGTM

PR #216, opened 2022-07-12
Improve cherry-pick

carlfriedrich:
This PR improves `gcp` (cherry-pick) with the following features:

1. **Add FORGIT_CHERRY_PICK_FZF_OPTS**
The `gcp` command was the only command which did not have a variable to set command-specific `fzf` options. Add the according variable to source code and documentation.

2. **Preserve commit order on cherry pick**
When using a `fzf` find string which maches multiple lines, the commits could appear in a wrong order. Add `--tiebreak=index` to ensure that the correct commit order is preserved.

3. **Preserve +/- information on cherry pick**
The output of `git cherry` which is used to build the `fzf` input list for `forgit::cherry::pick` usually prefixes every line with a '-' for commits that have an equivalent in the target branch, and a '+' for commits that do not.
Previously forgit removed this information from the list. However, for the actual cherry-picking process this information is relevant, so we should keep it.

4. **Reverse order on cherry pick**
`glo` displays the newest commits on top of the list, while `gcp` displays the newest commits at the bottom. Reverse the order of `gcp` so that it has the same order like `glo`.

## Check list

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [x] bash
    - [ ] zsh
    - [x] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/conf.d/forgit.plugin.fish
+++ b/conf.d/forgit.plugin.fish
@@ -344,8 +344,8 @@ function forgit::cherry::pick -d "git cherry-picking" --argument-names 'target'
     "
     echo $base
     echo $target
-    git cherry "$base" "$target" --abbrev -v | cut -d ' ' -f2- |
-        env FZF_DEFAULT_OPTS="$opts" fzf | cut -d' ' -f1 |
+    git cherry "$base" "$target" --abbrev -v |
+        env FZF_DEFAULT_OPTS="$opts" fzf | cut -d' ' -f2 |
         xargs -I% git cherry-pick %
 end
 
--- a/conf.d/forgit.plugin.fish
+++ b/conf.d/forgit.plugin.fish
@@ -340,6 +340,7 @@ function forgit::cherry::pick -d "git cherry-picking" --argument-names 'target'
         --preview=\"$preview\"
         $FORGIT_FZF_DEFAULT_OPTS
         -m -0
+        $FORGIT_CHERRY_PICK_FZF_OPTS
     "
     echo $base
     echo $target
--- a/conf.d/forgit.plugin.fish
+++ b/conf.d/forgit.plugin.fish
@@ -339,7 +339,7 @@ function forgit::cherry::pick -d "git cherry-picking" --argument-names 'target'
     set opts "
         --preview=\"$preview\"
         $FORGIT_FZF_DEFAULT_OPTS
-        -m -0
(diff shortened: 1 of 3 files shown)


conf.d/forgit.plugin.fish
```diff
@@ -339,12 +339,13 @@ function forgit::cherry::pick -d "git cherry-picking" --argument-names 'target'
...
         $FORGIT_FZF_DEFAULT_OPTS
-        -m -0
+        -m -0 --tiebreak=index
+        $FORGIT_CHERRY_PICK_FZF_OPTS
     "
     echo $base
     echo $target
-    git cherry "$base" "$target" --abbrev -v | cut -d ' ' -f2- |
-        env FZF_DEFAULT_OPTS="$opts" fzf | cut -d' ' -f1 |
+    git cherry "$base" "$target" --abbrev -v | tac |
```
cjappl, line 347: Is `tac` a standard UNIX command? I'm not able to see it on my version of OSX

cjappl:
Love the improvements, just want to understand if we can replicate what `tac` does in a command that is available on both linux and mac!

cjappl:
How about `tail -r` instead? Seems to work on my machine.

```

$ echo "hi
              hello
              world" | tail -r
world
hello
hi
$ echo "hi
              hello
              world" | /bin/cat
hi
hello
world
```

wfxr:
@carlfriedrich Unfortunately gnu `tail` (on linux) doesn't support `-r` option.

```sh
❯ seq 10 | tail -r
tail: invalid option -- 'r'
Try 'tail --help' for more information.
```

carlfriedrich:
Hi @cjappl and @wfxr, thanks for your feedback. I have replaced `tac` with a combination of `tac` and `tail -r`, like proposed in [this StackOverflow answer](https://stackoverflow.com/a/13867870/3018229). Tested with bash and fish. Are you okay with that?

The alternative would be using `sed` or `awk`, but both would be significantly slower, which might make a difference on huge repos.

cjappl:
Looks reasonable enough to me! will leave it to @wfxr to approve officially. Thanks for the adjustment

wfxr (approved):

PR #219, opened 2022-07-14
glo:: use forgit::diff when entering a commit

### LINKED ISSUE #217: Why not use forgit::diff on forgit::log enter?
## Check list

## Environment info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:
- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish

When using `glo` (`forgit::log`) we can press `Enter` to show the changeset of the selected commit. The presentation, however, is rendered using the basic `git diff` command. We have a much more user friendly diff command `gd` (`forgit::diff`) right in forgit, though.

@wfxr @cjappl
Is there a specific reason why we do not use this as an enter command?

If not, I would love to provide a patch to integrate this.

carlfriedrich:
When using `glo` (`forgit::log`) we can press `Enter` to show the changeset of the selected commit. The presentation, however, is rendered using the basic `git diff` command. We have a much more user friendly diff command `gd` (`forgit::diff`) right in forgit, though. Use this instead.

Fixes #217.

## Check list

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [x] bash
    - [x] zsh
    - [x] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -15,6 +15,9 @@
 # Set shell for fzf preview commands
 SHELL=/bin/bash
 
+# Get absolute forgit path
+FORGIT=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)/$(basename -- "${BASH_SOURCE[0]}")
+
 FORGIT_FZF_DEFAULT_OPTS="
 $FZF_DEFAULT_OPTS
 --ansi
@@ -69,11 +72,11 @@ _forgit_log() {
     local opts graph files log_format preview_cmd enter_cmd
     files=$(sed -nE 's/.* -- (.*)/\1/p' <<< "$*") # extract files parameters for `git show` command
     preview_cmd="echo {} | $_forgit_extract_sha | xargs -I% git show --color=always -U$_forgit_preview_context % -- $files | $_forgit_show_pager"
-    enter_cmd="echo {} | $_forgit_extract_sha | xargs -I% git show --color=always -U$_forgit_fullscreen_context % -- $files | $_forgit_show_pager"
+    enter_cmd="echo {} | $_forgit_extract_sha | xargs -I% ${FORGIT} diff %^! $files"
     opts="
         $FORGIT_FZF_DEFAULT_OPTS
         +s +m --tiebreak=index
-        --bind=\"enter:execute($enter_cmd | $_forgit_enter_pager)\"
+        --bind=\"enter:execute($enter_cmd)\"
         --bind=\"ctrl-y:execute-silent(echo {} | $_forgit_extract_sha | ${FORGIT_COPY_CMD:-pbcopy})\"
         --preview=\"$preview_cmd\"
         $FORGIT_LOG_FZF_OPTS


conf.d/forgit.plugin.fish
```diff
@@ -39,7 +39,8 @@ function forgit::log -d "git commit viewer"
     forgit::inside_work_tree || return 1
 
     set files (echo $argv | sed -nE 's/.* -- (.*)/\1/p')
-    set preview "echo {} |grep -Eo '[a-f0-9]+' |head -1 |xargs -I% git show --color=always % -- $files | $forgit_show_pager"
+    set preview_cmd "echo {} |grep -Eo '[a-f0-9]+' |head -1 |xargs -I% git show --stat --color=always % -- $files | $forgit_show_pager"
+    set diff_cmd 'echo {} | grep -Eo "[a-f0-9]+" | head -1 | xargs -I % bash -ic "forgit::diff %^1 %"'
```
cjappl, line 43: Could you extract the shared portion of these into a shared variable just to reduce duplication

```
echo {} |grep -Eo '[a-f0-9]+' |head -1
```

cjappl:
Looks reasonable enough to me! Thanks for the help, leaving to wfxr to approve officially

conf.d/forgit.plugin.fish
carlfriedrich: Good idea! Will do that next week, I am on holidays for the weekend. 🙂

cjappl:
Nice! Have a good trip.

conf.d/forgit.plugin.fish
cjappl: nice! have a good trip :)

carlfriedrich: @cjappl I have implemented your requested change in a separate commit, because it affects a lot more code than the sections I had modifed in the first place. I find it really useful for the code consistency, though, so I added it directly to this PR.
There was also a tiny bug left for git::cherry_pick, where the SHA was not correctly parsed if `FORGIT_LOG_GRAPH_ENABLE` was set to `false`. I fixed this on the way, since it was the only command that parsed the SHA differently than the other commands.
I tested the changes roughly in zsh and fish and rebased the PR on the current master, so we should be ready to go.

cjappl: Wow!! Big improvement, appreciate your thoroughness. 

Going to assign wfxr so he can give final approval

cjappl (approved):
Looks good to me. Please don't merge until @wfxr gives his thumbs up

wfxr:
Looks like something went wrong paring the commit hash?
image

carlfriedrich:
@wfxr Thanks for your feedback. I did not verify my changes in zsh in the first place, so I'm sorry that it didn't work.

After some checking it turns out that my (rather unrelated) [quotation changes](https://github.com/wfxr/forgit/pull/219/commits/0b3b2106b08aa672059248a1205b366ca034e4d6#diff-ab4a968ed4a48481877d55b8f4125a9d5b29c00425ff29515f14f87501837fcfR60) from `\"` to `'` are causing the problem. I always prefer using different quotes over escaping, so I changed this on the run, but obviously zsh interprets this differently than bash.

I updated my branch with the quotation changes reverted. Should work in zsh now.

carlfriedrich:
I just updated my branch again to synchronize the quotation changes to the fish version. Should be ready to go now.

carlfriedrich:
@wfxr Ah, I see. Obviously it tries to use bash for the function call to `forgit::diff`. What's your `$SHELL` environment variable set to?

wfxr:
@carlfriedrich Maybe you should capture the printed commit hash from `forgit::log` and then call `forgit::diff` with it.

carlfriedrich:
@wfxr Well, that's exactly what I'm doing. Strange. Can't reproduce this here. Can you try with a clean shell environment?

```
env -i TERM=xterm-color SHELL=/opt/homebrew/bin/zsh /opt/homebrew/bin/zsh
source forgit.plugin.zsh
glo
```

Does it happen there as well?

wfxr:
> @wfxr Well, that's exactly what I'm doing. Strange. Can't reproduce this here. Can you try with a clean shell environment?
> 
> ```
> env -i TERM=xterm-color SHELL=/opt/homebrew/bin/zsh /opt/homebrew/bin/zsh
> source forgit.plugin.zsh
> glo
> ```
> 
> Does it happen there as well?

@carlfriedrich Still the same.

> Maybe you should capture the printed commit hash from forgit::log and then call forgit::diff with it.

I meant:

```
forgit::log() {
    # ....
    commit_hash=(eval "git log $graph --color=always --format='$log_format' $* $forgit_emojify" |
        FZF_DEFAULT_OPTS="$opts" fzf | capture_commit_hash)
    # call forgit::diff with $commit_hash
}
```

BTW remove `enter` keybind for fzf opts so that the selected line will be printed when hit `enter`.

wfxr:
@carlfriedrich Would you mind if I add a commit to this pr?

carlfriedrich:
@wfxr Give me a few minutes, I am fixing it right now. Your proposal above would not be the same, because the log routine would quit after exiting from the first diff.

wfxr:
@carlfriedrich Still crashes:
image

I'm using [zsh-defer](https://github.com/romkatv/zsh-defer) to load most of the plugins including `forgit` in my `zshrc`. Maybe the plugin is not ready when calling `forgit::diff`?

carlfriedrich:
@wfxr Ah, I wasn't aware of zsh-defer until now. Yes, in this case forgit is obviously not loaded yet when `forgit::diff` is called.

Different to bash, unfortunately [zsh does not support function exporting](https://superuser.com/a/1515756/1036029), so there's no way to pass the functions of the current environment to a new subshell. Starting an interactive shell explicitly, which loads `.zshrc` and along with that `forgit` again, seems to be the only way to have forgit functions in a subshell.

Is there a way to either
- bypass zsh-defer loading or
- wait until defered loading is completed?

carlfriedrich:
@wfxr I found a better solution: instead of a waiting loop, do not rely on forgit being loaded via `.bashrc` / `.zshrc` at all, but reload it explicitly instead. This seems far more reasonable and should work in both bash and zsh. Can you check again?

carlfriedrich:
@wxfr Sorry for spam. :-) I just re-added the `-i` flag for spawning an interactive shell, because otherwise user defined forgit variable settings would be lost for the subshell. If these get loaded deferred as well, I wouldn't know how to work around this, though. Can you check if this works now for you?

wfxr:
Hi @carlfriedrich Thanks for your effort for this PR. It works now. 

https://user-images.githubusercontent.com/6105425/179879898-04420b87-7613-4033-afc4-03c92f1559f1.mp4

However there still some issues.

1. When hit `enter` the following message appears at the top of the window:

image


2. I found I cannot preview the detailed changes through the preview window anymore, which makes it difficult to quickly browse changes. 

image

3. As the video shows the screen will be messed up every first time I hit `enter` in the sub shell.

image

IMO entering a interactive shell in a fzf keybinding is too tricky which will cause some unexpected behaviors we don't known even if the above problems were solved. The other thing I concern is there are a log of guys who use oh-my-zsh or enabled many plugins take quite a few seconds to load the shell. This change will result in `forgit::log` with little usability.

So I tend to keep it simple and stupid.

carlfriedrich:
@wfxr Thanks a lot for your detailed feedback. Really appreciate your responsiveness and will to check my changes.

> Hi @carlfriedrich Thanks for your effort for this PR. Now I can jump into the `forgit::diff`.

Yeah, finally! :-)

> However there still some issues.
> 
>     1. When hit `enter` the following message appears at the top of the window:

Okay, I have no idea where that comes from. :-/

>     2. I found I cannot see the detailed changes through the preview window anymore, which makes it difficult to quickly browse changes.

That was a deliberate change. I like the preview short and overview-like, then being able to enter the change to see the detailed diff. I understand, however, that this is personal preference and doesn't meet everyone's taste.

>     3. As the video shows the screen will be messed up every first time I hit `enter` in the sub shell.

I have no idea about that either, and cannot reproduce it unfortunately.

> IMO entering a interactive shell in a fzf keybinding is too tricky which will cause some unexpected behaviors we don't known even if the above problems were solved. The other thing I concern is there are a log of guys who use oh-my-zsh or enabled many plugins take quite a few seconds to load the shell. This change will result in `forgit::log` with little usability.

I see your point here. Obviously your zsh setup is already so much different from my test environment that it causes issues with this change, so there will probably be more issues with other setups.

I just wanted to bring something upstream that I am using in bash for quite a few months now, so that other people can benefit from it. However, I didn't expect it to be incompatible with zsh. Too bad. :-/

One last option would be to make this feature bash-only, but I assume that you would like to keep the features for all shells on par?

And one last general question for the bigger picture: I am not familiar with the zsh / oh-my-zsh plugin system. Is there a way to have a plugin that adds its folder to the PATH, so that we could have forgit in form of an executable script instead of env functions? Would that be an option? Because IMO we could use forgit feature chaining in more scenarios than just with `glo` -> `gd` (e.g. cherry picking by first interactively selecting the source branch), which would make forgit much more user-friendly. And with forgit being an executable script we wouldn't have the issues concerning the functions being unknown to subshells.

carlfriedrich:
So here's another shot for integrating `forgit::diff` into `forgit::log`, similar to the cherry-pick implementation of #216 and just like proposed by @wfxr [above](https://github.com/wfxr/forgit/pull/219#issuecomment-1188692634). This implementation does not mess with subshells in the enter command anymore and makes use of the actual fzf output.

I do not find it ideal, though. After quitting the diff, a new fzf instance is started, so the previous selection is gone. This makes it hard to browse through subsequent commit diffs. Unfortunately fzf [does not provide a way to pre-select a line](https://github.com/junegunn/fzf/issues/924).

@wfxr @cjappl What do you think? On one hand I think this is an improvement concerning usability, because forgit's diff function provides a better (and more consistent) user experience than a plain `git diff`. On the other hand the resetting log window after each diff is a step back imo.

wfxr:
@carlfriedrich I tested this implementation. As you said, both improvements and flaws are significant. How about making it an optional feature to satisfy users who like this feature more without affecting others?

carlfriedrich:
With the changes of #241 we have finally a proper way to do this. Since the new commit was actually part of #241 during testing and approved along with those changes, I am directly merging this as well.

nathanchance:
For what it's worth, this messes with the ability to view larger commit messages in `forgit::log`, which means I cannot use this much for my Linux kernel work :/ is there a way for this feature to coexist with that ability?

carlfriedrich:
@nathanchance I'm sorry that this disrupts your workflow.

You can bring back the old behaviour by defining a custom enter command and adding it to your `FORGIT_LOG_FZF_OPTS` variable before sourcing forgit:

```shell
export FORGIT_LOG_ENTER_CMD="echo {} | grep -Eo '[a-f0-9]+' | head -1 | tr -d '[:space:]' | xargs -I% git show --color=always % | LESS='-Rc' less"
export FORGIT_LOG_FZF_OPTS="--bind='enter:execute($FORGIT_LOG_ENTER_CMD)'"
```

But may I ask why the new behavior actually prevents you from viewing the commit message? Don't you have it displayed in your preview window?
FYI: You can scroll within forgit's preview window by defining [fzf keybindings in `FORGIT_FZF_DEFAULT_OPTS`](https://www.mankier.com/1/fzf#Key/Event_Bindings). forgit already has defined [standard values for this](https://github.com/wfxr/forgit/blob/42c8e005fece723de68d78906bb3ba7b70ff65c7/bin/git-forgit#L25..L26) (Down: Alt+J / Alt+N, Up: Alt+K / Alt+P).

Would one of these two options work for you?

nathanchance:
> @nathanchance I'm sorry that this disrupts your workflow.

No worries!

> You can scroll within forgit's preview window by defining [fzf keybindings in `FORGIT_FZF_DEFAULT_OPTS`](https://www.mankier.com/1/fzf#Key/Event_Bindings). forgit already has defined [standard values for this](https://github.com/wfxr/forgit/blob/42c8e005fece723de68d78906bb3ba7b70ff65c7/bin/git-forgit#L25..L26) (Down: Alt+J / Alt+N, Up: Alt+K / Alt+P).

Aha, I did not know about being able to scroll the preview window! That will work just fine for this, thanks a lot for pointing it out!

carlfriedrich:
@nathanchance Great, thanks for your feedback!

PR #220, opened 2022-07-14
Fish didn't like the tac/tail switch. Also removed some print statements

cjappl:
Fix for fish `gcp` after the recent change.

Working well on my machine now! @carlfriedrich would you review and make sure this seems reasonable? 

@wfxr would you try `gcp` with your config just for some QA and make sure it's running well :)


cjappl:
No apologies necessary! happens to me all the time 😄 

Thanks for the look over, going to merge

wfxr:
@cjappl Sorry for the late reply. It works as expected in zsh.

PR #221, opened 2022-07-17
Improve forgit::diff

carlfriedrich:
While we can see the compact diff in the preview window on `gd`, I find it useful to see the complete file when entering the diff. Set the diff context to 10000 for this purpose.

Before:
grafik

After:
grafik

Furthermore, display the diff range in the fzf prompt to see what we're looking at. With the normal `git diff` this is part of the diff output anyway, but with `diff-so-fancy` it's not, so this should be useful.

Before:
grafik

After:
grafik

## Check list

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [x] bash
    - [ ] zsh
    - [x] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/forgit.plugin.zsh
+++ b/forgit.plugin.zsh
@@ -16,19 +16,22 @@ forgit_ignore_pager=${FORGIT_IGNORE_PAGER:-$(hash bat &>/dev/null && echo 'bat -
 forgit_blame_pager=${FORGIT_BLAME_PAGER:-$(git config pager.blame || echo "$forgit_pager")}
 
 forgit_log_format=${FORGIT_LOG_FORMAT:-%C(auto)%h%d %s %C(black)%C(bold)%cr%Creset}
+forgit_fullscreen_context=${FORGIT_FULLSCREEN_CONTEXT:-10}
+forgit_preview_context=${FORGIT_PREVIEW_CONTEXT:-3}
 
 # git commit viewer
 forgit::log() {
     forgit::inside_work_tree || return 1
-    local cmd opts graph files log_format
+    local opts graph files log_format preview_cmd enter_cmd
     files=$(sed -nE 's/.* -- (.*)/\1/p' <<< "$*") # extract files parameters for `git show` command
-    cmd="echo {} |grep -Eo '[a-f0-9]+' |head -1 |xargs -I% git show --color=always % -- $files | $forgit_show_pager"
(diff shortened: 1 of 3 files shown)


cjappl:
Will you do us a favor and attach a screenshot of before and after?

It can be useful in documenting how things are supposed to perform in case of bugs!

Appreciate all of your submissions recently :)

carlfriedrich:
@cjappl Sure, I have updated the PR description with screenshots.

forgit.plugin.zsh
```diff
@@ -54,12 +54,16 @@ forgit::diff() {
         fi
     }
     repo="$(git rev-parse --show-toplevel)"
-    cmd="cd '$repo' && echo {} |sed 's/.*] *//' | sed 's/  ->  / /' |xargs -I% git diff --color=always $commits -- % | $forgit_diff_pager"
+    get_files="cd '$repo' && echo {} | sed 's/.*] *//' | sed 's/  ->  / /'"
+    preview_cmd="$get_files | xargs -I% git diff --color=always $commits -- % | $forgit_diff_pager"
+    # Show additional context on enter compared to preview
+    enter_cmd="$get_files | xargs -I% git diff --color=always -U10000 $commits -- % | $forgit_diff_pager"
```
wfxr, line 60: Maybe `10` is enough? Imagine there will be too many irrelevent lines displied when you modified a large file.

wfxr (changes requested):

forgit.plugin.zsh
wfxr: > Would you accept making the diff context size configurable via a variable?

LGTM!

carlfriedrich: @wfxr Great, I added a commit which introduces variables for the diff context. Can you check?

wfxr:
@carlfriedrich LGTM!

I have renamed the option name and applied this feature to `forgit::log` (bash / zsh) as well based on your commits. Can you help finish the `fish` part?

wfxr (approved):

carlfriedrich:
@wfxr Thanks for merging!

PR #231, opened 2022-08-16
Adding wraps to fish - adding way to gcb an existing branch

cjappl:
I added a bunch of "wraps" commands to our functions. This is basically a quick and easy way to get function completion for free. Users can now tab complete most commands that take an argument

### DIFF
--- a/conf.d/forgit.plugin.fish
+++ b/conf.d/forgit.plugin.fish
@@ -300,17 +300,11 @@ function forgit::branch::delete -d "git checkout branch deleter" --wraps "git br
 end
 
 
-# note we "wrap" delete, we just need a snazzy way to get the branch list here
-function forgit::checkout::branch -d "git checkout branch selector" --argument-names 'input_branch_name' --wraps "git branch --delete" 
+function forgit::checkout::branch -d "git checkout branch selector" --argument-names 'input_branch_name' --wraps "git branch"
     forgit::inside_work_tree || return 1
 
     if test -n "$input_branch_name"
-        if git branch --list | grep "$input_branch_name" > /dev/null
-            git switch "$input_branch_name"
-        else
-            git switch -c "$input_branch_name"
-        end
-
+        git checkout -b "$input_branch_name"
         set checkout_status $status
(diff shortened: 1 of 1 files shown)


conf.d/forgit.plugin.fish
```diff
@@ -300,12 +300,17 @@ function forgit::branch::delete -d "git checkout branch deleter"
 end
 
 
-
-function forgit::checkout::branch -d "git checkout branch selector" --argument-names 'input_branch_name'
+# note we "wrap" delete, we just need a snazzy way to get the branch list here
+function forgit::checkout::branch -d "git checkout branch selector" --argument-names 'input_branch_name' --wraps "git branch --delete"
```
carlfriedrich, line 304: Why not use `--wraps "git branch --all"` here, just like it is used for the fzf input? This way completion and fzf window would show the same options, which would be less surprising to me.

carlfriedrich (commented):
Hey @cjappl, such a good idea and such a cool feature of fish!

In bash/zsh it's definitely not that easy, but I think it sould be possible to achieve similar behavior using `complete` (for bash) / `compdef` (for zsh), we just have to find the according completion functions for each command.

carlfriedrich:
Can we maybe split this into two MRs? One for the completion and one for the branch switching? That would make the changesets more clear.

carlfriedrich:
For bash, adding bash completion for gcb would work like this:

`__git_complete gcb _git_checkout`

@wfxr Do you know the equivalent for zsh?

cjappl:
> Can we maybe split this into two MRs? One for the completion (this one) and one for the branch switching (a new one)? That would make the changesets more clear.

Yeah, good idea! Should have done it to begin with. I'm going to make this one the autocomplete one, and the other one will be the branch functionality.

cjappl:
I'm going to merge this as is, seeing as I think it's a boost for our fish users. If we want to open a task or PR to do autocomplete in zsh/bash that would be cool too :)  I have no idea if that functionality exists or not for those shells.

PR #232, opened 2022-08-16
Add ability to checkout existing branch with gcb

cjappl:
Part two of the original #231 

It would be nice to add this functionality to bash/zsh @wfxr or @carlfriedrich if you have time :) 

Going to merge this basically straight away, as I see it being a huge gain to my personal workflow!

### DIFF
--- a/conf.d/forgit.plugin.fish
+++ b/conf.d/forgit.plugin.fish
@@ -304,7 +304,11 @@ function forgit::checkout::branch -d "git checkout branch selector" --argument-n
     forgit::inside_work_tree || return 1
 
     if test -n "$input_branch_name"
-        git checkout -b "$input_branch_name"
+        if git branch --list | grep "$input_branch_name" > /dev/null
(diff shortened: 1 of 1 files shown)


carlfriedrich:
@cjappl I don't understand what this patch actually does. Wasn't it possible before to checkout existing branches?

cjappl:
At least in fish the previous behavior if you “gcb main” (where main existed before) was an error:

“branch main already exists, cannot be created”

This may work in zsh/bash, I’m not sure, but it hadn’t worked in fish until this pr!

This nee logic in psuedocode is:

If an argument is provided to gcb
If that argument is in the list of branches
Check out the existing branch
Else
Create and check out that branch
Return

Where as before it was

If an argument is provided to gcb
Create and check out that branch

That create would fail if the branch already existed.

Does that make sense? It’s totally possible I missed something here.

Also apologies if the formatting ended up whacky, doing this on my phone at 5AM! 😀

PR #234, opened 2022-09-12
Add ability to checkout existing branch with gcb

carlfriedrich:
Port @cjappl 's changes of #232 / d4b4ce39087a2dfe4812d54de7a9865fb2ba9855 to zsh/bash.

`gcb` can now be called with an existing branch as an argument.

## Check list

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/forgit.plugin.zsh
+++ b/forgit.plugin.zsh
@@ -258,7 +258,18 @@ forgit::checkout::file() {
 # git checkout-branch selector
 forgit::checkout::branch() {
     forgit::inside_work_tree || return 1
-    [[ $# -ne 0 ]] && { git checkout -b "$@"; return $?; }
+    # if called with arguments, check if branch exists, else create a new one
+    if [[ $# -ne 0 ]]; then
+        if git show-branch "$@" &>/dev/null; then
+            git switch "$@"
+        else
+            git switch -c "$@"
+        fi
+        checkout_status=$?
+        git status --short
+        return $checkout_status
+    fi
+
     local cmd preview opts branch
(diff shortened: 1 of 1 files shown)


forgit.plugin.zsh
```diff
@@ -258,7 +258,17 @@ forgit::checkout::file() {
...
+    # if called with arguments, check if branch exists, else create a new one
+    if [[ $# -ne 0 ]]; then
+        if git show-branch "$@" &>/dev/null; then
+            git switch "$@"
+            return $?
+        else
+            git checkout -b "$@"
+            return $?
+        fi
+    fi
```
cjappl, line 270: If we can, it would be great to do similar to what we have in fish, that is in psuedocode:

```
saved_old_error_code = git switch_or_checkout
git status --short
return $checkout_status
```

Other than that this looks good! I'll add @wfxr to review, but with this requested change I'm happy to approve unilaterally, as I wrote the original fish PR

cjappl (changes requested):

forgit.plugin.zsh
carlfriedrich: Thanks for your feedback! Good idea, I have added the git status call.

cjappl (approved):
Cool! Looks good to me. I'm going to merge directly as I feel this change isn't super risky and gives good functionality :) 

I've been loving it in fish.

PR #235, opened 2022-09-13
Add bash completions for forgit functions and aliases

carlfriedrich:
This is my shot for porting @cjappl 's changes of #231 to bash. Not every forgit function supports arguments at the moment, but at least for the ones that support it completion is a useful thing, so I added it for all functions at once, just like in fish.

@wfxr What do you think? And do you know how this could be implemented for zsh?

Tested all completions in bash.

## Check list

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,6 +289,8 @@ export FORGIT_LOG_FZF_OPTS='
 - `gd` supports specifying revision(eg, `gd HEAD~`, `gd v1.0 README.md`).
 - Call `gi` with arguments to get the wanted `.gitignore` contents directly(eg, `gi cmake c++`).
 - You can use the commands as sub-commands of `git`, see [#147](https://github.com/wfxr/forgit/issues/147) for details.
+- Put [completions/git-forgit.bash](https://github.com/wfxr/forgit/blob/master/completions/git-forgit.bash) in
+  `~/.local/share/bash-completion/completions` to have bash tab completion for `git forgit` and configured git aliases.
 
 ### 📃 License
 
--- a/completions/git-forgit.bash
+++ b/completions/git-forgit.bash
@@ -1,0 +1,106 @@
+# forgit completions for bash
+
+# When using forgit as a subcommand of git, put this file in one of the
+# following places and it will be loaded automatically on tab completion of
+# 'git forgit' or any configured git aliases of it:
+#
+#   /usr/share/bash-completion/completions
+#   ~/.local/share/bash-completion/completions
+
+_git_branch_delete()
+{
+	__gitcomp_nl "$(__git_heads)"
+}
+
+_git_checkout_branch()
+{
+	__gitcomp_nl "$(__git branch -a --format '%(refname:short)')"
+}
+
+_git_checkout_file()
+{
+	__gitcomp_nl "$(__git ls-files --modified)"
+}
+
+_git_checkout_tag()
+{
+	__gitcomp_nl "$(__git_tags)"
+}
+
+_git_stash_show()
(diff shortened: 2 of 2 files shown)


cjappl:
Cool! I had no idea that bash supported such a thing 👀

wfxr:
> @wfxr What do you think? And do you know how this could be implemented for zsh?

@carlfriedrich Good idea! I'm not familiar with this but you can refer to fzf: https://github.com/junegunn/fzf/tree/master/shell

There are two suggestions:
1. Complete the original functions instead of the aliases?
2. Move the completion code into a separate file like many other tools to avoid code bloat.

carlfriedrich:
@wfxr Thanks for your feedback.

>     1. Complete the original functions instead of the aliases?

Are there actually people who use the functions directly? Or does zsh complete aliases automatically equally to the commands they alias to? Bash at least doesn't.

>     2. Move the completion code into a separate file like many other tools to avoid code bloat.

Is this compatible with the oh-my-zsh plugin system? Where should these completions go within the forgit tree then?

Since we're mainly re-using existing git completions, I don't think the code gets bloated too much. And the fish code includes the completions directly in the code as well. But I'm open for discussion here.

wfxr:
@carlfriedrich Thank you. Serveral lines don't make the code bloated but many times of serveral lines will do. I prefer to keep the single main file short and efficient. Not just for running efficiently, but also loading efficiently, and reading efficiently.

> Are there actually people who use the functions directly? Or does zsh complete aliases automatically equally to the commands they alias to? Bash at least doesn't.

`zsh` does support complete aliases the same as it's original command.

> Is this compatible with the oh-my-zsh plugin system?

IDK but should not a problem since there are many pure completion plugins in omz. Also most plug managers support to specify the completion file path when loading the plugin.

> Where should these completions go within the forgit tree then?

A new directory in root named `completions` is ok for me.

carlfriedrich:
> @carlfriedrich Thank you. Serveral lines don't make the code bloated but many times of serveral lines will do. I prefer to keep the single main file short and efficient. Not just for running efficiently, but also loading efficiently, and reading efficiently.

@wfxr OK, I get your point here.

> > Are there actually people who use the functions directly? Or does zsh complete aliases automatically equally to the commands they alias to? Bash at least doesn't.
> 
> Someone use custom aliases instead the default. `zsh` does support complete aliases the same as it's original command.

Ah, good to know. I will consider this.
 
> > Where should these completions go within the forgit tree then?
> 
> A new directory in root named `completions` is ok for me.

Alright, I will transfer the completions accordingly.

carlfriedrich:
@wfxr I updated this PR to the current master. Do you mind if I merge this? It adds completions for bash in a dedicated file. I have actually been using these for several weeks now, so they are proven working. The patch should not interfere with zsh or fish environments.

carlfriedrich:
I decied to adapt this PR to support the following two completions only:

- `git forgit` calls (e.g. `git forgit add`)
- `git forgit` aliases (e.g. `git a` with `alias.a=forgit add` in git config)

These are my only actual use cases and they don't require any changes in the code, so I consider this PR minimal obstrusive and will merge it directly now.

For the plugin functions/aliases completion I will create another PR.

PR #237, opened 2022-09-29
Improve handling of arguments in forgit::add

qubidt:
currently, passing arguments to `forgit::add` bypasses the interactive selector and are passed directly to `git add`. This isn't particularly useful. So pass the arguments to the `git status` call instead, so fzf can be used to interactively select from the list.

This is particularly useful when calling `git forgit add` against a subdirectory or a shell glob (e.g. `git forgit add *(om[1,5])`)

## Check list

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/forgit.plugin.zsh
+++ b/forgit.plugin.zsh
@@ -102,8 +102,6 @@ forgit::diff() {
 # git add selector
 forgit::add() {
     forgit::inside_work_tree || return 1
-    # Add files if passed as arguments
-    [[ $# -ne 0 ]] && git add "$@" && git status -su && return
 
     local changed unmerged untracked files opts preview extract
     changed=$(git config --get-color color.status.changed red)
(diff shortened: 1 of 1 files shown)


carlfriedrich:
Hey @qubidt, thanks for your suggestion.

The behavior for passing arguments to `forgit::add` has been deliberately implemented that way in e73cef6856e826173454ddfa1df9fb37fe9cff2f. It makes adding a file via the command line possible using the same command as adding a file interactively, so we don't have to use different commands for each use case. We actually handle arguments the same way on most other forgit commands as well (e.g. `forgit::checkout::branch` or `forgit::revert::commit`). I noticed that this is not documented, yet, I will put that on my to-do-list.

However, I see your use-case, and I also see that fzf does not support glob syntax. Since I assume we do not want to change the default behavior, I would rather only accept this in combination with an opt-in config, though.

@wfxr @cjappl What do you say?

qubidt:
That's a very reasonable compromise! 😃I'm out of town currently so I can't implement it but I'll update the pr when I'm able

cjappl:
Pinging @qubidt - we will close soon unless we hear a response / updated PR :)

qubidt:
sorry! dropped off my plate but I'll see if I can revisit this soon!

cjappl:
Closing, can re-open in the future if we get some new revisions :)

PR #238, opened 2022-10-03
Do not "set -e" in standalone forgit

carlfriedrich:
Setting the error option in bash bypasses error handling in the forgit functions, because the script immediatley exits in case of an error. Remove the option to make error handling possible. Bash scripts automatically return the exit code of the last statement, so standalone forgit will still return useful error codes.

## Check list

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -10,8 +10,6 @@
 # This gives users the choice to set aliases inside of their git config instead
 # of their shell config if they prefer.
 
-set -e
-
 source "$FORGIT_INSTALL_DIR/forgit.plugin.zsh"
 
 cmd="$1"


carlfriedrich:
Hey @wren, since you added the `git-forgit` script, can you check if this PR would break anything I don't see?
In #236 we added error handling to some of the forgit functions. Those are are rendered useless when using forgit via `git-forgit` due to the error option, so I would like to remove it.

wren:
@carlfriedrich I don't think it would break anything, but I haven't gone through to test the new error handling, so I can't say for sure.

carlfriedrich:
> @carlfriedrich I don't think it would break anything, but I haven't gone through to test the new error handling, so I can't say for sure.

@wren Thanks for your reply! Then I'm gonna merge this. If any error case does not behave as it should, please create an issue, so that we can improve the error handling.

PR #239, opened 2022-10-03
Fix function call in standalone forgit

carlfriedrich:
From the standalone forgit command we call the according forgit function by replacing a '_' with '::'. Since we have more complex function names like 'forgit::reset::head' or ' forgit::stash::show' we have to replace ALL instances of '_' instead of just the first one.

## Check list

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:


wren:
@carlfriedrich The current code replaces the first underscore with `::` to match the function names. This doesn't count `forgit::`, though (since that's hardcoded). So, `forgit::stash:show` and `forgit::reset::head` should both already work (I have `ss = forgit stash_show` as an alias in my own config).

When `git-forgit` was [merged](https://github.com/wfxr/forgit/commit/eb6db722cd494c4678a0e8c849c9d838ff42042c), there weren't any functions that had more than that many colons, so it covered all the use cases at the time (but was admittedly not very future-proof).

The only function I see now that has more colons is `forgit::cherry::pick::from::branch` (which looks to have been [merged](https://github.com/wfxr/forgit/commit/b9c0e7b58fbd18e0c040b5fc827196261f6e71ff) afterward). So, this fix should work to update `git-forgit` to work with newer functions like `forgit::cherry::pick::from::branch`.

Good find!

wren (approved):
👍

carlfriedrich:
> The only function I see now that has more colons is `forgit::cherry::pick::from::branch` (which looks to have been [merged](https://github.com/wfxr/forgit/commit/b9c0e7b58fbd18e0c040b5fc827196261f6e71ff) afterward). So, this fix should work to update `git-forgit` to work with newer functions like `forgit::cherry::pick::from::branch`. 

@wren You're totally right. I actually even noticed the error on `forgit::cherry::pick::from::branch`, but then I just chose different functions for the bug report without checking those, forgetting about the hard-coded `forgit::` extraction.

Thanks for your quick reply!

PR #241, opened 2022-10-15
Move implementation to standalone forgit

carlfriedrich:
This is something I wanted to do for quite a while, and during the last days I finally had some time to check out how to do it right (at least that's what I hope 😁).

I labeled this a draft, because this is kind of a new direction for forgit and I would like to discuss if we can all agree on this first, or if you have any other ideas how to improve this suggestion. If we actually can agree on it, I will adapt the documentation before merging this.

What this patchset basically does is move the actual forgit functionality from the shell plugin (`forgit.plugin.zsh` / `forgit.plugin.fish`) to the standalone forgit executable script (`git-forgit`). This has the following advantages:

1. We don't have to maintain two different forgit implementations for two different shells (zsh/fish) anymore.
2. We can call forgit functions from within other forgit functions (which finally enables an easy implementation for #219).
3. We reduce plugin loading time since the plugin code becomes much smaller.

The simplest way of using forgit after this change is to put `git-forgit` somewhere in your PATH and call it via `git forgit`, without the need for any other files.

We keep the shell plugins, though, and maintain backwards compatibility by retaining the plugin functions as simple wrappers around the `git-forgit` script, making the plugins mainly a collection of aliases only (which is what actually most other shell plugins are, too).

Here's some more detailed information on the patches contained in the patchset:

1. The first patch looks quite large, as it contains lots of changed lines. The only thing it does, though, is move the forgit functions over to `git-forgit`, give them a new (more [shell script convention compliant](https://github.com/icy/bash-coding-style#function-names)) naming style and re-implement the original plugin functions (of both zsh and fish version) as simple wrappers around `git-forgit`. It also moves the detection of `FORGIT_INSTALL_DIR` from the bottom to the top of `forgit.plugin.zsh`.

2. The second patch refactors the detection of `FORGIT_INSTALL_DIR` to match the [standardized $0 handling for zsh plugins](https://zdharma-continuum.github.io/Zsh-100-Commits-Club/Zsh-Plugin-Standard.html#zero-handling), which I noticed wasn't the case before.

3. The third patch just changes the order of the functions in the fish version of the plugin to match the order of the zsh version and also adds `forgit::checkout::tag` and `forgit::blame` to it, which I noticed were missing until now.

4. The fourth patch is the new implementation of #219 and it's just here temporarily for testing. When checking out this branch you should see `forgit::diff` opening when entering a commit in `forgit::log`. I will remove the patch from this branch before merging this PR and move it to #219 instead.

I tested this branch in zsh using [**zplug**](https://github.com/zplug/zplug). Since I think we comply with the [zsh plugin standard](https://zdharma-continuum.github.io/Zsh-100-Commits-Club/Zsh-Plugin-Standard.html#bin-dir), I assume it should also work with other plugin managers as well.
In fish I tested it using [**fisher**](https://github.com/jorgebucaran/fisher) (requiring fisher version [4.4.3](https://github.com/jorgebucaran/fisher/releases/tag/4.4.3) because I had to add correct handling for symbolic links in order to make it work there). It should also work with [**oh-my-fish**](https://github.com/oh-my-fish/oh-my-fish), but since that does not support installing plugins from a specific branch, I could not actually test it there. Installing it and then manually changing the branch did work, though.

So @wfxr, @cjappl and @wren (also adding you here since you added `git-forgit` in the first place), I am very eager to hear your opinions on this. And of course I would like to hear if the change works in your environment. I tried to not make this a breaking change, but we all know there's always something you don't consider, so I'm quite curious to hear from you. 😄

## Check list

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [x] bash
    - [x] zsh
    - [x] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/conf.d/bin/git-forgit
+++ b/conf.d/bin/git-forgit
@@ -1,0 +1,1 @@
+../../bin/git-forgit
--- a/forgit.plugin.zsh
+++ b/forgit.plugin.zsh
@@ -3,7 +3,7 @@
 
 forgit::error() { printf "%b[Error]%b %s\n" '\e[0;31m' '\e[0m' "$@" >&2; return 1; }
 
-# set installation path
+# determine installation path
 if [[ -n "$ZSH_VERSION" ]]; then
     # shellcheck disable=2277,2296,2299
     0="${ZERO:-${${0:#$ZSH_ARGZERO}:-${(%):-%N}}}"
@@ -15,95 +15,95 @@ elif [[ -n "$BASH_VERSION" ]]; then
 else
     forgit::error "Only zsh and bash are supported"
 fi
-export FORGIT_INSTALL_DIR
+export FORGIT="$FORGIT_INSTALL_DIR/bin/git-forgit"
 
 # register shell functions
 forgit::log() {
-    "$FORGIT_INSTALL_DIR/bin/git-forgit" log "$@"
+    "$FORGIT" log "$@"
 }
 
 forgit::diff() {
-    "$FORGIT_INSTALL_DIR/bin/git-forgit" diff "$@"
+    "$FORGIT" diff "$@"
 }
 
 forgit::add() {
-    "$FORGIT_INSTALL_DIR/bin/git-forgit" add "$@"
+    "$FORGIT" add "$@"
 }
 
 forgit::reset::head() {
-    "$FORGIT_INSTALL_DIR/bin/git-forgit" reset_head "$@"
+    "$FORGIT" reset_head "$@"
 }
 
 forgit::stash::show() {
-    "$FORGIT_INSTALL_DIR/bin/git-forgit" stash_show "$@"
+    "$FORGIT" stash_show "$@"
 }
 
 forgit::clean() {
-    "$FORGIT_INSTALL_DIR/bin/git-forgit" clean "$@"
+    "$FORGIT" clean "$@"
 }
 
 forgit::cherry::pick() {
-    "$FORGIT_INSTALL_DIR/bin/git-forgit" cherry_pick "$@"
+    "$FORGIT" cherry_pick "$@"
 }
 
 forgit::cherry::pick::from::branch() {
-    "$FORGIT_INSTALL_DIR/bin/git-forgit" cherry_pick_from_branch "$@"
+    "$FORGIT" cherry_pick_from_branch "$@"
 }
 
 forgit::rebase() {
-    "$FORGIT_INSTALL_DIR/bin/git-forgit" rebase "$@"
+    "$FORGIT" rebase "$@"
 }
 
 forgit::fixup() {
-    "$FORGIT_INSTALL_DIR/bin/git-forgit" fixup "$@"
+    "$FORGIT" fixup "$@"
 }
 
 forgit::checkout::file() {
-    "$FORGIT_INSTALL_DIR/bin/git-forgit" checkout_file "$@"
+    "$FORGIT" checkout_file "$@"
 }
 
 forgit::checkout::branch() {
-    "$FORGIT_INSTALL_DIR/bin/git-forgit" checkout_branch "$@"
+    "$FORGIT" checkout_branch "$@"
 }
 
 forgit::checkout::tag() {
-    "$FORGIT_INSTALL_DIR/bin/git-forgit" checkout_tag "$@"
+    "$FORGIT" checkout_tag "$@"
 }
 
 forgit::checkout::commit() {
-    "$FORGIT_INSTALL_DIR/bin/git-forgit" checkout_commit "$@"
+    "$FORGIT" checkout_commit "$@"
 }
 
 forgit::branch::delete() {
-    "$FORGIT_INSTALL_DIR/bin/git-forgit" branch_delete "$@"
+    "$FORGIT" branch_delete "$@"
 }
 
 forgit::revert::commit() {
-    "$FORGIT_INSTALL_DIR/bin/git-forgit" revert_commit "$@"
+    "$FORGIT" revert_commit "$@"
 }
 
 forgit::blame() {
-    "$FORGIT_INSTALL_DIR/bin/git-forgit" blame "$@"
+    "$FORGIT" blame "$@"
 }
 
 forgit::ignore() {
-    "$FORGIT_INSTALL_DIR/bin/git-forgit" ignore "$@"
+    "$FORGIT" ignore "$@"
 }
 
 forgit::ignore::update() {
-    "$FORGIT_INSTALL_DIR/bin/git-forgit" ignore_update "$@"
+    "$FORGIT" ignore_update "$@"
 }
 
 forgit::ignore::get() {
-    "$FORGIT_INSTALL_DIR/bin/git-forgit" ignore_get "$@"
+    "$FORGIT" ignore_get "$@"
 }
 
 forgit::ignore::list() {
-    "$FORGIT_INSTALL_DIR/bin/git-forgit" ignore_list "$@"
+    "$FORGIT" ignore_list "$@"
 }
 
 forgit::ignore::clean() {
-    "$FORGIT_INSTALL_DIR/bin/git-forgit" ignore_clean "$@"
+    "$FORGIT" ignore_clean "$@"
 }
 
 # register aliases
--- a/forgit.plugin.zsh
+++ b/forgit.plugin.zsh
@@ -2,6 +2,7 @@
 # MIT (c) Wenxuan Zhang
 
 forgit::error() { printf "%b[Error]%b %s\n" '\e[0;31m' '\e[0m' "$@" >&2; return 1; }
+forgit::warn() { printf "%b[Warn]%b %s\n" '\e[0;33m' '\e[0m' "$@" >&2; }
 
 # determine installation path
 if [[ -n "$ZSH_VERSION" ]]; then
@@ -9,13 +10,29 @@ if [[ -n "$ZSH_VERSION" ]]; then
     0="${ZERO:-${${0:#$ZSH_ARGZERO}:-${(%):-%N}}}"
     # shellcheck disable=2277,2296,2298
     0="${${(M)0:#/*}:-$PWD/$0}"
-    FORGIT_INSTALL_DIR="${0:h}"
+    INSTALL_DIR="${0:h}"
 elif [[ -n "$BASH_VERSION" ]]; then
-    FORGIT_INSTALL_DIR="$(dirname -- "${BASH_SOURCE[0]}")"
+    INSTALL_DIR="$(dirname -- "${BASH_SOURCE[0]}")"
 else
     forgit::error "Only zsh and bash are supported"
 fi
-export FORGIT="$FORGIT_INSTALL_DIR/bin/git-forgit"
+FORGIT="$INSTALL_DIR/bin/git-forgit"
+
+# backwards compatibility:
+# export all user-defined FORGIT variables to make them available in git-forgit
+unexported_vars=0
+set | awk -F '=' '{ print $1 }' | grep FORGIT_ | while read -r var; do
+    if ! export | grep -q "^$var="; then
+        if [[ $unexported_vars == 0 ]]; then
+            forgit::warn "Config options have to be exported in future versions of forgit."
+            forgit::warn "Please update your config accordingly:"
+        fi
+        forgit::warn "  export $var"
+        unexported_vars=$((unexported_vars + 1))
+        # shellcheck disable=SC2163
+        export "$var"
+    fi
+done
 
 # register shell functions
 forgit::log() {
--- a/forgit.plugin.zsh
+++ b/forgit.plugin.zsh
@@ -1,510 +1,107 @@
 #!/usr/bin/env bash
 # MIT (c) Wenxuan Zhang
+
 forgit::warn() { printf "%b[Warn]%b %s\n" '\e[0;33m' '\e[0m' "$@" >&2; }
-forgit::info() { printf "%b[Info]%b %s\n" '\e[0;32m' '\e[0m' "$@" >&2; }
-forgit::inside_work_tree() { git rev-parse --is-inside-work-tree >/dev/null; }
-# tac is not available on OSX, tail -r is not available on Linux, so we use either of them
-forgit::reverse_lines() { tac 2> /dev/null || tail -r; }
 
-forgit::previous_commit() {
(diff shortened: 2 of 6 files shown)


cjappl:
Wow! Very interesting. Will check this out early next week.

Definitely something I’ve considered in the past as well, and the fish and zsh/bash were inherently moving away from each other minorly Just due to little differences in implementation.

Will dig in more soon.

conf.d/bin/git-forgit
```diff
@@ -0,0 +1 @@
+../../bin/git-forgit
```
wren, line 1: I'm curious what the addition of this new file does. @carlfriedrich, can you explain?

carlfriedrich: @wren The fish plugin manager [fisher](https://github.com/jorgebucaran/fisher) only installs files in [these three folders](https://github.com/jorgebucaran/fisher#creating-a-plugin):

- `completions`
- `conf.d`
- `functions`

So symlinking `bin/git-forgit` to `conf.d/bin/git-forgit` was the only way I found to actually install the `git-forgit` script for fish.

@cjappl Maybe you know more about the fish plugin architecure than me and might have a better idea? Having a `bin` subdir in your `conf.d` directory surely seems a bit unorthodox. I found that it works, though, and couldn't come up with a cleaner idea.

carlfriedrich:
> Definitely something I’ve considered in the past as well, and the fish and zsh/bash were inherently moving away from each other minorly Just due to little differences in implementation.

@cjappl Yep, I noticed that, too. In this patchset, however, I took the zsh implementation only and completely threw away the fish implementation. So if you notice any regressions in fish compared to the former version, let's discuss them right away to get the most consistent and feature-rich version of forgit for both shells as a result!

conf.d/bin/git-forgit
cjappl: Great question...

I unfortunately have no knowledge of fisher/fish plugins. You may be able to `git blame` and hunt down the original person who implemented that and ask them. I didn't contribute to the plugin-ification of forgit

cjappl:
@carlfriedrich is this usable in it's current state? AKA if I download and use this branch, it should behave like normal with no modifications to my workflow?

If so, I'll make it my default branch for a bit, and if it works as intended I'm all for this change. No need to have such duplication as long as both shells will work as we like them today :)

carlfriedrich:
@cjappl Yes, it should work just like before indeed, except that you will have the zsh implementation from now on, so you might notice those slight differences you already talked about.

If you don't install forgit via a plugin manager, just make sure that you have `conf.d/bin/git-forgit` in addition to the plugin file itself.

cjappl:
Awesome! Alright, giving it a shot, will report back after giving it some soak time

cjappl:
```
gcb forMattoAES67Crash-publish
/Users/chrisapple/code/personal//forgit/conf.d/bin/git-forgit: line 16: realpath: command not found
```

First bug report, would you like me to capture these some way? I'm also happy to go back and dig in myself, but my cycles may be slow

carlfriedrich:
@cjappl Ah, didn't think about that there's no coreutils on Mac. For now, if you have homebrew you can install it:

```
brew install coreutils
```

I will come up with a portable solution for this as soon as I get to it, though.

cjappl:
Damn mac… always a pain to figure out what’s on which system.

Thanks! Just ping me if you find a cross platform solution and I can keep banging on it :)

carlfriedrich:
@cjappl I added a patch to this PR with a cross-platform solution for the path detection of `git-forgit`. Your error should be gone with this.

forgit.plugin.zsh
```diff
@@ -1,511 +1,112 @@
...
-    graph=--graph
-    [[ $FORGIT_LOG_GRAPH_ENABLE == false ]] && graph=
-    log_format=${FORGIT_GLO_FORMAT:-$forgit_log_format}
-    eval "git log $graph --color=always --format='$log_format' $* $forgit_emojify" |
-        FZF_DEFAULT_OPTS="$opts" fzf
-    fzf_exit_code=$?
-    # exit successfully on 130 (ctrl-c/esc)
-    [[ $fzf_exit_code == 130 ]] && return 0
-    return $fzf_exit_code
+    "$FORGIT_INSTALL_DIR/bin/git-forgit" log "$@"
```
wfxr, line 23: Give `"$FORGIT_INSTALL_DIR/bin/git-forgit"` a name like `$FORGIT_BIN`?

wfxr (approved):
LGTM. I noticed a slight delay when executing forgit commands after this patch. But it's not a big problem.

forgit.plugin.zsh
carlfriedrich: That's a legit suggestion, and I even already did this within `git-forgit`, so it totally makes sense to also do it in the plugin. I will add a patch.

carlfriedrich:
@wfxr Thanks for your review. I added a patch which implements your suggestion.

I also added another patch which makes the shell plugin return an error if the installation path cannot be determined (since then it's not usable).

I'm sorry about the delay this patch adds to your workflow. I guess that's the price of calling an external script instead of a pure shell function implementation. I'm surprised that you notice the difference, though. In my case it's still super fast, and I wouldn't have noticed any difference.

wfxr:
> I'm sorry about the delay this patch adds to your workflow. I guess that's the price of calling an external script instead of a pure shell function implementation. I'm surprised that you notice the difference, though. In my case it's still super fast, and I wouldn't have noticed any difference.

@carlfriedrich NVM. This delay is acceptable.

Some documents will be out of date after this pr merged. For example the following description maybe should probably be removed:
 
https://github.com/wfxr/forgit/blob/2872548075e63bc83a0b960e2813b16571998563/README.md?plain=1#L57-L65

carlfriedrich:
> Some documents will be out of date after this pr merged. For example the following description maybe should probably be removed:
> 
> https://github.com/wfxr/forgit/blob/2872548075e63bc83a0b960e2813b16571998563/README.md?plain=1#L57-L65

@wfxr Thanks for the hint. I will definitely go through the documentation and update it where necessary as soon as @cjappl has approved this as well.

cjappl:
Screen Shot 2022-10-24 at 9 45 57 AM

Apologies I'm just throwing bugs at you and not helping fix any of them! I have been slammed at work recently. Hopefully I'll be able to contribute soon and fix some of these things instead of just bringing them up :)

carlfriedrich:
@cjappl No worries, that's totally fine! I am happy to push this forward and your bug reports are very welcome.

I am not at a computer at the moment, but from what I can see in the screenshot your preview command seems to run fish code, which shouldn't be the case. Do you have a custom preview command defined?

carlfriedrich:
@cjappl I actually can't reproduce this on my machine in fish. Preview looks fine here:

grafik

I assume you either have a custom preview command configured, or there's another forgit version cluttering your shell environment. Have you tried starting a clean shell?

```
env -i HOME="$HOME" TERM="$TERM" fish
```

cjappl:
Sorry, haven’t forgotten about this! Work has been slammed. Will circle back soon :)

cjappl:
OK: Trying again and the previous issue is resolved, but a different one occurs (I made sure to completely kill my shell session, then come back "fresh".
Screen Shot 2022-11-10 at 8 48 25 AM


which corresponds to this line:
```

    preview="
        file=\$(echo {} | $extract)
        if (git status -s -- \$file | grep '^??') &>/dev/null; then  # diff with /dev/null for untracked files
            git diff --color=always --no-index -- /dev/null \$file | $_forgit_diff_pager | sed '2 s/added:/untracked:/'
        else
            git diff --color=always -- \$file | $_forgit_diff_pager
        fi"

```

I'm wondering if it always executes these commands in $SHELL? and we have to be smarter about using zsh/bash to run these commands?

I can confirm that changing that first `file=` to fish syntax `set file` at least changes the output, meaning that I was poking the right spot.

Screen Shot 2022-11-10 at 8 50 58 AM

Any thoughts on how we can massage this for folks who use fish as their native shell?

carlfriedrich:
@cjappl Thanks for your response. I could reproduce this with setting my ' $SHELL' variable to something other than bash. fzf indeed evaluates this variable to run the preview commands. Since all preview commands are implemented in bash now, we have to ensure that fzf uses bash to run these. I added a new commit which sets the `$SHELL` variable accordingly within forgit.

I am not sure if this might have side effects. It only changes the default shell for the forgit process, though, so I can't think of anything wrong with this, because I think it is only relevant for the previews. Do I overlook anything?

Anyway, in my case it works now with fish as the default shell. Can you check again?

cjappl:
Sweet! This one is mostly working for me. I'm going to stick on this branch in general for the near future as I test it out.

One thing I am noticing that I'm happy to submit a commit for (just want to make sure other folks are seeing it).

I am seeing that `$#` is constantly 1, even when I don't pass in args. This means that in functions like so:

```
# git add selector
_forgit_add() {
    # Add files if passed as arguments
    [[ $# -ne 0 ]] && git add "$@" && git status -su && return
```

I was always getting through this conditional, trying to `git add ""` and geting a fatal error:
```
> ga
fatal: empty string is not a valid pathspec. please use . instead if you meant to match all paths
Nothing to add.
```

However, changing all of these 
```
[[ $# -ne 0 ]]
```

to 
```
[[ $# -ne 1 ]]
```

fixed it. Is that true for anyone else?? You can replicate by putting an echo in the top of the function, and then calling `ga` with no parameters.

```
# git add selector
_forgit_add() {
    # Add files if passed as arguments
    echo $#
    [[ $# -ne 1 ]] && git add "$@" && git status -su && return


... outputs

> ga
1
Nothing to add.
```

cjappl:
Overall, this version is the first version really working for me, can't wait to stress test it more and hopefully give thumbs up soon!! Great work @carlfriedrich , really appreciate you putting this together. 🎉

carlfriedrich:
@cjappl Thanks for your good testing and feedback.

> I am seeing that $# is constantly 1, even when I don't pass in args.

I was able to reproduce this. It happened in fish only, though, so I could track it down to the fish plugin code and already pushed a fix for it to this branch. It occured because I did not know that [passing `$argv` in fish does not need double quotes](https://stackoverflow.com/a/42379014/3018229). If you update the branch the error should be gone.

> Another small, unimportant note I'm realizing on first blush is that fish users MUST export environment variables, such that bash can use them

Yes, that's indeed true for all shells, I forgot to mention this in the PR description. I will update the documentation accordingly.
Maybe we can export all relevant variables (if they are set) in the plugin code so that users don't need to update their configuration when switching to the new forgit version? @wfxr Do you think that's a useful idea?

wfxr:
> @cjappl @wfxr I added code for automatic exporting all `FORGIT_` variables to both shell plugins for backwards compatibility. This way users do not have to adjust their configuration as soon as this gets merged. This seems best for me. What do you think? Do you find it problematic implicitly exporting variables into the current shell?

LGTM.

Maybe we can detect whether these `FORGIT_` variables defined by `export` or by `set`. If by `set`, display a warning message to help users update their config. Then we can remove these implicitly exporting logic in the future. (maybe 1 or 3 months?)

> And @wfxr can you maybe check your workflow on the current branch again? Since I added some modifications for fish compatibility I'd like to make sure that everything still works with zsh.

I tested and found nothing breaks with zsh.

cjappl:
Did a quick edit - in fish you don't need `read -r` (and in fact the option doesn't exist at least on my implementation).

See docs https://fishshell.com/docs/current/fish_for_bash_users.html#subshells

@wfxr 's approach seems reasonable! I think this version we have in the code works well too.

I'm not sure how folks feel about breaking backwards compatibility - but we could just consider updating the documentation and letting folks figure it out. I was able to sort it out in 30 seconds - that may be easier than any kind of code solution to this

carlfriedrich:
> Maybe we can detect whether these `FORGIT_` variables defined by `export` or by `set`. If by `set`, display a warning message to help users update their config. Then we can remove these implicitly exporting logic in the future. (maybe 1 or 3 months?)

@wfxr Great idea, thanks for the suggestion! I updated the commit, so that a warning is displayed now if anything has to be exported.

> Did a quick edit - in fish you don't need `read -r` (and in fact the option doesn't exist at least on my implementation).

@cjappl Thanks for the hint! I squashed this fix into the latest commit.

> I'm not sure how folks feel about breaking backwards compatibility - but we could just consider updating the documentation and letting folks figure it out. I was able to sort it out in 30 seconds - that may be easier than any kind of code solution to this

I will update the documentation as well, but I think a warning is indeed a useful additional way of telling the people that we're about to change something.

Can you both check if the warnings are displayed correctly?

wfxr:
> Can you both check if the warnings are displayed correctly?

@carlfriedrich LGTM.

cjappl:
After sleeping on this last night:

I think we should merge this if everyone is OK with it! I’ve found the basic functionality works as intended in fish, and we can just file tickets / fixes as things come up.

I don’t think we should halt this one going in for much longer - seeing as it’s a big architectural change I think we should just take the plunge and we can sort out the issues as they happen?

Thoughts?

carlfriedrich:
@cjappl Thanks for your feedback! For me that sounds good, and I actually had similar thoughts during the last days. :-)

Before merging this I will update the documentation to match the new architecture. It would be great if you both could review those as well, but I will give you a ping as soon as I pushed those changes.

cjappl:
Thanks will do.

carlfriedrich:
So I updated the documentation and finalized the branch for merging. As I already announced in the PR description, I removed the patches for using `gd` in `glo` and will merge these afterwards in #219.

@wfxr @cjappl Can you verify the documentation changes, check if I overlooked anything, and do one last functional verification in your environment? Thanks a lot!

cjappl (approved):

cjappl:
Everything is operational on a basic pass on my end! After merging, I will file any tickets, or fix issues as I seem them! 

great work 👍

cjappl:
I think lets get this submitted ASAP - more PRs are coming in (and there are a few things I want to do) but we should wait until this is merged so we don't have to duplicate the effort back to this implementation

carlfriedrich:
@cjappl You are totally right. I actually wanted to wait for @wfxr to give his final "go", but since everything seems to be working in our environments, I will merge this now. Thanks to you both for your support on bringing this into master! ❤️

Issue #233, opened 2022-09-12
Have a proper versioning system


arihant2math:
## Check list

- [x] I have read through the [README](https://github.com/wfxr/forgit/blob/master/README.md)
- [x] I have the latest version of forgit
- [x] I have searched through the existing issues

## Environment info

- OS
    - [x] Linux
    - [x] Mac OS X
    - [x] Windows
    - [x] Others:
- Shell
    - [x] bash
    - [x] zsh
    - [x] fish

## Problem / Steps to reproduce
forgit is not versioned. This is a problem because brew (see #194) depends on forgit being versioned.

carlfriedrich:
I'd like to get this rolling. IMO the only thing to do is start creating tags.

Some questions to discuss @wfxr @cjappl:

1. Is it okay for you if I do this? I am volunteering to maintain the release process, so no work on your side, if you are fine with that. @wfxr If you as the initial creator of forgit want to maintain releases, though, I would be totally fine with that as well.

2. Do you have an opinion on whether to use [semantic versioning](https://semver.org/) or [calendar versioning](https://calver.org/)? Or do you have another preferred versioning scheme? IMO semantic versioning (major.minor.fix) seems legit for this project. We probably wouldn't ever use the fix revision, though.

3. Should we define a version for the current master already? If you prefer semantic versioning as well, do you think [we are on 1.0.0 already](https://semver.org/#how-do-i-know-when-to-release-100)? Since forgit is out there in the public and used by lots of people, and we try to keep changes backwards-compatible all the time, I would say yes. So we could define a 1.0.0 version. @wfxr Maybe there is a former version of forgit which you would think of as the first stable version? Or we could just tag the current master with 1.0.0?

Let me know what you think.

wfxr:
@carlfriedrich

1. Totally fine. And I appreciate all your work for `forgit`.

2. I think [calendar versioning](https://calver.org/) is more appreciate for a project like `forgit`. We can release new versions periodically without having to think too much about how to define a new version tag or when to do a release. The release processes may be fully automated through the github workflow.

3. This seems not be a problem if we choose [calendar versioning](https://calver.org/)?

cjappl:
All of the above WFM! 🚀

carlfriedrich:
@wfxr 

>     1. Totally fine. And I appreciate all your work for `forgit`.

Great, thank you!

[...]

PR #248, opened 2022-11-24
Meta: GitHub tag action: add commit key

carlfriedrich:
The COMMIT_KEY secret contains a private SSH key. The associated public key has been added as a deploy key in the GitHub project. See: https://docs.github.com/en/developers/overview/managing-deploy-keys#deploy-keys

This is necessary to make commits done by github-actions able to trigger further github-actions. See:
https://stackoverflow.com/q/60418323/3018229.

## Check list

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [ ] zsh
    - [ ] fish
- OS
    - [ ] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/.github/workflows/tag.yaml
+++ b/.github/workflows/tag.yaml
@@ -15,6 +15,12 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
+          # The COMMIT_KEY secret contains a private SSH key. The associated public key
+          # has been added as a deploy key in the GitHub project. See:
+          # https://docs.github.com/en/developers/overview/managing-deploy-keys#deploy-keys
+          # This is necessary to make commits done by github-actions able to trigger
+          # further github-actions. See: https://stackoverflow.com/q/60418323/3018229.
+          ssh-key: "${{secrets.COMMIT_KEY}}"
 
       - name: Create Tag
         run: .github/scripts/tag.sh


carlfriedrich:
In #233 we agreed on this release workflow:

1. Tag creation (triggered on every 1st day of the month)
    - Check if current master is newer than latest tag. If yes:
    - Create a release tag

2. Release creation (triggered on new tags)
    - Create a GitHub release
    - Generate a changelog

Unfortunately GitHub actions are [not able to trigger other GitHub actions automatically by default](https://stackoverflow.com/questions/60418323/triggering-a-new-workflow-from-another-workflow). So when the first action creates a tag, the second action is not triggered. This happens only if a user pushes the tag manually.

In order to make this work, we need to supply a [deploy key](https://docs.github.com/en/developers/overview/managing-deploy-keys#deploy-keys) to the repository. This is basically an SSH keypair, of which the private key has to be added as a [repository secret](https://docs.github.com/en/actions/security-guides/encrypted-secrets#creating-encrypted-secrets-for-a-repository) called `COMMIT_KEY`, and the public key has to be added as a [deploy key](https://docs.github.com/en/developers/overview/managing-deploy-keys#setup-2) with write access.

Both have to be configured in the repository settings, to which I don't have access. @wfxr Is there a possibility to give me access to these settings? Could be temporarily, if that's okay for you. Or alternatively could you generate the keypair and set the keys up as described?

Sorry for the circumstances, I wasn't aware of this before. :-/

After the keys have been set up, this PR can be merged.

carlfriedrich:
@wfxr Did you get my ping here? We actually need the SSH keypair in the project for the automatic releases to work.

With the beginning of December today we could already witness that the release workflow does not work without it. While the `tag` workflow [successfully ran](https://github.com/wfxr/forgit/actions/runs/3588253849/jobs/6039448935) and created the [22.12.0 tag](https://github.com/wfxr/forgit/releases/tag/22.12.0), this tag creation did not trigger the `release` workflow because [GitHub prevents this](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow) when using the default `GITHUB_TOKEN`.

wfxr:
@carlfriedrich Sorry for the very late reply. I have a bad cold these days.

> Is there a possibility to give me access to these settings? Could be temporarily, if that's okay for you. Or alternatively could you generate the keypair and set the keys up as described?

Seems there's nowhere to give you access to these settings. I have tried to set up the required keys. Let's see if it works.

carlfriedrich:
> @carlfriedrich Sorry for the very late reply. I have a bad cold these days.

@wfxr No worries, get well soon!
 
> Seems there's nowhere to give you access to these settings. I have tried to set up the required keys. Let's see if it works.

Great, we'll see in January then. :-) I will re-create the 22.12.1 tag manually to have a proper December release.

Issue #85, opened 2020-04-16
glo: doesn't display proper diffs with commit history slice


NullSense:
## Environment info

- OS
    - [x] Linux
- Shell
    - [x] zsh

## Problem / Steps to reproduce
`glo first_ref..last_ref`

e.g.: `glo 188a849b..73e5ca0d`

## Expected behavior:
Shows the commits only including that slice, and each diff is separate.

## Actual behavior:
Shows the commits only including that slice, but if you hover over any of them, it shows the diff of the whole slice itself, not each separate commit.

wfxr:
Hi @NullSense As `REDME.md` descripted, `glo` accept **path** arguments to restrain the items listed in fzf (eg, glo main.go test.go). It's just a coincidence that commit ref can be used either.

But commit ref as the parameter for `glo` should be a good improvement. I will consider implementing it.

wfxr:
@NullSense Implemented. Now can filter both refs and files like this:
`glo 188a849b..73e5ca0d -- main.go`

wfxr:
@cjappl Ping...
This feature can be ported to `fish` version.

cjappl:
Yep! I’m on it, sorry for the delay I should have told you I saw it :)

I’ll be working on it tomorrow, just had a weekend off.

Cheers! Hope you are staying safe and healthy.

wfxr:
@cjappl That's all right.
I'm fine. And hope you are safe and healthy too friend!

cjappl:
Good to go on fish :) Thanks for reporting @NullSense

PR #249, opened 2022-11-26
fix: forgit::log can not match files

ZenLian:
## Check list

## Description

**Bug**: `glo -- files` should show specified files only, but instead it shows all changed files.
**Fix**: There is a redundent blank in sed regex.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [x] bash
    - [x] zsh
    - [ ] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -70,7 +70,7 @@ _forgit_preview_context=${FORGIT_PREVIEW_CONTEXT:-3}
 _forgit_log() {
     _forgit_inside_work_tree || return 1
     local opts graph files log_format preview_cmd enter_cmd
-    files=$(sed -nE 's/.* -- (.*)/\1/p' <<< "$*") # extract files parameters for `git show` command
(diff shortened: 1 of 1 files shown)


carlfriedrich:
@ZenLian Thanks for your PR! We just merged a significant change which conflicts with your patch. Can you rebase this on current master?

ZenLian:
@carlfriedrich Rebase done. Sorry for my late reply.

carlfriedrich (approved):
@ZenLian Thank you! I can confirm that your patch actually changes the behavior from "show all files in the diff" to "show only given files in the diff", and I think that the last one should be correct.

Since @wfxr originally implemented this in #85 I would leave the final approval to him.

cjappl:
Seeing as @wfxr has been busy recently, I think we should consider merging this! What do you think @carlfriedrich ?

PR #251, opened 2022-11-26
Feature: Added git stash push selector

### LINKED ISSUE #243: Git stash interactive
## Check list

- [X] I have read through the [README](https://github.com/wfxr/forgit/blob/master/README.md)
- [X] I have the latest version of forgit
- [X] I have searched through the existing issues

## Environment info

- OS
    - [X] Linux
    - [X] Mac OS X
    - [X] Windows
    - [X] Others:
- Shell
    - [X] bash
    - [X] zsh
    - [X] fish

More of a feature request then an issue. Does it make sense to have an interactive stash?
I can of course interactively add the things i don't want in my stash, `stash save --keep-index` then reset everything.
Just wondering if other had that need? Could not find anything in the current open/closed issues

sandr01d:
## Check list

## Description

Added an interactive git stash push selector, that works similar to `forgit::add`. The alias for it is `gsp`. I thought about passing arguments to the command to be able to use e.g. `-m "name"` to be able to name the stash entry, but had no luck implementing this so far. If anyone has an idea, let me know and I'll update my PR. From my perspective we could merge this PR without this additional feature. 

This closes #243 

https://user-images.githubusercontent.com/88739791/204108329-c8846a1b-3dbc-411a-be13-873e2d9bf39e.mp4

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Breaking change
- [X] Documentation change

## Test environment

- Shell
    - [X] bash
    - [X] zsh
    - [X] fish
- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -204,6 +204,30 @@ _forgit_stash_show() {
     return $fzf_exit_code
 }
 
+# git stash push selector
+_forgit_stash_push() {
+    _forgit_inside_work_tree || return 1
+    # Stash files if passed as arguments
+    [[ $# -ne 0 ]] && git stash push -u "$@" && return
+    local opts preview files
+    opts="
+        $FORGIT_FZF_DEFAULT_OPTS
+        -m
+        $FORGIT_STASH_PUSH_OPTS
+    "
+    preview="
+        if (git ls-files {} --error-unmatch) &> /dev/null; then
+            git diff --color=always {} | $_forgit_diff_pager
+        else
+            git diff --color=always /dev/null {} | $_forgit_diff_pager
+        fi
+    "
+    # Show both modified and untracked files
+    files=$(git ls-files --modified --others | FZF_DEFAULT_OPTS="$opts" fzf --preview="$preview")
+    [[ -z "$files" ]] && return 1
+    echo "${files[@]}" | tr '\n' '\0' | git stash push -u --pathspec-file-nul --pathspec-from-file -
+}
+
 # git clean selector
 _forgit_clean() {
     _forgit_inside_work_tree || return 1
@@ -542,6 +566,7 @@ valid_commands=(
     "reset_head"
     "revert_commit"
     "stash_show"
+    "stash_push"
 )
 
 cmd="$1"
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -65,6 +65,7 @@ _forgit_log_format=${FORGIT_LOG_FORMAT:-%C(auto)%h%d %s %C(black)%C(bold)%cr%Cre
 _forgit_log_preview_options="--graph --pretty=format:'$_forgit_log_format' --color=always --abbrev-commit --date=relative"
 _forgit_fullscreen_context=${FORGIT_FULLSCREEN_CONTEXT:-10}
 _forgit_preview_context=${FORGIT_PREVIEW_CONTEXT:-3}
+_forgit_is_file_tracked="(git ls-files {} --error-unmatch) &> /dev/null"
 
 # git commit viewer
 _forgit_log() {
@@ -216,7 +217,7 @@ _forgit_stash_push() {
         $FORGIT_STASH_PUSH_OPTS
     "
     preview="
-        if (git ls-files {} --error-unmatch) &> /dev/null; then
+        if $_forgit_is_file_tracked; then
             git diff --color=always {} | $_forgit_diff_pager
         else
             git diff --color=always /dev/null {} | $_forgit_diff_pager
@@ -488,7 +489,7 @@ _forgit_blame() {
     "
     flags=$(git rev-parse --flags "$@")
     preview="
-        if (git ls-files {} --error-unmatch) &>/dev/null; then
+        if $_forgit_is_file_tracked; then
             git blame {} --date=short $flags | $_forgit_blame_pager
         else
             echo File not tracked
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -208,6 +208,21 @@ _forgit_stash_show() {
 # git stash push selector
 _forgit_stash_push() {
     _forgit_inside_work_tree || return 1
+    local msg args
+    args=( "$@" )
+    while (( "$#" )); do
+        case "$1" in
+            # allow message as argument
+            -m|--message)
+            msg="$2"
+            shift 2
+            ;;
+            # ignore -u as it's used implicitly
+            -u|--include-untracked) shift ;;
+            # pass to git directly when encountering anything else
+            *) git stash push "${args[@]}" && return $?
+        esac
+    done
     local opts preview files
     opts="
         $FORGIT_FZF_DEFAULT_OPTS
@@ -224,7 +239,7 @@ _forgit_stash_push() {
     # Show both modified and untracked files
     files=$(git ls-files --exclude-standard --modified --others | FZF_DEFAULT_OPTS="$opts" fzf --preview="$preview")
     [[ -z "$files" ]] && return 1
-    echo "${files[@]}" | tr '\n' '\0' | git stash push ${1:+-m "$1"} -u --pathspec-file-nul --pathspec-from-file -
+    echo "${files[@]}" | tr '\n' '\0' | git stash push ${msg:+-m "$msg"} -u --pathspec-file-nul --pathspec-from-file -
 }
 
 # git clean selector
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -224,7 +224,7 @@ _forgit_stash_push() {
         fi
     "
     # Show both modified and untracked files
-    files=$(git ls-files --modified --others | FZF_DEFAULT_OPTS="$opts" fzf --preview="$preview")
+    files=$(git ls-files --exclude-standard --modified --others | FZF_DEFAULT_OPTS="$opts" fzf --preview="$preview")
     [[ -z "$files" ]] && return 1
     echo "${files[@]}" | tr '\n' '\0' | git stash push -u --pathspec-file-nul --pathspec-from-file -
 }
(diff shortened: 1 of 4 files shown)


carlfriedrich:
@sandr01d Thanks for your submission! This actually works fine for me. I can imagine adding your proposed `-m "message"` option later, but from my perspective we could merge this right away.

@cjappl @wfxr Can you check this in your environments as well?

sandr01d:
> @sandr01d Thanks for your submission! This actually works fine for me. I can imagine adding your proposed `-m "message"` option later, but from my perspective we could merge this right away.
> 
> @cjappl @wfxr Can you check this in your environments as well?

Thanks. Great work with #241 btw., this makes working so much easier!

cjappl:
Cool, good idea. Going to try it out locally. Also THANK YOU for the video describing the functionality, really helps in evaluating if it's working as intended 🎉 

I would be in favor of adding the message in the future, as I constantly do

`git stash save "something"` but agreed it doesn't have to be done now.

When you do it, you don't have to do the -m flag, I would just do

```
gsp "My Message"
```

bin/git-forgit
```diff
@@ -204,6 +205,30 @@ _forgit_stash_show() {
...
+    "
+    preview="
+        if $_forgit_is_file_tracked; then
+            git diff --color=always {} | $_forgit_diff_pager
+        else
+            git diff --color=always /dev/null {} | $_forgit_diff_pager
+        fi
+    "
+    # Show both modified and untracked files
+    files=$(git ls-files --modified --others | FZF_DEFAULT_OPTS="$opts" fzf --preview="$preview")
```
cjappl, line 227: I think what you may want is

```
    files=$(git ls-files --exclude-standard --modified --others | FZF_DEFAULT_OPTS="$opts" fzf --preview="$preview")

```

--exclude-standard (as far as I understand) makes it so files in your .gitignore don't show up. The first time I ran this I had a ton of my build folders/object files etc show up in the output!

cjappl (changes requested):

cjappl:
A couple more thoughts after playing with this for a bit:

1. I think adding the message as the first optional argument would be great. Can you investigate how much work that would be to add it? And add it if it is straight forward?
2. I can see as a next feature be a `gsa` "git stash apply" that can interactively apply the stash you want, which would be SO USEFUL! (no action here, just giving a huge +1 if anyone is inspired to create this)

bin/git-forgit
sandr01d: Good catch! I've fixed that now.

sandr01d:
> A couple more thoughts after playing with this for a bit:
> 
>     1. I think adding the message as the first optional argument would be great. Can you investigate how much work that would be to add it? And add it if it is straight forward?
> 
>     2. I can see as a next feature be a `gsa` "git stash apply" that can interactively apply the stash you want, which would be SO USEFUL! (no action here, just giving a huge +1 if anyone is inspired to create this)

1. I agree. I just tested this out and adding this is trivial (in fact I have a working solution ready). We'd have to remove the ability to stash files that were added as arguments, essentially this line:

```sh
 [[ $# -ne 0 ]] && git stash push -u "$@" && return
```
as there would be no way to determine whether the provided arguments are intended as message or files. I personally would prefer having the message, because you could still use an alias for the other behavior. What do you think?

2. Totally agree, that would be nice! We could add a command for `git stash pop` as well, that could share most of the code from `git stash apply`. I'll have a look at this when I find the time for it. This week is pretty packed for me, so it might take a while.

carlfriedrich:
>     2. I can see as a next feature be a `gsa` "git stash apply" that can interactively apply the stash you want, which would be SO USEFUL! (no action here, just giving a huge +1 if anyone is inspired to create this)

This is actually something I have implemented in my environment as a keybinding within `gss` (as well as another binding for dropping the stash), which I think is much more useful than having to run different commands for each of these actions. I am not at my computer right now, but if you find that interesting as well, I could post more details on this in the next days.

sandr01d:
> > 2. I can see as a next feature be a `gsa` "git stash apply" that can interactively apply the stash you want, which would be SO USEFUL! (no action here, just giving a huge +1 if anyone is inspired to create this)
> 
> This is actually something I have implemented in my environment as a keybinding within `gss` (as well as another binding for dropping the stash), which I think is much more useful than having to run different commands for each of these actions. I am not at my computer right now, but if you find that interesting as well, I could post more details on this in the next days.

I would definitely be interested in that.

cjappl:
> as there would be no way to determine whether the provided arguments are intended as message or files. I personally would prefer having the message, because you could still use an alias for the other behavior. What do you think?

I would personally prefer the message over the files. What do you think @carlfriedrich ?

> This is actually something I have implemented in my environment as a keybinding within gss (as well as another binding for dropping the stash), which I think is much more useful than having to run different commands for each of these actions. I am not at my computer right now, but if you find that interesting as well, I could post more details on this in the next days.

Very interesting! I would definitely be keen on seeing that.

sandr01d:
Alright, I've pushed support for adding an optional message like `gsp "message"`. Would be great if you could give this a quick test in your environments again.

sandr01d:
No idea why @wfxr got removed from the reviewers. I can't add him again...

cjappl:
Checking it out as my main branch, so far so good.

carlfriedrich (changes requested):
@sandr01d @cjappl

Sorry for coming back to this so late.

After thinking about this a bit more, I am not sure how I like the message as a positional argument, given that we don't do anything similar elsewhere in forgit.
On other commands we allow bypassing the interactive fzf selector, so that we can use forgit commands just like git commands, e.g. `git forgit checkout_branch mybranch` calls `git checkout mybranch` directly, without opening the fzf dialog.
Similarly, I would expect `git forgit stash myfile` to call `git stash myfile` directly.

The current implementation in this PR would still open the fzf dialog and then stash the files with the message `myfile`. While I definitely see that it's the most easy solution to provide a message, it doesn't feel coherent within forgit, though.

I have two suggestions to this:

1. Use `-m mymessage` instead, just like we already discussed above. We would still have to handle this independently of other positional arguments in order to get the "standard" forgit behavior for arguments as well.

2. Query the message interactively after selecting the files to stash, e.g. using `read -p "Optional stash message: " stash_msg`. Then append `-m $stash_msg` to the `git stash` command.
This would be something new to forgit, but I would consider this most user-friendly. However, I know that @wfxr likes to keep things simple, and this suggestion might be out of scope for forgit, so maybe option 1 might be the most cohesive one.

Let me know what you all think.

cjappl:
I would be ok dropping support for the message here. I’ve been going back and forth on it.

#2 is interesting though, I could be down with that. That seems like the best solution so far.

After playing with it for a bit the ergonomics of putting the message in before you stash or look at the files is strange.

carlfriedrich:
> After playing with it for a bit the ergonomics of putting the message in before you stash or look at the files is strange.

Yes, I actually agree with that.

sandr01d:
GitHub keeps removing the requested reviews for me, sorry! Could somebody re-assign @wfxr ?

> @sandr01d @cjappl
> 
> Sorry for coming back to this so late.
> 
> After thinking about this a bit more, I am not sure how I like the message as a positional argument, given that we don't do anything similar elsewhere in forgit. On other commands we allow bypassing the interactive fzf selector, so that we can use forgit commands just like git commands, e.g. `git forgit checkout_branch mybranch` calls `git checkout mybranch` directly, without opening the fzf dialog. Similarly, I would expect `git forgit stash myfile` to call `git stash push myfile` directly.
> 
> The current implementation in this PR would still open the fzf dialog and then stash the files with the message `myfile`. While I definitely see that it's the most easy solution to provide a message, it doesn't feel coherent within forgit, though.
> 
> I have two suggestions to this:
> 
>     1. Use `-m mymessage` instead, just like we already discussed above. We would still have to handle this independently of other positional arguments in order to get the "standard" forgit behavior for arguments as well.
> 
>     2. Query the message interactively after selecting the files to stash, e.g. using `read -p "Optional stash message: " stash_msg`. Then append `-m $stash_msg` to the `git stash` command.
>        This would be something new to forgit, but I would consider this most user-friendly. However, I know that @wfxr likes to keep things simple, and this suggestion might be out of scope for forgit, so maybe option 1 might be the most cohesive one.
> 
> 
> Let me know what you all think.

I see your point there. I've tested this feature throughout the week and I also don't like that this behaves differently than everything else in forgit. Do I understand this correctly, that with 2. we would always prompt the user for a message? If so, I would favor 1. I often stash my changes for a very short period of time and don't want to add a message for that. In this kind of situations this could be annoying.

carlfriedrich:
@sandr01d

> I see your point there. I've tested this feature throughout the week and I also don't like that this behaves differently than everything else in forgit. Do I understand this correctly, that with 2. we would always prompt the user for a message? If so, I would favor 1. I often stash my changes for a very short period of time and don't want to add a message for that. In this kind of situations this could be annoying.

Same here, totally agree. So let's move on with option 1 then.

You can still name your stash afterwards if you like to keep it, e.g. using this alias which I got [from StackOverflow](https://stackoverflow.com/questions/25931026/how-can-i-rename-a-git-stash):

```shell
git config alias.stash-rename '!_() { rev=$(git rev-parse $1) && git stash drop $1 || exit 1 ; git stash store -m "$2" $rev; }; _'
```

sandr01d:
I've implemented option 1 now. I find using this way more intuitive now. Let me know what you think :)

cjappl (approved):
WFM: Will leave to @wfxr or @carlfriedrich to give final approval

carlfriedrich:
@sandr01d @cjappl Thanks a lot, works great!

PR #246, opened 2022-11-16
Automatic releases

carlfriedrich:
This PR adds a GitHub action for automatically creating a GitHub release for each tag (refs #233).

The action is executed on every tag and creates a GitHub release. The release contains a .tar.gz asset and a changelog consisting of all commit messages since the previous tag. Commits with a commit message starting with "Meta" are excluded from the changelog.

I tested this action on my forgit fork and you can see a generated changelog here:

https://github.com/carlfriedrich/forgit/releases

## Check list

- [x] I have performed a self-review of my code
- [x] I have commented my code in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [ ] zsh
    - [ ] fish
- OS
    - [ ] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

PR #252, opened 2022-11-26
Provide an input branch with gcp again

cjappl:
Noticed there was a change recently that broke the ability to pass in an input branch to `gcp`.

This change is the following:

- Fix faulty logic that just led to a `git checkout -b` being called when you pass in an argument to `gcp`
- If an argument was passed in, check if it is a valid branch, if so, set the `input_branch` var
- If `input_branch` is set, use it to do the cherry pick, as opposed to the first FZF selection.

PLEASE HELP ME REVIEW THIS, @carlfriedrich and @wfxr. I am such a bash syntax newbie, I don't know if I'm writing bad code 😄 

## Check list

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [x] bash
    - [x] zsh
    - [x] fish
- OS
    - [ ] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -249,7 +249,7 @@ _forgit_cherry_pick_from_branch() {
     _forgit_inside_work_tree || return 1
     local cmd preview opts branch exitval input_branch args
     args=("$@")
-    if [[ $# -ne 0 ]] && [[ $(git rev-parse --verify ${args[0]} 2>/dev/null) ]]; then
+    if [[ $# -ne 0 ]] && [[ $(git rev-parse --verify "${args[0]}" 2>/dev/null) ]]; then
         input_branch=${args[0]}
     fi
     cmd="git branch --color=always --all | LC_ALL=C sort -k1.1,1.1 -rs"
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -249,7 +249,7 @@ _forgit_cherry_pick_from_branch() {
     _forgit_inside_work_tree || return 1
     local cmd preview opts branch exitval input_branch args
     args=("$@")
-    if [[ $# -ne 0 ]] && [[ $(git rev-parse --verify "${args[0]}" 2>/dev/null) ]]; then
+    if [[ $# -ne 0 ]]; then
         input_branch=${args[0]}
     fi
     cmd="git branch --color=always --all | LC_ALL=C sort -k1.1,1.1 -rs"
@@ -276,7 +276,7 @@ _forgit_cherry_pick_from_branch() {
         _forgit_cherry_pick "$branch"
 
         exitval=$?
-        [[ $exitval != 130 ]] && return $exitval
+        [[ $exitval != 130 ]] || [[ $# -ne 0 ]] && return $exitval
     done
 }
 
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -247,10 +247,10 @@ _forgit_cherry_pick() {
 
 _forgit_cherry_pick_from_branch() {
     _forgit_inside_work_tree || return 1
-    local cmd preview opts branch exitval input_branch
-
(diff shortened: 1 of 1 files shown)


cjappl:
also, if someone could pull this down and test it on their machine, that would also be great. Again, very new with bash syntax so I want to get used to it, but I'm nervous about mistakes :)

cjappl:
Test cases:

In a repo with a `main` branch and a `feature` branch:

1. Checkout `feature`
2. `gcp main` - should result in a list of commits from main, you should be able to tab select the commits to cherry pick them
3. `gcp` - should come up with a list of all the branches in the repo. selecting one of them will bring up a list of commits from that branch
4. `gcp someBranchThatDoesntExist` should behave like 3, just ignore the bad input (should we error out here??)

carlfriedrich:
@cjappl Thanks for your patch. I actually don't know why the heck I have added this `git checkout` line?! I will take a look into this in the next days. :-)

cjappl:
Thanks!! Appreciate it :)

bin/git-forgit
```diff
@@ -247,8 +247,11 @@ _forgit_cherry_pick() {
 
 _forgit_cherry_pick_from_branch() {
     _forgit_inside_work_tree || return 1
-    [[ $# -ne 0 ]] && { git checkout -b "$@"; return $?; }
-    local cmd preview opts branch exitval
+    local cmd preview opts branch exitval input_branch args
+    args=("$@")
+    if [[ $# -ne 0 ]] && [[ $(git rev-parse --verify "${args[0]}" 2>/dev/null) ]]; then
```
carlfriedrich, line 252: I don't think that this function needs to check if the branch exists. In fact I find it even confusing when forgit silently ignores my argument if it does not exist. So I would suggest to remove the second half of the if-condition here and leave the error handling to the actual `git` git call.

bin/git-forgit
```diff
@@ -261,7 +264,13 @@ _forgit_cherry_pick_from_branch() {
...
+        if [[ -z $input_branch ]]; then
+            branch="$(eval "$cmd" | FZF_DEFAULT_OPTS="$opts" fzf | awk '{print $1}')"
+        else
+            branch=$input_branch
+        fi
+
+        unset input_branch
         [[ -z "$branch" ]] && return 1
 
         _forgit_cherry_pick "$branch"
```
carlfriedrich, line 276: We should somehow bypass the while loop here when there is an argument given. Otherwise quitting the fzf dialog via `ctrl+c` shows the branch selection (which hasn't been shown before), while I would expect it to return to the command line directly.

I would suggest adding another condition to the last line of the loop which will return the exitval if the branch was given via argument.

carlfriedrich (changes requested):
@cjappl I took a look into this and it basically works for me, good work! Just two small suggested changes that would improve the user experience IMO.

cjappl:
Great suggestions @carlfriedrich ! Let me know if that fixes those up. I have tested and confirmed the functionality still works as intended. Now with less annoying "stuck in the loop" when you give an input branch. Definitely a good improvement.

carlfriedrich (approved):
@cjappl Great, works fine for me now! Just one more small comment: Since #246 we're creating automatic releases with a changelog from the commit history now, so I would like to encourage you to squash the commits on this branch into a single one with a proper commit message. Thanks a lot!

Leaving final approval to @wfxr.

cjappl:
Sounds good @carlfriedrich ! Will "Squash and Merge" do what you're wanting? That's what I usually do anyway.

carlfriedrich:
@cjappl Yep, that's fine! The new commit message for the squash commit will appear in the changelog then.

Issue #257, opened 2022-12-06
Backwards compatibility check does not check for declare


gbtec-markusmeier:
## Check list

- [x] I have read through the [README](https://github.com/wfxr/forgit/blob/master/README.md)
- [x] I have the latest version of forgit
- [x] I have searched through the existing issues

## Environment info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:
- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish

## Problem / Steps to reproduce

When loading forgit, backwards compatilibility checks reports variables not being exported although they are.

Reason is that bash's export returns `declare -x FORGIT_PAGER="less"` and not `FORGIT_PAGER="less"`.

We should additionally check for `declare -x` in https://github.com/wfxr/forgit/blob/master/forgit.plugin.zsh#L25.

```bash
$ bash --version
GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu)
```

Issue #256, opened 2022-12-06
Backwards compatibility check reports functions and not only vars


gbtec-markusmeier:
## Check list

- [x] I have read through the [README](https://github.com/wfxr/forgit/blob/master/README.md)
- [x] I have the latest version of forgit
- [x] I have searched through the existing issues

## Environment info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:
- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish

### Problem
Forgit detects variables and functions in bash instead of only variables. See:
https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html

### Reproduction

Write a function that sets a `FORGIT` environment variable and then source forgit:
```bash
load_forgit() {
    export FORGIT_PAGER=less
    source "${HOME}/bin/forgit/forgit.plugin.sh"
}

load_forgit
```

### Result

```
[Warn] Config options have to be exported in future versions of forgit.
[Warn] Please update your config accordingly:
[Warn]   export FORGIT_COPY_CMD
[Warn]   export FORGIT_DIFF_FZF_OPTS
[Warn]   export FORGIT_LOG_FZF_OPTS
[Warn]   export FORGIT_PAGER
[Warn]   export FORGIT_STASH_FZF_OPTS
[Warn]   export export FORGIT_PAGER
bash: export: `export FORGIT_PAGER': not a valid identifier
```

### Expectation

The test in https://github.com/wfxr/forgit/blob/master/forgit.plugin.zsh#L24 should correctly report only variables, not strings in functions.

### Deeper analysis

Bash's `set` returns (amongst others) the above defined function:

```bash
load_forgit ()
{
    export FORGIT_PAGER=less;
    source "${HOME}/bin/forgit/forgit.plugin.sh"
}
```

When grepping the result, line `    export FORGIT_PAGER=less;` is processed.

### Solution

Detect only variables in bash without functions:

https://unix.stackexchange.com/questions/176001/how-can-i-list-all-shell-variables

Note that  `set -o posix ; set` is only available in bash, not in zsh, so we would need to distinguish that in https://github.com/wfxr/forgit/blob/master/forgit.plugin.zsh#L24

PR #258, opened 2022-12-06
fix: compatibility check not working in bash

gbtec-markusmeier:
fixes https://github.com/wfxr/forgit/issues/256
fixes https://github.com/wfxr/forgit/issues/257

## Check list

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [x] bash
    - [x] zsh
    - [ ] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

Fixes:
* https://github.com/wfxr/forgit/issues/256
* https://github.com/wfxr/forgit/issues/257

### DIFF
--- a/forgit.plugin.zsh
+++ b/forgit.plugin.zsh
@@ -21,8 +21,9 @@ FORGIT="$INSTALL_DIR/bin/git-forgit"
 # backwards compatibility:
 # export all user-defined FORGIT variables to make them available in git-forgit
 unexported_vars=0
+[[ -n "$BASH_VERSION" ]] && set -o posix
 set | awk -F '=' '{ print $1 }' | grep FORGIT_ | while read -r var; do
-    if ! export | grep -q "^$var="; then
+    if ! export | grep -q "\(^$var=\|^export $var=\)"; then
         if [[ $unexported_vars == 0 ]]; then
             forgit::warn "Config options have to be exported in future versions of forgit."
             forgit::warn "Please update your config accordingly:"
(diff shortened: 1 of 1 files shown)


forgit.plugin.zsh
```diff
@@ -21,8 +21,9 @@ FORGIT="$INSTALL_DIR/bin/git-forgit"
 # backwards compatibility:
 # export all user-defined FORGIT variables to make them available in git-forgit
 unexported_vars=0
+[[ -n "$BASH_VERSION" ]] && set -o posix
```
carlfriedrich, line 24: Maybe it would be good to add a describing comment here why we're calling `set -o posix`.

carlfriedrich (changes requested):
@gbtec-markusmeier Thanks a lot for your bug reports and this fix. Do I understand it correctly that this solution only fixes the situation for `bash`? Do we need a different behavior for `zsh`?

gbtec-markusmeier:
> @gbtec-markusmeier Thanks a lot for your bug reports and this fix. Do I understand it correctly that this solution only fixes the situation for `bash`? Do we need a different behavior for `zsh`?

Zsh works fine, we don't have that problem there: zsh reports only variables when calling `set`.

carlfriedrich (approved):
@gbtec-markusmeier Great, thanks a lot. Since it works for me as well in both shells, I will directly merge this. In fish this hasn't been reported as a problem so far.

PR #260, opened 2022-12-14
feat: allow overriding default $SHELL /bin/bash

Congee:
Some systems do not have /bin/bash. e.g., Alpine and NixOS. This change allows the user to provide one's own path of bash to fzf. It solves the error `fork/exec /bin/bash`.

Note that `fzf --preview 'xxxx'` essentially executes `$SHELL -c 'xxxx'`, so I tested on bash, zsh, and fish forgit::blame and it worked.

## Check list

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [x] bash
    - [x] zsh
    - [x] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -12,8 +12,10 @@
 # This gives users the choice to set aliases inside of their git config instead
 # of their shell config if they prefer.
 
-# Set shell for fzf preview commands
-SHELL=/bin/bash
+# shellcheck suggests "command -v xxx" instead. But it's not a replacement to
+# `which`. See https://github.com/koalaman/shellcheck/issues/1162
+# shellcheck disable=2230
+SHELL="$(which bash)" # Set shell for fzf preview commands
 
 # Get absolute forgit path
 FORGIT=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)/$(basename -- "${BASH_SOURCE[0]}")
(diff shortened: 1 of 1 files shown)


carlfriedrich:
@Congee Thanks for your submission. I wasn't aware that bash could be available in different places. We cannot use your proposed solution, though, because we have to enforce usage of bash as a shell for the preview commands. Otherwise users that have set their `SHELL` to something different (e.g. `fish`) will have non-working preview in fzf.

How about this:

```shell
SHELL=$(which bash)
```

Would that work for you?

Congee:
@carlfriedrich yo, I have updated this PR and the commit message as you suggested :grin:

bin/git-forgit
```diff
@@ -13,7 +13,8 @@
 # of their shell config if they prefer.
 
 # Set shell for fzf preview commands
-SHELL=/bin/bash
+# shellcheck disable=2230
```
cjappl, line 16: Is there a reason this error needed to be suppressed? Can you post a comment saying what the warning outputted before it was suppressed?

Congee: yeah I rewrote this commit and added a comment

carlfriedrich: @Congee Thanks. Your comment was interpreted by shellcheck, because it started with `# shellcheck`, so I slightly edited the commend to make it stop complaining. Thanks for your contribution!

Congee: ah thannnnnnnnks!

Issue #268, opened 2022-12-26
apperance truncated


jeusdi:
## Check list

- [x] I have read through the [README](https://github.com/wfxr/forgit/blob/master/README.md)
- [x] I have the latest version of forgit
- [x] I have searched through the existing issues

## Environment info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish

## Problem / Steps to reproduce

I've recorded an [asciinema](https://asciinema.org/a/mEsZudXAo4r1AnfdfWR32Tj37) in order to show the behavior.
1. diff left pane seems to be truncated.
2. blame theme is not applied.

Here my `~/.gitconfig`:

```
[core]
	editor = nvim
    pager = delta

[interactive]
    diffFilter = delta --color-only

[delta]
    navigate = true    # use n and N to move between diff sections
    light = false      # set to true if you're in a terminal w/ a light background color (e.g. the default macOS terminal)
    side-by-side = true

[merge]
    conflictstyle = diff3

[diff]
    colorMoved = default
```
As you can see, I'm using `delta`.

Any ideas?

cjappl:
I’ve noticed this as well when the diff is side by side instead of interleaved.

Can you try disabling side-by-side for delta from your git config and see if it fixes it?

I think one way to make this view better would be messing with the forgit override variables. You could try moving the preview window to below, and it may render better.

Cool asciinema! Never seen that before.

jeusdi:
I've tested disabling side-by-side.
The preview window content is expanded up to the right of terminal.
When I set up `side-by-side` again, "truncation" behavior arises again.

1. Why does this problem appear?
2. Do you think it can be fixed using variables?

burgessa23:
I'm seeing the same thing on macOS Ventura

0xRZ:
Related to https://github.com/wfxr/forgit/issues/121

cjappl:
Can you try the solution here https://github.com/wfxr/forgit/issues/121#issuecomment-968118124

Thanks for that suggestion @0xRZ !!

sandr01d:
Regarding the theme not being applied with blame:

There are two things at play here:

1. delta is not able to apply syntax highlighting when receiving it's input from a pipe, see dandavison/delta#1231
[...]

PR #277, opened 2023-01-13
Allow specifying custom blame date format

sandr01d:
## Check list

## Description

Allow the user to set a custom date format for `gbl` in `FORGIT_BLAME_DATE` for the preview. The current value (short) stays the default, so the preview does not get cluttered. This is especially useful with delta, which currently fails to apply it's blame theme when date is set to anything other than iso8601 or iso8601-local, see https://github.com/wfxr/forgit/issues/268#issuecomment-1377916562

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [X] bash
    - [X] zsh
    - [X] fish
- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,6 +274,7 @@ export FORGIT_LOG_FZF_OPTS='
 | `FORGIT_LOG_FORMAT`         | git log format                           | `%C(auto)%h%d %s %C(black)%C(bold)%cr%Creset` |
 | `FORGIT_PREVIEW_CONTEXT`    | lines of diff context in preview mode    | 3                                             |
 | `FORGIT_FULLSCREEN_CONTEXT` | lines of diff context in fullscreen mode | 10                                            |
+| `FORGIT_BLAME_DATE`         | git blame date format                    | short                                         |
 
 ### 📦 Optional dependencies
 
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -540,7 +540,7 @@ _forgit_blame() {
     flags=$(git rev-parse --flags "$@")
     preview="
         if $_forgit_is_file_tracked; then
-            git blame {} --date=short $flags | $_forgit_blame_pager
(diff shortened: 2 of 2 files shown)


carlfriedrich:
@sandr01d Thanks for your PR. For me this is basically fine.

I wonder, however, if we could implement a more general solution for passing custom arguments to git commands, which would also cover #281 for example.

@cjappl @wfxr What do you say?

cjappl:
> @sandr01d Thanks for your PR. For me this is basically fine.
> 
> I wonder, however, if we could implement a more general solution for passing custom arguments to git commands, which would also cover #281 for example.
> 
> @cjappl @wfxr What do you say?

I think that's ok, but I'd say let's merge this one now and think of that solution separately unless you already have a good one in mind. 

Just to not let it get in the way of a nice improvement! when we have that, we can easily swap this out. I'm good with this going in as-is

sandr01d:
> @sandr01d Thanks for your PR. For me this is basically fine.
> 
> I wonder, however, if we could implement a more general solution for passing custom arguments to git commands, which would also cover #281 for example.
> 
> @cjappl @wfxr What do you say?

I think a more general solution would be great, ideally one where the code can be shared across the different functions. I'm open to help implement such a solution, but it is currently hard for me to find the time. Also there should probably be a discussion about what exactly we'd want to have. So I agree with @cjappl's suggestion to merge this PR now and swap it out later.

carlfriedrich:
@sandr01d I have filed PR #292 with a more general solution which should cover your use case. Could you check if setting the following variable does what you want using this branch?

```shell
export FORGIT_BLAME_GIT_OPTS="--date=iso8601"
```

sandr01d:
Closing in favor of #292

PR #286, opened 2023-02-06
Fix broken _forgit_cherry_pick

sandr01d:
## Check list

## Description

I noticed that cherry-picking was broken for me, git always gave me the error message `fatal: bad revision ''`. The issue seems to be with the `cut -d' ' -f2` command which removes all the commit ids (at least for me).
I've refactored the code to use the read commands internal field separation instead. Would be especially courious to know if this works fine on your environment @cjappl, since you implemented the original functionality and I guess there might be some incompatibility between macOS and Linux at play here.

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [X] bash
    - [X] zsh
    - [X] fish
- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -289,10 +289,10 @@ _forgit_cherry_pick() {
     [[ -z "$fzf_selection" ]] && return $fzf_exitval
 
     commits=()
-    while IFS="" read -r line
+    while IFS=" " read -r line
     do
         commits+=("$line")
-    done < <(echo "$fzf_selection" | _forgit_reverse_lines | cut -d' ' -f2)
+    done < <(echo "$fzf_selection" | _forgit_reverse_lines)
 
     [ ${#commits[@]} -eq 0 ] && return 1
 


cjappl:
Works for me! Thanks for cleaning up my trash.

Will also request a review from @carlfriedrich in case he has time, if not I will merge it next week after I have some time to play with it

wfxr:
Hi @sandr01d, I tested this patch on linux and found the same error message displayed:

```
❯ gcp
fatal: bad revision ''
```

sandr01d:
> Hi @sandr01d, I tested this patch on linux and found the same error message displayed:
> 
> ```
> ❯ gcp
> fatal: bad revision ''
> ```

That's odd, I'm on Linux too and it works fine for me. Does this only happen for you with my patch or also without it? Also, does this happen regardless of the commit you pick or only on specific ones?

carlfriedrich:
@sandr01d @wfxr I have created another PR #294 which should also cover this error. Can you check?

sandr01d:
Closing in favor of #294

PR #290, opened 2023-03-13
Detect unexported variables in fish correctly (Fix #289)

### LINKED ISSUE #289: Detection of unexported variables is incorrect in fish if the variable has an empty string value
## Check list

## Environment info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:
- Shell
    - [ ] bash
    - [ ] zsh
    - [x] fish

In my `config.fish` I have:
```fish
set -x FORGIT_FZF_DEFAULT_OPTS "

--ansi
--height='80%'
--bind='alt-k:preview-up,alt-p:preview-up'
--bind='alt-j:preview-down,alt-n:preview-down'
--bind='ctrl-r:toggle-all'
--bind='ctrl-s:toggle-sort'
--bind='?:toggle-preview'
--bind='alt-w:toggle-preview-wrap'
--preview-window='right:60%'
+1

"
set -x FORGIT_GI_REPO_LOCAL ""
set -x FORGIT_GI_REPO_REMOTE "https://github.com/dvcs/gitignore"
set -x FORGIT_GI_TEMPLATES ""
```

Everytime I open a new fish shell I get
```
[Warn] Config options have to be exported in future versions of forgit.                                                                                       
[Warn] Please update your config accordingly:
```
The full message gets cut off by my prompt which is annoying but not forgit's fault.

Doing `set -x | grep FORGIT_` I can see all of those variables exported of course, so the question is why `set -x | grep -q "^$var "` in `forgit.plugin.fish` seems to fail

It seems to be due to the trailing whitespace in the argument to grep. `set -x | grep -q "^$var "` fails but `set -x | grep -q "^$var"`

When the variable's value is an empty string, the output of `set -x` does not have any trailing whitespace.

Unless there is some reason I am missing, simply removing the trailing whitespace from `grep -q "^$var "` should work

IndianBoy42:
## Check list

## Description

See #289 for the complete details

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [ ] zsh
    - [x] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/conf.d/forgit.plugin.fish
+++ b/conf.d/forgit.plugin.fish
@@ -7,14 +7,14 @@ if [ ! -e "$FORGIT" ]
 end
 
 function forgit::warn
-    printf "%b[Warn]%b %s\n" '\e[0;33m' '\e[0m' "$argv" >&2;
+    printf "%b[Warn]%b %s\n" '\e[0;33m' '\e[0m' "$argv" >&2
 end
 
 # backwards compatibility:
 # export all user-defined FORGIT variables to make them available in git-forgit
 set unexported_vars 0
 set | awk -F ' ' '{ print $1 }' | grep FORGIT_ | while read var
-    if ! set -x | grep -q "^$var "
+    if not set -x | grep -q "^$var\b"
         if test $unexported_vars = 0
             forgit::warn "Config options have to be exported in future versions of forgit."
             forgit::warn "Please update your config accordingly:"


cjappl:
Seems totally reasonable to me. Thanks for catching. Requested review from @carlfriedrich as well to make sure that tracks with his understanding.

I'll merge and close later in the week regardless. Thanks for the contribution

carlfriedrich:
Hi @IndianBoy42, thanks for the catch! I think the reason for the space in the grep expression was to mark the end of the variable name. Otherwise `FORGIT_SOMETHING` would also match `FORGIT_SOMETHING_ELSE`, which is not what we want here.

So we actually need the space. We could, however, extend the expression to *also* match the variable name followed by the end of the line (`$`), which I assume is what you have in your case? Could you check that?

carlfriedrich (changes requested):
Hi @IndianBoy42, thanks for the catch! I think the reason for the space in the grep expression was to mark the end of the variable name. Otherwise `FORGIT_SOMETHING` would also match `FORGIT_SOMETHING_ELSE`, which is not what we want here.

So we actually need the space. We could, however, extend the expression to *also* match the variable name followed by the end of the line (`$`), which I assume is what you have in your case? Could you check that?

IndianBoy42:
> So we actually need the space. We could, however, extend the expression to _also_ match the variable name followed by the end of the line (`$`), which I assume is what you have in your case? Could you check that?

Oh, that makes sense, it makes me think `\b` for word boundary potentially matches the intent most closely

carlfriedrich (approved):
Seems reasonable to me! Leaving final approval and merge for @cjappl.

cjappl:
Thanks for the submission @IndianBoy42

PR #292, opened 2023-03-17
Add git options for each forgit command

carlfriedrich:
## Check list

- [x] I have performed a self-review of my code
- [x] I have commented my code in hard-to-understand areas
- [x] I have made corresponding changes to the documentation

## Description

The git behavior within forgit can now be customized via dedicated variables for each forgit command, e.g. `FORGIT_ADD_GIT_OPTS` is passed to the `git add` call within `ga`.

This should make #277 obsolete.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

PR #296, opened 2023-03-18
Fix git options

carlfriedrich:
Adding options to a command in a quoted variable as introduced in #292 causes problems in several cases unfortunately (e.g. git rebase interpreting an empty string argument as a remote, which leads to an error "unknown upstream ''").

Change the implementation in a way that each git command is stored in a shell variable, including the options, which is then evaluated.

## Check list

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -153,10 +153,11 @@ _forgit_diff() {
 # git add selector
 _forgit_add() {
     _forgit_inside_work_tree || return 1
+    local git_add changed unmerged untracked files opts preview extract
+    git_add="git add $FORGIT_ADD_GIT_OPTS"
     # Add files if passed as arguments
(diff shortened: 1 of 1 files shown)


cjappl:
Works for me :) good job catching it.

carlfriedrich:
@cjappl Thanks for the quick check!

Gonna merge this directly as it is a fix for a bug introduced in #292. Hopefully everything works as it should. If you encounter any unintended behavior, please open an issue for it and assign me directly.

PR #304, opened 2023-05-05
feat: Use fzf-tmux if fzf options are set

### LINKED ISSUE #303: `forgit` does not open in a tmux popup window if fzf tmux options are set
## Check list

## Environment info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish

I have `FZF_TMUX=1` and `FZF_TMUX_OPTS='-p60%,80%'` set to be `export`ed by my `.zshrc`. When those are set, the built-in FZF helpers like Alt-C, etc. will default to `fzf-tmux` instead of `fzf`. And by setting `-p` with a `tmux` version >= 3.2, it uses the popup windows instead of a split:

image

I don't know for sure what the root cause is here, but I'm _**guessing**_ what's happening here is that `forgit` is hardcoded to use the `fzf` binary. It might be nice if `forgit` could detect `FZF_TMUX` or `FZF_TMUX_OPTS`, and if they're set it could behave like the built-in forgit extensions and default to `fzf-tmux`. This should be safe to do universally, as the fzf [README](https://github.com/junegunn/fzf#fzf-tmux-script) states that `fzf-tmux` is smart enough to be well behaved if invoked outside of a `tmux` session:

> It will still work even when you're not on tmux, silently ignoring -[pudlr] options, so you can invariably use fzf-tmux in your scripts.

dljsjr:
## Check list

## Overview

The meat of this change is to inspect whether or not we are in a `tmux` pane, and if so, we check the `FZF_TMUX` and `FZF_TMUX_OPTS` variables. If those are set, we use `fzf-tmux` instead of `fzf` and we forward the appropriate options.

This resolves #303 

## Changes
- Add a `do_fzf` function for abstracting fzf invocation based on shell state as outlined above

- Only hardcode `-0`/`--exit-0` for `fzf`, as it causes flicker w/ tmux popup

- Replace all invocations of `fzf` with `do_fzf`

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [x] bash
    - [x] zsh
    - [ ] fish
- OS
    - [x] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,17 +68,6 @@ Install `forgit` in just one click.
 
 [![Install with Fig](https://fig.io/badges/install-with-fig.svg)](https://fig.io/plugins/other/forgit)
 
-### `fzf-tmux` support
-
-If `FZF_TMUX` is not 0, or if `FZF_TMUX_OPTS` are set, then `forgit` will use `fzf-tmux`.
-
-More information can be found in [the fzf README](https://github.com/junegunn/fzf#fzf-tmux-script).
-
-If you use a command that can return no results, for example `gd`, and you are using `fzf-tmux`, you might
-see flicker as the pane or popup is opened and then closed. Setting the variable `FORGIT_FZF_SHOW_NO_MATCHES=1`
-will leave the open pane or popup window open. This might be useful if you don't like the flicker or if you
-need to make sure that `fzf` or `forgit` aren't closing for other reasons.
-
 ### 📝 Features
 
 - **Interactive `git add` selector** (`ga`)
@@ -317,6 +306,17 @@ export FORGIT_LOG_FZF_OPTS='
 | `FORGIT_LOG_FORMAT`         | git log format                           | `%C(auto)%h%d %s %C(black)%C(bold)%cr%Creset` |
 | `FORGIT_PREVIEW_CONTEXT`    | lines of diff context in preview mode    | 3                                             |
 | `FORGIT_FULLSCREEN_CONTEXT` | lines of diff context in fullscreen mode | 10                                            |
+### `fzf-tmux` support
+
+If `FZF_TMUX` is not 0, or if `FZF_TMUX_OPTS` are set, then `forgit` will use `fzf-tmux`.
+
+More information can be found in [the fzf README](https://github.com/junegunn/fzf#fzf-tmux-script).
+
+If you use a command that can return no results, for example `gd`, and you are using `fzf-tmux`, you might
+see flicker as the pane or popup is opened and then closed. Setting the variable `FORGIT_FZF_SHOW_NO_MATCHES=1`
+will leave the open pane or popup window open. This might be useful if you don't like the flicker or if you
+need to make sure that `fzf` or `forgit` aren't closing for other reasons.
+
 
 ### 📦 Optional dependencies
 
--- a/README.md
+++ b/README.md
@@ -68,6 +68,17 @@ Install `forgit` in just one click.
 
 [![Install with Fig](https://fig.io/badges/install-with-fig.svg)](https://fig.io/plugins/other/forgit)
 
+### `fzf-tmux` support
+
+If `FZF_TMUX` is not 0, or if `FZF_TMUX_OPTS` are set, then `forgit` will use `fzf-tmux`.
+
+More information can be found in [the fzf README](https://github.com/junegunn/fzf#fzf-tmux-script).
+
+When `fzf-tmux` is used, the "Exit if no matches" behavior is disabled because it can cause flicker
+with the tmux popup window. If you'd rather not see the empty pane or popup when using `fzf-tmux`,
+you can add `--exit-0` or `-0` to `FZF_TMUX_OPTS` to get the immediate close behavior with slight
+flicker.
+
 ### 📝 Features
 
 - **Interactive `git add` selector** (`ga`)
--- a/README.md
+++ b/README.md
@@ -74,10 +74,10 @@ If `FZF_TMUX` is not 0, or if `FZF_TMUX_OPTS` are set, then `forgit` will use `f
 
 More information can be found in [the fzf README](https://github.com/junegunn/fzf#fzf-tmux-script).
 
-When `fzf-tmux` is used, the "Exit if no matches" behavior is disabled because it can cause flicker
-with the tmux popup window. If you'd rather not see the empty pane or popup when using `fzf-tmux`,
-you can add `--exit-0` or `-0` to `FZF_TMUX_OPTS` to get the immediate close behavior with slight
-flicker.
+If you use a command that can return no results, for example `gd`, and you are using `fzf-tmux`, you might
+see flicker as the pane or popup is opened and then closed. Setting the variable `FORGIT_FZF_SHOW_NO_MATCHES=1`
+will leave the open pane or popup window open. This might be useful if you don't like the flicker or if you
+need to make sure that `fzf` or `forgit` aren't closing for other reasons.
 
 ### 📝 Features
 
--- a/README.md
+++ b/README.md
@@ -301,22 +301,12 @@ export FORGIT_LOG_FZF_OPTS='
 
 #### other options
 
-| Option                      | Description                              | Default                                       |
-|-----------------------------|------------------------------------------|-----------------------------------------------|
-| `FORGIT_LOG_FORMAT`         | git log format                           | `%C(auto)%h%d %s %C(black)%C(bold)%cr%Creset` |
-| `FORGIT_PREVIEW_CONTEXT`    | lines of diff context in preview mode    | 3                                             |
-| `FORGIT_FULLSCREEN_CONTEXT` | lines of diff context in fullscreen mode | 10                                            |
-### `fzf-tmux` support
-
-If `FZF_TMUX` is not 0, or if `FZF_TMUX_OPTS` are set, then `forgit` will use `fzf-tmux`.
-
-More information can be found in [the fzf README](https://github.com/junegunn/fzf#fzf-tmux-script).
-
-If you use a command that can return no results, for example `gd`, and you are using `fzf-tmux`, you might
-see flicker as the pane or popup is opened and then closed. Setting the variable `FORGIT_FZF_SHOW_NO_MATCHES=1`
-will leave the open pane or popup window open. This might be useful if you don't like the flicker or if you
-need to make sure that `fzf` or `forgit` aren't closing for other reasons.
-
+| Option                          | Description                                          | Default                                           |
+|---------------------------------|------------------------------------------------------|---------------------------------------------------|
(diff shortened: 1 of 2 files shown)


dljsjr:
I can address the lint failures when I’m back on a keyboard.

dljsjr:
Sent up a fixup commit to address the shellcheck lints, I'll squash it down before merge once approved :)

cjappl:
Don't stress about squashing, we use "squash and merge"

I'll hopefully get to test out and review soon! I've never heard about tmux-fzf but it sounds cool.

Thanks for the contribution, we will be back to you with comments/approval etc when we can 😀

dljsjr:
> I've never heard about tmux-fzf but it sounds cool.

For a heavy tmux user that uses a lot of panes/splits, the pop-up mode is invaluable:

CleanShot 2023-05-06 at 12 21 21

bin/git-forgit
```diff
@@ -136,14 +151,14 @@ _forgit_diff() {
     enter_cmd="cd '$repo' && $get_files | xargs -0 $git_diff -U$_forgit_fullscreen_context -- | $_forgit_diff_pager"
     opts="
         $FORGIT_FZF_DEFAULT_OPTS
-        +m -0 --bind=\"enter:execute($enter_cmd | $_forgit_enter_pager)\"
```
carlfriedrich, line 139: Is there a reason why you remove `-0` here but not in all other functions?

README.md
```diff
@@ -65,6 +65,17 @@ Install `forgit` in just one click.
...
+
+If `FZF_TMUX` is not 0, or if `FZF_TMUX_OPTS` are set, then `forgit` will use `fzf-tmux`.
+
+More information can be found in [the fzf README](https://github.com/junegunn/fzf#fzf-tmux-script).
+
+When `fzf-tmux` is used, the "Exit if no matches" behavior is disabled because it can cause flicker
+with the tmux popup window. If you'd rather not see the empty pane or popup when using `fzf-tmux`,
+you can add `--exit-0` or `-0` to `FZF_TMUX_OPTS` to get the immediate close behavior with slight
+flicker.
+
```
carlfriedrich, line 78: Great that you updated the documentation! I would prefer this block a bit further down, though, preferably after "shell aliases" or "git integration".

bin/git-forgit
```diff
@@ -19,10 +19,25 @@
...
+#
+# While the fzf docs say that `fzf-tmux` is safe to use in scripts even outside
+# of a tmux session, we still employ defensive variable tests.
+do_fzf() {
+  if [ -n "${TMUX_PANE-}" ] && { [ "${FZF_TMUX:-0}" != 0 ] || [ -n "${FZF_TMUX_OPTS-}" ]; }; then
+   # We don't hardcode --exit-0 here because it causes flicker when using
+   # tmux + popups instead of panes. Could be revisited.
+   fzf-tmux "${FZF_TMUX_OPTS:--d${FZF_TMUX_HEIGHT:-40%}}" "$@"
+  else
+   fzf -0 "$@"
```
carlfriedrich, line 36: I don't like hard-coding `-0` at this point for two reasons:

1. All other `fzf` arguments are defined somewhere else, this makes the final list of arguments harder to see in the code.
2. It actually changes behavior for some commands, since not every command had `-0` set in its argument list before.

The second point is questionable, though. I actually don't see any reason why we don't set `-0` on every command.
@wfxr @cjappl Do you have an idea? E.g. why do we set it on `gss` but not on `glo`?

carlfriedrich (changes requested):
Hey @dljsjr, thanks for your patch! I added some feedback to certain lines.

bin/git-forgit
dljsjr: I simply missed them 😅

bin/git-forgit
dljsjr: I'm gonna make a guess and say that the `-0` is only being set on commands where it's possible to have no results; `git log` is never gonna have no lines, for example.

That said it wouldn't *hurt* to have it everywhere. The "proper" fix is probably to introduce two separate `*DEFAULT_OPTS` variables for tmux and non-tmux but I was going for minimum-entropy with this changeset. I can do a more diligent refactor.

bin/git-forgit
```diff
@@ -38,6 +37,35 @@ $FZF_DEFAULT_OPTS
 $FORGIT_FZF_DEFAULT_OPTS
 "
 
+# If FORGIT_FZF_SHOW_NO_MATCHES is 1, then we don't pass `--exit-0` to fzf/fzf-tmux.
```
dljsjr, line 40: Since this is really `forgit`'s behavior, we make it a forgit variable instead of making it implicit to the variables we pass to `fzf`.

bin/git-forgit
```diff
@@ -38,6 +37,35 @@ $FZF_DEFAULT_OPTS
...
+    FORGIT_USE_FZF_TMUX=1
+fi
+
+
+# This function provides an abstraction for conditionally invoking
+# either `fzf` or `fzf-tmux` depending on the state of the current shell.
+#
+# While the fzf docs say that `fzf-tmux` is safe to use in scripts even outside
+# of a tmux session, we still employ defensive variable tests.
+do_fzf() {
```
dljsjr, line 60: Much cleaner :)

bin/git-forgit
dljsjr: This behavior is now user-pickable across the board: https://github.com/wfxr/forgit/pull/304/files#diff-e7404556b00c1c7af95effe409a0f05ded287236878c22735194ee0bc0b98d11R40

dljsjr:
@carlfriedrich I made some adjustments:

1. `-0` defaults to enabled across the board and the behavior is now configurable via a FORGIT-prefixed variable, mentioned in the README. Example:
	CleanShot 2023-05-07 at 16 37 49
2. Made the splitting and sharing of the regular options and the tmux options much more explicit
3. Adjusted the README as requested

bin/git-forgit
dljsjr: https://github.com/wfxr/forgit/pull/304#discussion_r1186918216

bin/git-forgit
carlfriedrich: That's a good solution, thank you. IMO the variable name might be more self-explaining, though, and maybe we could invert the meaning. How do you like `FORGIT_EXIT_ON_EMPTY_LIST`, which would then default to 1?

README.md
```diff
@@ -303,6 +303,17 @@ export FORGIT_LOG_FZF_OPTS='
...
+If `FZF_TMUX` is not 0, or if `FZF_TMUX_OPTS` are set, then `forgit` will use `fzf-tmux`.
+
+More information can be found in [the fzf README](https://github.com/junegunn/fzf#fzf-tmux-script).
+
+If you use a command that can return no results, for example `gd`, and you are using `fzf-tmux`, you might
+see flicker as the pane or popup is opened and then closed. Setting the variable `FORGIT_FZF_SHOW_NO_MATCHES=1`
+will leave the open pane or popup window open. This might be useful if you don't like the flicker or if you
+need to make sure that `fzf` or `forgit` aren't closing for other reasons.
+
```
carlfriedrich, line 316: Maybe we can put this information under "Optional dependencies", the block below. IMO that's where support for optional tools is listed.

And since the new `FORGIT_...` variable is not tmux specific, we could also list it above under "other options" and just reference it from the tmux doc entry.

bin/git-forgit
carlfriedrich: Top!

bin/git-forgit
carlfriedrich: Great!

carlfriedrich (changes requested):
Great, thanks for your effort. Just two more small change requests from my side.

bin/git-forgit
dljsjr: I think renaming the variable makes sense, and I did originally consider having it default to `1`, but when I thought about it, the convention amongst shell tools in general seems to be that having a variable completely unset should be implicitly falsy. I'm honestly still on the fence. I could go either way but I figured it might be worth making a comment and having at least a small back-and-forth about it.

carlfriedrich: > I think renaming the variable makes sense, and I did originally consider having it default to 1, but when I thought about it, the convention amongst shell tools in general seems to be that having a variable completely unset should be implicitly falsy. I'm honestly still on the fence. I could go either way but I figured it might be worth making a comment and having at least a small back-and-forth about it.

@dljsjr That definitely makes sense and seems like a sane default behavior. So how about `FORGIT_SHOW_FZF_ON_EMPTY_LIST` then?

carlfriedrich (approved):

carlfriedrich:
@dljsjr Great, looks good code-wise for me now. Thanks a lot for your contribution!

I Did not do any functional tests, though, since I do not use tmux.
@cjappl Do you want to do any more testing or review? Otherwise I will merge this shortly.

cjappl:
Sure. Let me test with and without this plugin and make sure it’s all good! I’ll do that today, and merge if everything looks good.

dljsjr:
@cjappl Hmm, interesting. I didn't modify any of that code, and I don't use `fish`.

I could probably get it set up and do a bisect. I *did* rebase this change on top of the latest `origin/master` after my last round of revisions so it also could have pulled in a regression from upstream.

dljsjr:
@cjappl do you have anything in your `.tmux.conf` that forces a specific shell type, such as the macos reattach to user namespace stuff for pasteboard cooperation?

I think what's going on here is that the spawned `tmux` pane is executing things in `fish` instead of `bash` but I'm not 100% sure why yet. It's definitely not a regression, just something that's unique to the tmux configuration, and I'm not 100% sure why just right now.

dljsjr:
@cjappl I've made the proposed change *only* for `ga` on this branch for right now; let me know if that works.

cjappl:
Unfortunately no dice, similar issue

Screen Shot 2023-05-13 at 12 47 45 PM


I think the best option I'd recommend is you install and test it on `fish`. Otherwise, I can do the work on `fish` at some point in the future but it may take a bit for me to get to it!

cjappl (changes requested):
Blocking merge for now until fish issues resolved

dljsjr:
@cjappl Guess this is my excuse to give `fish` a try for the first time in like, 8 years :)

cjappl:
Haha! It’s a blessing and a curse :)

Appreciate you being willing to chase it down. Shout if you need a test subject, happy to run anything you come up with.

carlfriedrich:
@dljsjr Shall we reopen this? What are the chances that you'll get back to this?

PR #306, opened 2023-05-08
Use correct IFS store/restore mechanism

### LINKED ISSUE #301: grc giving back error "git branch: command not found"
When trying to run grc this morning, I got the error:

/Users/chrisapple/code/personal//forgit/conf.d/bin/git-forgit: line 568: git branch : command not found

I think there are two things going wrong here:
1. The command should be `git revert` not `git branch`
2. Something is fishy about how the command is being evaluated on line 592, not sure how to resolve it.

Repro'd on OSX

carlfriedrich:
The previous implementation had the problem that if IFS was not set before, it was set to an empty string after restoring, which is not the same as being unset. This broke grc / git forgit revert_commit, since "git revert" was being interpreted as the command name instead of command and argument.

The correct way is to check whether IFS is unset and, if so, unset it again afterwards.

See for reference:
https://unix.stackexchange.com/a/264947/317320

Fix #301.

## Check list

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -314,10 +314,11 @@ _forgit_cherry_pick() {
     [[ $fzf_exitval != 0 ]] && return $fzf_exitval
     [[ -z "$fzf_selection" ]] && return $fzf_exitval
 
-    old_IFS=$IFS IFS=$'\n' 
+    ${IFS+"false"} && unset old_IFS || old_IFS="$IFS"
+    IFS=$'\n'
     # shellcheck disable=2207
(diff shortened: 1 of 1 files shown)


carlfriedrich:
@cjappl Can you check if this PR fixes `grc` for you?

cjappl:
Works like a charm! Great sleuthing @carlfriedrich . 

I'm going to merge

PR #313, opened 2023-06-14
feat: add a keybind to open current file in editor

wfxr:
## Check list

## Description

Add a keybind `alt-e` to allow open current file in default editor. This will facilitate doing diff and modify loop. 

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -138,7 +138,7 @@ _forgit_diff() {
         $FORGIT_FZF_DEFAULT_OPTS
         +m -0 --bind=\"enter:execute($enter_cmd | $_forgit_enter_pager)\"
         --preview=\"$preview_cmd\"
-        --bind=\"alt-e:execute-silent($EDITOR \$\($get_files\) >/dev/tty </dev/tty)\"
+        --bind=\"alt-e:execute-silent($EDITOR \$\($get_files\) >/dev/tty </dev/tty)+refresh-preview\"
         $FORGIT_DIFF_FZF_OPTS
         --prompt=\"$commits > \"
     "
@@ -176,7 +176,7 @@ _forgit_add() {
         $FORGIT_FZF_DEFAULT_OPTS
         -0 -m --nth 2..,..
         --preview=\"$preview\"
-        --bind=\"alt-e:execute-silent($EDITOR \$\(echo {} | $extract\) >/dev/tty </dev/tty)\"
+        --bind=\"alt-e:execute-silent($EDITOR \$\(echo {} | $extract\) >/dev/tty </dev/tty)+refresh-preview\"
         $FORGIT_ADD_FZF_OPTS
     "
     files=$(git -c color.status=always -c status.relativePaths=true status -su |
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -138,6 +138,7 @@ _forgit_diff() {
         $FORGIT_FZF_DEFAULT_OPTS
         +m -0 --bind=\"enter:execute($enter_cmd | $_forgit_enter_pager)\"
         --preview=\"$preview_cmd\"
+        --bind=\"alt-e:execute-silent($EDITOR \$\($get_files\) >/dev/tty </dev/tty)\"
         $FORGIT_DIFF_FZF_OPTS
         --prompt=\"$commits > \"
     "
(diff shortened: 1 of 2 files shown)


bin/git-forgit
```diff
@@ -138,6 +138,7 @@ _forgit_diff() {
         $FORGIT_FZF_DEFAULT_OPTS
         +m -0 --bind=\"enter:execute($enter_cmd | $_forgit_enter_pager)\"
         --preview=\"$preview_cmd\"
+        --bind=\"alt-e:execute-silent($EDITOR \$\($get_files\) >/dev/tty </dev/tty)\"
```
sandr01d, line 141: Changes made in the editor do currently not become visible in the preview immediately after exiting (they do become visible when changing selection though). We could make them visible immediately like this:

```suggestion
        --bind=\"alt-e:execute-silent($EDITOR \$\($get_files\) >/dev/tty </dev/tty)+refresh-preview\"
```

bin/git-forgit
```diff
@@ -178,6 +176,7 @@ _forgit_add() {
         $FORGIT_FZF_DEFAULT_OPTS
         -0 -m --nth 2..,..
         --preview=\"$preview\"
+        --bind=\"alt-e:execute-silent($EDITOR \$\(echo {} | $extract\) >/dev/tty </dev/tty)\"
```
sandr01d, line 179: Same here

```suggestion
        --bind=\"alt-e:execute-silent($EDITOR \$\(echo {} | $extract\) >/dev/tty </dev/tty)+refresh-preview\"
```

sandr01d (changes requested):
Great idea, this could massively improve my workflow! I've requested a minor improvement though.

wfxr:
> Great idea, this could massively improve my workflow! I've requested a minor improvement though.

Done.

carlfriedrich:
@wfxr Great addition, works for me!

@sandr01d Great idea as well! However, refreshing does not work correctly for me. For some reason, the fzf screen is not fully redrawn after quitting the editor, so that I get a mixture of shell promts and fzf. Any idea why this might be the case?

And some further idea: consider you edit the file and revert all the changes, then the file should not appear in the fzf dialog at all afterwards. Maybe this can be achieved with the `reload()` action of fzf? Haven't tried it, because I do not have time right now. But maybe someone of you might check that?

wfxr:
> @sandr01d Great idea as well! However, refreshing does not work correctly for me. For some reason, the fzf screen is not fully redrawn after quitting the editor, so that I get a mixture of shell promts and fzf. Any idea why this might be the case?

@carlfriedrich Press `CTRL-L` to flush the buffer usually works when I sometimes have a similar problem when resizing the window.

sandr01d:
> @sandr01d Great idea as well! However, refreshing does not work correctly for me. For some reason, the fzf screen is not fully redrawn after quitting the editor, so that I get a mixture of shell promts and fzf. Any idea why this might be the case?

@carlfriedrich I can not reproduce this. Does this happen only with the `refresh-preview` or also without it?

> I have changed my mind, and tend to keep the file list as fzf started, even if changes on a specified file are reverted. This gives us the opportunity to select and edit it again easily. So IMO @sandr01d's `refresh-preview` improvement is good enough. What do you think?

I definitively see the use case here. Might be a bit confusing for some users though. Maybe we could be transparent about it and indicate that the file has no changes anymore in the preview, similar as we do here:

https://github.com/wfxr/forgit/blob/ec1bd214683d1244f2512222dee11ac467e4ed17/bin/git-forgit#LL588C1-L594C6

carlfriedrich:
> @carlfriedrich I can not reproduce this. Does this happen only with the `refresh-preview` or also without it?

@sandr01d It actually happens with and without it, so adding it is good in either way.

> @carlfriedrich Press `CTRL-L` to flush the buffer usually works when I sometimes have a similar problem when resizing the window.

@wfxr That works indeed, thanks for the hint!

> I have changed my mind, and tend to keep the file list as fzf started, even if changes on a specified file are reverted. This gives us the opportunity to select and edit it again easily. So IMO @sandr01d's `refresh-preview` improvement is good enough. What do you think?

I see your point and agree with you.

bin/git-forgit
```diff
@@ -163,10 +164,7 @@ _forgit_add() {
     untracked=$(git config --get-color color.status.untracked red)
     # NOTE: paths listed by 'git status -su' mixed with quoted and unquoted style
     # remove indicators | remove original path for rename case | remove surrounding quotes
-    extract="
-        sed 's/^.*]  //' |
-        sed 's/.* -> //' |
-        sed -e 's/^\\\"//' -e 's/\\\"\$//'"
+    extract="sed 's/^.*]  //' | sed 's/.* -> //' | sed -e 's/^\\\"//' -e 's/\\\"\$//'"
```
carlfriedrich, line 167: Just one note: this seems like an unrelated refactoring change. Should we remove this in order to keep the commit clean?

wfxr: Good catch! This change was made when I debug the keybind. Let's revert it.

wfxr: Done.

carlfriedrich (approved):
@wfxr Great, approved!

sandr01d (approved):

PR #316, opened 2023-07-01
Fix infinite loop when passing invalid arguments to gsp

sandr01d:
## Check list

## Description

Passing invalid arguments to `_forgit_stash_push` results in an infinite loop printing gits error message. You can test this by running `gsp -b`.
With this PR we only print the error message once and return gits exit code.

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [X] bash
    - [X] zsh
    - [X] fish
- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -249,7 +249,7 @@ _forgit_stash_push() {
             # ignore -u as it's used implicitly
             -u|--include-untracked) shift ;;
             # pass to git directly when encountering anything else
-            *) $git_stash_push "${args[@]}" && return $?
+            *) $git_stash_push "${args[@]}"; return $?
         esac
     done
     local opts preview files


carlfriedrich (commented):
@sandr01d Thanks for the catch! This is actually something I put on my to-do list a while ago as well.
We do this in all functions: return only if the command with the appended arguments was successful. This does not lead to infinite loops in other functions, but it hides error messages, because the forgit dialog pops up if something was wrong with the arguments.
IMO we should always return immediately if arguments were given, independently of the error code. If something goes wrong, I want to see the error message. So we should do a PR to change the behavior like this for all functions at once.
What do you think?

sandr01d:
I agree with you, I prefer to have the error message. Also there are a few functions were we already do return in case of an error, so refactoring this would make things more consistent.
There is an exception to this: In `gbl` we actually make use of this behavior. We do not return in case `git blame` returned an error, but use the parameters later on. This allows things like `gbl --color-by-age`. I do something similar in #317. In those cases we could just check whether all passed arguments are flags and only invoke the git command without fzf if there were non-flag parameters passed.
I can update #317 accordingly. Would you prefer to have a separate PR for the refactoring or have this included in this one?

carlfriedrich:
@sandr01d That sounds totally reasonable and I like the idea of checking for flags.
IMO you can merge #317 directly and then do a follow up PR to do the refactoring.
Thanks for your work on this!

sandr01d:
Closing in favor of #318

PR #318, opened 2023-07-11
Do not hide error messages when trying to return early

sandr01d:
## Check list

## Description

Most of forgits functions allow to invoke git directly when arguments were passed. In most of these functions we only return early when the git command that is executed returns successfully. This hides error messages from the user. As discussed with @carlfriedrich in #316, I've modified this behavior to also return early when the git command was not successful, so we're transparent about gits error messages and return values. Because `gbl` and `gclean` allow passing arguments to the git command that is invoked when we don't return early, I've modified their behavior to only invoke git directly (and return early) in the case that non flag arguments were passed.
I've also modified `gclean` to use the `-q` flag instead of piping it's output to /dev/null.
This PR supersedes #316.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [X] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [X] zsh
    - [ ] fish
- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -53,6 +53,18 @@ _forgit_previous_commit() {
     fi
 }
 
+_forgit_contains_non_flags() {
+    while (("$#")); do
+        case "$1" in
+        -*) shift ;;
+        *)
+            return 0
+            ;;
+        esac
+    done
+    return 1
+}
+
 # optional render emoji characters (https://github.com/wfxr/emoji-cli)
 hash emojify &>/dev/null && _forgit_emojify='|emojify'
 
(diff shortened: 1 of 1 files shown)


sandr01d:
I've used this for a week now and did not find any issues. I'd like to merge this PR soon. Any thoughts on this?

cjappl (approved):

carlfriedrich:
Sorry for the late reply! Thanks a lot for your work on this @sandr01d . I haven't had time to test it out, yet, but code-wise it looks good to me.
I would suggest squashing the two commits when merging this, though, and use the really good explanation in the PR's description as a commit message.

carlfriedrich (approved):

sandr01d:
> I would suggest squashing the two commits when merging this, though, and use the really good explanation in the PR's description as a commit message.

Will do.

I'll leave this hanging for a bit in case somebody wants to test this. I'll merge this next weekend when there are no objections.

PR #320, opened 2023-08-02
Meta: Fix AUR action not triggering

sandr01d:
## Check list

## Description

Turns out that GitHub actions do not trigger other actions by default. See this [Stack Overflow question](https://stackoverflow.com/questions/69063452/github-actions-on-release-created-workflow-trigger-not-working) for more details. Using `workflow_run` instead of `release` as a trigger should solve this.

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [ ] zsh
    - [ ] fish
- OS
    - [ ] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/.github/workflows/aur.yaml
+++ b/.github/workflows/aur.yaml
@@ -1,9 +1,11 @@
 name: AUR
 
 on:
-  release:
+  workflow_run:
+    workflows:
+      - "Release"
     types:
-      - "released"
+      - completed
 
 jobs:
   aur:


sandr01d:
Pinging @carlfriedrich for a review because you appear to have commented in the stack overflow discussion I linked above, so I'm guessing you ran into this at some point as well. Also, are you aware of a way to properly test this without having to wait for the next release?

cjappl (approved):

carlfriedrich:
@sandr01d Yes, I stumbled upon the same problem and used the method using a deploy key in the end, just like I described in the linked Stack Overflow thread.

Hence I am surprised that your previous implementation did not work, because the `release` action is correctly triggered after `tag`. I assume tagging and releasing are different kinds of triggers that are handled differently by GitLab, and the deploy key method obviously does not work for the release trigger.

However, your new solution seems legit, and we should give it a try.

I am not aware of a way to test this without a release tag. You can, however, always push a tag manually, so we don't have to wait for the automatic monthly releases. That's what the last number in our version scheme is for, so you could push a tag `2023.08.1` for testing.

carlfriedrich (approved):

PR #322, opened 2023-08-10
checkout_branch: support "-" argument

carlfriedrich:
git natively supports `-` as an argument to `git switch` and `git checkout`. It is shorthand for `@{-1}`, which is a way to refer to the last branch you were on.

forgit used to interpret `-` as a branch name, detect that it does not exist yet and create a new one with this name, which does not work.

Add a check whether `-` is passed on the command line and do not create a new branch in this case

## Check list

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -454,7 +454,7 @@ _forgit_checkout_branch() {
     _forgit_inside_work_tree || return 1
     # if called with arguments, check if branch exists, else create a new one
     if [[ $# -ne 0 ]]; then
-        if git show-branch "$@" &>/dev/null; then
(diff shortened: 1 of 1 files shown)


cjappl (approved):
Seems reasonable to me

sandr01d (approved):
Works fine

PR #326, opened 2023-10-15
Reduce deferred execution

sandr01d:
## Check list

## Description

We've removed deferred code and replaced it with functions instead. This includes all usages of `eval`. There are four different types of functions we've added:

1. Preview functions (e.g. `_forgit_add_preview`) that are responsible for the `fzf` preview
1. Edit functions (e.g. `_forgit_edit_add_file`) that are responsible for editing files with your `$EDITOR` from within forgit functions
1. Git functions (e.g. `_forgit_git_add`) that make simple git commands reusable
1. Utility functions (e.g. `_forgit_yank_sha` or `_forgit_quote_files`) that were created either out of necessity or implement functions that were previously stored as deferred code in variables.

We've exposed some of these functions as forgit commands. This includes all preview and edit functions and some of the utility functions, so they can be invoked from `fzf`'s subshell. Additionaly some of the git functions were exposed to make it possible to use them with `xargs`.

In some places were we used deferred code before, we now make use of arrays to allow building commands while having control over globbing and word splitting. We do this for all the git opts environment variables, such as `FORGIT_ADD_GIT_OPTS`.

In places were we could not avoid deferred execution (e.g. when having to pass arguments to preview functions as we do in `_forgit_log`) we ensured variables are quoted properly. Variables not quoted properly were one of the main source of bugs previously.  

### General
We now make use of the `-a` flag with `read` which does exist in bash, but not in zsh. This is fine, since we always use bash for executing /bin/git-forgit since #241.

### `gbl`
It is now possible to pass more than one argument to `gbl` (e.g. `gbl --color-by-age --color-lines`). This was a bug previously and the fix is a side effect of us now using an array for the flags instead of relying on `eval`.  

### `grc`
I've removed passing the files variable to the preview, as well as passing parameters to the git log command with `$*`. These were creating issues in the new implantation and they were actually always empty, because `_forgit_revert` exits early when any arguments are passed. I copied them by mistake when I implemented `_forgit_revert`, they did never actually serve any purpose.

### `glo`, `grb` & `gfu`
File arguments are now parsed using a different approach using `_forgit_quote_files`. The only difference in behavior to the `sed` command we had previously is that the string that is printed out has each file name wrapped in single quotes, so they can be used safely without having to worry about globbing or word splitting. There is no need to quote this string when passing it to a preview function, because the individual file names inside it are already properly quoted. I changed the files variable names to `quoted_files` to indicate this.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [X] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [X] bash
    - [X] zsh
    - [X] fish
- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -80,6 +80,20 @@ _forgit_extract_sha() {
     grep -Eo '[a-f0-9]+' | head -1 | tr -d '[:space:]'
 }
 
+# extract the first git sha and copy it to the clipboard
+_forgit_yank_sha() {
+    echo "$1" | _forgit_extract_sha | ${FORGIT_COPY_CMD:-pbcopy}
+}
+
+# extract the first stash name in the input
+_forgit_extract_stash_name() {
+    cut -d: -f1 | tr -d '[:space:]'
+}
+
+# extract the first stash name and copy it to the clipboard
+_forgit_yank_stash_name() {
+    echo "$1" | _forgit_extract_stash_name | ${FORGIT_COPY_CMD:-pbcopy}
+}
 
 # parse a space separated string into an array
 # arrays parsed with this function are global
@@ -131,7 +145,7 @@ _forgit_log() {
         $FORGIT_FZF_DEFAULT_OPTS
         +s +m --tiebreak=index
         --bind=\"enter:execute($FORGIT log_enter {} $files)\"
-        --bind=\"ctrl-y:execute-silent(echo {} | $_forgit_extract_sha | ${FORGIT_COPY_CMD:-pbcopy})\"
+        --bind=\"ctrl-y:execute-silent($FORGIT yank_sha {})\"
         --preview=\"$FORGIT log_preview {} $files\"
         $FORGIT_LOG_FZF_OPTS
     "
@@ -362,7 +376,7 @@ _forgit_stash_show() {
     opts="
         $FORGIT_FZF_DEFAULT_OPTS
         +s +m -0 --tiebreak=index --bind=\"enter:execute($FORGIT stash_show_preview {} | $_forgit_enter_pager)\"
-        --bind=\"ctrl-y:execute-silent(echo {} | cut -d: -f1 | tr -d '[:space:]' | ${FORGIT_COPY_CMD:-pbcopy})\"
+        --bind=\"ctrl-y:execute-silent($FORGIT yank_stash_name {})\"
         --preview=\"$FORGIT stash_show_preview {}\"
         $FORGIT_STASH_FZF_OPTS
     "
@@ -534,7 +548,7 @@ _forgit_rebase() {
     opts="
         $FORGIT_FZF_DEFAULT_OPTS
         +s +m --tiebreak=index
-        --bind=\"ctrl-y:execute-silent(echo {} | $_forgit_extract_sha | ${FORGIT_COPY_CMD:-pbcopy})\"
+        --bind=\"ctrl-y:execute-silent($FORGIT yank_sha {})\"
         --preview=\"$FORGIT file_preview {} $files\"
         $FORGIT_REBASE_FZF_OPTS
     "
@@ -568,7 +582,7 @@ _forgit_fixup() {
     opts="
         $FORGIT_FZF_DEFAULT_OPTS
         +s +m --tiebreak=index
-        --bind=\"ctrl-y:execute-silent(echo {} | $_forgit_extract_sha | ${FORGIT_COPY_CMD:-pbcopy})\"
+        --bind=\"ctrl-y:execute-silent($FORGIT yank_sha {})\"
         --preview=\"$FORGIT file_preview {} $files\"
         $FORGIT_FIXUP_FZF_OPTS
     "
@@ -697,7 +711,7 @@ _forgit_checkout_commit() {
     opts="
         $FORGIT_FZF_DEFAULT_OPTS
         +s +m --tiebreak=index
-        --bind=\"ctrl-y:execute-silent(echo {} | $_forgit_extract_sha | ${FORGIT_COPY_CMD:-pbcopy})\"
+        --bind=\"ctrl-y:execute-silent($FORGIT yank_sha {})\"
         --preview=\"$FORGIT checkout_commit_preview {}\"
         $FORGIT_CHECKOUT_COMMIT_FZF_OPTS
     "
@@ -919,6 +933,8 @@ private_commands=(
     "reset_head_preview"
     "stash_push_preview"
     "stash_show_preview"
+    "yank_sha"
+    "yank_stash_name"
     "log_preview"
     "log_enter"
     "exec_diff"
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -66,7 +66,13 @@ _forgit_contains_non_flags() {
 }
 
 # optional render emoji characters (https://github.com/wfxr/emoji-cli)
-hash emojify &>/dev/null && _forgit_emojify='|emojify'
+_forgit_emojify() {
+    if hash emojify &>/dev/null; then
+        emojify
+    else
+        cat
+    fi
+}
 
 # extract the first git sha occurring in the input and strip trailing newline
 _forgit_extract_sha="grep -Eo '[a-f0-9]+' | head -1 | tr -d '[:space:]'"
@@ -127,7 +133,8 @@ _forgit_log() {
     log_format=${FORGIT_GLO_FORMAT:-$_forgit_log_format}
     _forgit_log_git_opts=()
     _forgit_parse_array _forgit_log_git_opts "$FORGIT_LOG_GIT_OPTS"
-    eval "git log $graph --color=always --format='$log_format' ${_forgit_log_git_opts[*]} $* $_forgit_emojify" |
+    eval "git log $graph --color=always --format='$log_format' ${_forgit_log_git_opts[*]} $*" |
+        _forgit_emojify |
         FZF_DEFAULT_OPTS="$opts" fzf
     fzf_exit_code=$?
     # exit successfully on 130 (ctrl-c/esc)
@@ -511,7 +518,7 @@ _forgit_rebase() {
     git_rebase="git rebase -i ${_forgit_rebase_git_opts[*]}"
     graph=--graph
     [[ $FORGIT_LOG_GRAPH_ENABLE == false ]] && graph=
-    cmd="git log $graph --color=always --format='$_forgit_log_format' $* $_forgit_emojify"
+    cmd="git log $graph --color=always --format='$_forgit_log_format' $*"
     files=$(sed -nE 's/.* -- (.*)/\1/p' <<< "$*") # extract files parameters for `git show` command
     opts="
         $FORGIT_FZF_DEFAULT_OPTS
@@ -520,7 +527,7 @@ _forgit_rebase() {
         --preview=\"$FORGIT file_preview {} $files\"
         $FORGIT_REBASE_FZF_OPTS
     "
-    target_commit=$(eval "$cmd" | FZF_DEFAULT_OPTS="$opts" fzf | eval "$_forgit_extract_sha")
+    target_commit=$(eval "$cmd" | _forgit_emojify | FZF_DEFAULT_OPTS="$opts" fzf | eval "$_forgit_extract_sha")
     if [[ -n "$target_commit" ]]; then
         prev_commit=$(_forgit_previous_commit "$target_commit")
 
@@ -544,7 +551,7 @@ _forgit_fixup() {
     git_fixup="git commit --fixup ${_forgit_fixup_git_opts[*]}"
     graph=--graph
     [[ $FORGIT_LOG_GRAPH_ENABLE == false ]] && graph=
-    cmd="git log $graph --color=always --format='$_forgit_log_format' $* $_forgit_emojify"
+    cmd="git log $graph --color=always --format='$_forgit_log_format' $*"
     files=$(sed -nE 's/.* -- (.*)/\1/p' <<< "$*")
     opts="
         $FORGIT_FZF_DEFAULT_OPTS
@@ -553,7 +560,7 @@ _forgit_fixup() {
         --preview=\"$FORGIT file_preview {} $files\"
         $FORGIT_FIXUP_FZF_OPTS
     "
-    target_commit=$(eval "$cmd" | FZF_DEFAULT_OPTS="$opts" fzf | eval "$_forgit_extract_sha")
+    target_commit=$(eval "$cmd" | _forgit_emojify | FZF_DEFAULT_OPTS="$opts" fzf | eval "$_forgit_extract_sha")
     if [[ -n "$target_commit" ]] && $git_fixup "$target_commit"; then
         prev_commit=$(_forgit_previous_commit "$target_commit")
         # rebase will fail if there are unstaged changes so --autostash is needed to temporarily stash them
@@ -681,7 +688,8 @@ _forgit_checkout_commit() {
     "
     graph=--graph
     [[ $FORGIT_LOG_GRAPH_ENABLE == false ]] && graph=
-    commit="$(eval "git log $graph --color=always --format='$_forgit_log_format' $_forgit_emojify" |
+    commit="$(eval "git log $graph --color=always --format='$_forgit_log_format'" |
+        _forgit_emojify |
         FZF_DEFAULT_OPTS="$opts" fzf | eval "$_forgit_extract_sha")"
     _forgit_git_checkout_commit "$commit"
 }
@@ -734,7 +742,7 @@ _forgit_revert_commit() {
     local cmd opts files commits IFS
     [[ $# -ne 0 ]] && { _forgit_git_revert "$@"; return $?; }
 
-    cmd="git log --graph --color=always --format='$_forgit_log_format' $* $_forgit_emojify"
+    cmd="git log --graph --color=always --format='$_forgit_log_format' $*"
     opts="
         $FORGIT_FZF_DEFAULT_OPTS
         +s --tiebreak=index
@@ -752,7 +760,8 @@ _forgit_revert_commit() {
     ${IFS+"false"} && unset old_IFS || old_IFS="$IFS"
     IFS=$'\n'
     # shellcheck disable=2207
-    commits=($(eval "$cmd" | 
+    commits=($(eval "$cmd" |
+        _forgit_emojify |
         nl |
         FZF_DEFAULT_OPTS="$opts" fzf --preview="$FORGIT revert_preview {}" -m |
         sort -n -k 1 | 
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -71,6 +71,16 @@ hash emojify &>/dev/null && _forgit_emojify='|emojify'
 # extract the first git sha occurring in the input and strip trailing newline
 _forgit_extract_sha="grep -Eo '[a-f0-9]+' | head -1 | tr -d '[:space:]'"
 
+# parse a space separated string into an array
+# arrays parsed with this function are global
+_forgit_parse_array() {
+    ${IFS+"false"} && unset old_IFS || old_IFS="$IFS"
+    # read the value of the second argument
+    # into an array that has the name of the first argument
+    IFS=" " read -r -a "$1" <<< "$2"
+    ${old_IFS+"false"} && unset IFS || IFS="$old_IFS"
+}
+
 _forgit_pager=${FORGIT_PAGER:-$(git config core.pager || echo 'cat')}
 _forgit_show_pager=${FORGIT_SHOW_PAGER:-$(git config pager.show || echo "$_forgit_pager")}
 _forgit_diff_pager=${FORGIT_DIFF_PAGER:-$(git config pager.diff || echo "$_forgit_pager")}
@@ -102,7 +112,9 @@ _forgit_log() {
     graph=--graph
     [[ $FORGIT_LOG_GRAPH_ENABLE == false ]] && graph=
     log_format=${FORGIT_GLO_FORMAT:-$_forgit_log_format}
-    eval "git log $graph --color=always --format='$log_format' $FORGIT_LOG_GIT_OPTS $* $_forgit_emojify" |
+    _forgit_log_git_opts=()
+    _forgit_parse_array _forgit_log_git_opts "$FORGIT_LOG_GIT_OPTS"
+    eval "git log $graph --color=always --format='$log_format' ${_forgit_log_git_opts[*]} $* $_forgit_emojify" |
         FZF_DEFAULT_OPTS="$opts" fzf
     fzf_exit_code=$?
     # exit successfully on 130 (ctrl-c/esc)
@@ -146,7 +158,9 @@ _forgit_diff() {
     # prevent fzf from interpreting this substring by escaping the opening bracket.
     # The string is evaluated a few subsequent times, so we need multiple escapes.
     escaped_commits=${commits//\{/\\\\\{}
-    git_diff="git diff --color=always $FORGIT_DIFF_GIT_OPTS $escaped_commits"
+    _forgit_diff_git_opts=()
+    _forgit_parse_array _forgit_diff_git_opts "$FORGIT_DIFF_GIT_OPTS"
+    git_diff="git diff --color=always ${_forgit_diff_git_opts[*]} $escaped_commits"
     preview_cmd="cd '$repo' && $get_files | xargs -0 $git_diff -U$_forgit_preview_context -- | $_forgit_diff_pager"
     enter_cmd="cd '$repo' && $get_files | xargs -0 $git_diff -U$_forgit_fullscreen_context -- | $_forgit_diff_pager"
     opts="
@@ -157,7 +171,7 @@ _forgit_diff() {
         $FORGIT_DIFF_FZF_OPTS
         --prompt=\"$commits > \"
     "
-    eval "git diff --name-status $FORGIT_DIFF_GIT_OPTS $commits -- ${files[*]} | sed -E 's/^([[:alnum:]]+)[[:space:]]+(.*)$/[\1]	\2/'" |
+    eval "git diff --name-status ${_forgit_diff_git_opts[*]} $commits -- ${files[*]} | sed -E 's/^([[:alnum:]]+)[[:space:]]+(.*)$/[\1]	\2/'" |
         sed 's/	/  ->  /2' | expand -t 8 |
         FZF_DEFAULT_OPTS="$opts" fzf
     fzf_exit_code=$?
@@ -170,7 +184,9 @@ _forgit_diff() {
 _forgit_add() {
     _forgit_inside_work_tree || return 1
     local git_add changed unmerged untracked files opts preview extract
-    git_add="git add $FORGIT_ADD_GIT_OPTS"
+    _forgit_add_git_opts=()
+    _forgit_parse_array _forgit_add_git_opts "$FORGIT_ADD_GIT_OPTS"
+    git_add="git add ${_forgit_add_git_opts[*]}"
     # Add files if passed as arguments
     [[ $# -ne 0 ]] && { $git_add "$@" && git status -su; return $?; }
 
@@ -210,7 +226,9 @@ _forgit_add() {
 _forgit_reset_head() {
     _forgit_inside_work_tree || return 1
     local git_reset_head cmd files opts rootdir
-    git_reset_head="git reset -q $FORGIT_RESET_HEAD_GIT_OPTS HEAD"
+    _forgit_reset_head_git_opts=()
+    _forgit_parse_array _forgit_reset_head_git_opts "$FORGIT_RESET_HEAD_GIT_OPTS"
+    git_reset_head="git reset -q ${_forgit_reset_head_git_opts[*]} HEAD"
     [[ $# -ne 0 ]] && { $git_reset_head "$@" && git status --short; return $?; }
     rootdir=$(git rev-parse --show-toplevel)
     cmd="git diff --staged --color=always -- $rootdir/{} | $_forgit_diff_pager "
@@ -232,7 +250,9 @@ _forgit_stash_show() {
     local git_stash_show git_stash_list cmd opts
     git_stash_show="git stash show --color=always --ext-diff"
     [[ $# -ne 0 ]] && { $git_stash_show "$@"; return $?; }
-    git_stash_list="git stash list $FORGIT_STASH_SHOW_GIT_OPTS"
+    _forgit_stash_show_git_opts=()
+    _forgit_parse_array _forgit_stash_show_git_opts "$FORGIT_STASH_SHOW_GIT_OPTS"
+    git_stash_list="git stash list ${_forgit_stash_show_git_opts[*]}"
     cmd="echo {} |cut -d: -f1 |xargs -I% $git_stash_show % |$_forgit_diff_pager"
     opts="
         $FORGIT_FZF_DEFAULT_OPTS
@@ -252,7 +272,9 @@ _forgit_stash_show() {
 _forgit_stash_push() {
     _forgit_inside_work_tree || return 1
     local git_stash_push msg args
-    git_stash_push="git stash push $FORGIT_STASH_PUSH_GIT_OPTS"
+    _forgit_stash_push_git_opts=()
+    _forgit_parse_array _forgit_stash_push_git_opts "$FORGIT_STASH_PUSH_GIT_OPTS"
+    git_stash_push="git stash push ${_forgit_stash_push_git_opts[*]}"
     args=( "$@" )
     while (( "$#" )); do
         case "$1" in
@@ -291,7 +313,9 @@ _forgit_clean() {
     _forgit_inside_work_tree || return 1
     _forgit_contains_non_flags "$@" && { git clean -q "$@"; return $?; }
     local git_clean files opts
-    git_clean="git clean $FORGIT_CLEAN_GIT_OPTS"
+    _forgit_clean_git_opts=()
+    _forgit_parse_array _forgit_clean_git_opts "$FORGIT_CLEAN_GIT_OPTS"
+    git_clean="git clean ${_forgit_clean_git_opts[*]}"
     opts="
         $FORGIT_FZF_DEFAULT_OPTS
         -m -0
@@ -307,7 +331,9 @@ _forgit_clean() {
 _forgit_cherry_pick() {
     local git_cherry_pick base target preview opts fzf_selection fzf_exitval
 
-    git_cherry_pick="git cherry-pick $FORGIT_CHERRY_PICK_GIT_OPTS"
+    _forgit_cherry_pick_git_opts=()
+    _forgit_parse_array _forgit_cherry_pick_git_opts "$FORGIT_CHERRY_PICK_GIT_OPTS"
+    git_cherry_pick="git cherry-pick ${_forgit_cherry_pick_git_opts[*]}"
 
     base=$(git branch --show-current)
     [[ -z "$base" ]] && echo "Current commit is not on a branch." && return 1
@@ -387,7 +413,9 @@ _forgit_cherry_pick_from_branch() {
 _forgit_rebase() {
     _forgit_inside_work_tree || return 1
     local git_rebase cmd preview opts graph files target_commit prev_commit
-    git_rebase="git rebase -i $FORGIT_REBASE_GIT_OPTS"
+    _forgit_rebase_git_opts=()
+    _forgit_parse_array _forgit_rebase_git_opts "$FORGIT_REBASE_GIT_OPTS"
+    git_rebase="git rebase -i ${_forgit_rebase_git_opts[*]}"
     graph=--graph
     [[ $FORGIT_LOG_GRAPH_ENABLE == false ]] && graph=
     cmd="git log $graph --color=always --format='$_forgit_log_format' $* $_forgit_emojify"
@@ -412,7 +440,9 @@ _forgit_fixup() {
     _forgit_inside_work_tree || return 1
     git diff --cached --quiet && echo 'Nothing to fixup: there are no staged changes.' && return 1
     local git_fixup cmd preview opts graph files target_commit prev_commit
-    git_fixup="git commit --fixup $FORGIT_FIXUP_GIT_OPTS"
+    _forgit_fixup_git_opts=()
+    _forgit_parse_array _forgit_fixup_git_opts "$FORGIT_FIXUP_GIT_OPTS"
+    git_fixup="git commit --fixup ${_forgit_fixup_git_opts[*]}"
     graph=--graph
     [[ $FORGIT_LOG_GRAPH_ENABLE == false ]] && graph=
     cmd="git log $graph --color=always --format='$_forgit_log_format' $* $_forgit_emojify"
@@ -439,7 +469,9 @@ _forgit_fixup() {
 _forgit_checkout_file() {
     _forgit_inside_work_tree || return 1
     local git_checkout cmd files opts
-    git_checkout="git checkout $FORGIT_CHECKOUT_FILE_GIT_OPTS"
+    _forgit_checkout_file_git_opts=()
+    _forgit_parse_array _forgit_checkout_file_git_opts "$FORGIT_CHECKOUT_FILE_GIT_OPTS"
+    git_checkout="git checkout ${_forgit_checkout_file_git_opts[*]}"
     [[ $# -ne 0 ]] && { $git_checkout -- "$@"; return $?; }
     cmd="git diff --color=always -- {} | $_forgit_diff_pager"
     opts="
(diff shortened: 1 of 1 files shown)


sandr01d:
The failing CI test is due to #327

bin/git-forgit
```diff
@@ -602,18 +613,14 @@ _forgit_blame() {
...
-            echo File not tracked
-        fi
-    "
-    file=$(FZF_DEFAULT_OPTS="$opts" fzf --preview="$preview")
+    ${IFS+"false"} && unset old_IFS || old_IFS="$IFS"
+    IFS=$'\n'
+    #shellcheck disable=2207
+    flags=($(git rev-parse --flags "$@"))
+    ${old_IFS+"false"} && unset IFS || IFS="$old_IFS"
+    file=$(FZF_DEFAULT_OPTS="$opts" fzf --preview="$FORGIT blame_preview {} ${flags[*]}")
```
carlfriedrich, line 621: I would propose moving this into a function, e.g. `_forgit_extract_flags`.
As this is a bugfix, though, I would transfer the change into a follow-up bugfix PR to keep this refactoring PR clean.

carlfriedrich (changes requested):
@sandr01d Great work, thanks for starting on this! I would propose implementing this change for all forgit functions within this single PR and move all bugfixing stuff to later PRs. What do you think?

bin/git-forgit
sandr01d: This does make sense to me. I just looked into this and actually the `flags` variable has to be an array so that we can get rid of the `eval` in the last line. This is one of the reasons I made this change. I forgot to make that clear in the description. Because getting rid of `eval` usage is one of the goals of this PR and I think we should keep it as is and I will do a follow up PR that moves this into a `_forgit_extract_flags` function. Is that fine with you?

carlfriedrich: I see, thanks for the explanation. I'm fine with that.

sandr01d:
Since there haven't been any objections so far, I'm moving this forward and started refactoring the other functions. If anyone wants to jump in feel free to do so and send a PR to this branch.

sandr01d:
A note regarding `_forgit_revert`: I've removed passing the `files` variable to the preview, as well as passing parameters to the git log command with `$*`.  These were creating issues in the new implantation and they were actually always empty, because `_forgit_revert` exits early when any parameters are passed. I copied them by mistake when I implemented `_forgit_revert`, they did never actually serve any purpose.

I'll keep collecting implementation details as comments here, so we can combine them in a commit message when we merge this.

carlfriedrich:
> Since there haven't been any objections so far, I'm moving this forward and started refactoring the other functions. If anyone wants to jump in feel free to do so and send a PR to this branch.

@sandr01d Thanks a lot for keeping this going! I'll probably have some time by the end of next week and am motivated to take over some functions as well.

I will switch forgit to this branch for my daily work from now on in order to find regressions early while we're going. @cjappl @wfxr You maybe too?

bin/git-forgit
```diff
@@ -539,22 +540,27 @@ _forgit_checkout_commit() {
...
-        --preview=\"$preview\"
         $FORGIT_BRANCH_DELETE_FZF_OPTS
     "
 
-    cmd="git branch --color=always | LC_ALL=C sort -k1.1,1.1 -rs"
-    branches=$(eval "$cmd" | FZF_DEFAULT_OPTS="$opts" fzf | awk '{print $1}')
+    branches=$(
+        git branch --color=always |
+        LC_ALL=C sort -k1.1,1.1 -rs |
+        FZF_DEFAULT_OPTS="$opts" fzf --preview="$FORGIT branch_delete_preview {1}" |
```
carlfriedrich: Hey @sandr01d, is there a reason why you moved the `--preview` option from the `opts` variable to the `fzf` argument list? This way we cannot override the preview in `FORGIT_BRANCH_DELETE_FZF_OPTS` anymore.

sandr01d:
No, that appears to be a mistake on my side and we should move it back. Good catch! I don't have time to work on this today or tomorrow, so feel free to push a commit to this PR. Otherwise I'll change it back the next time I get around to it (probably Saturday).

bin/git-forgit
sandr01d: No, that appears to be a mistake on my side and we should move it back. Good catch! I don't have time to work on this today or tomorrow, so feel free to push a commit to this PR. Otherwise I'll change it back the next time I get around to it (probably Saturday).

carlfriedrich:
Notes to `_forgit_log`:

1. Not sure how to correctly fix the shellcheck. When I quote `$graph` it does not show anything in my case.
2. I think the code was (and is) meant to show only the files in the preview that have been passed as arguments. This does not work, though, but it did not work before as well.

carlfriedrich:
I found a way to correctly use the `$graph` variable. Still not sure how we correctly handle `$FORGIT_LOG_GIT_OPTS`, though. And files still not work either.

carlfriedrich:
So I came to the conclusion that we just have to disable the specific shellcheck for passing `$FORGIT_LOG_GIT_OPTS` to the `git` call. The variable will be a space separated string, and thus we have to use it like that. Within bash, the clean way would be to use an array. This does not work for other shells, though. Any other thoughts on this?

And I think I have finished `_forgit_log` now. Passing files actually works when using `--` (e.g. `git forgit log -- README.md). @sandr01d Can you do a quick review of my implementation?

sandr01d:
> So I came to the conclusion that we just have to disable the specific shellcheck for passing `$FORGIT_LOG_GIT_OPTS` to the `git` call. The variable will be a space separated string, and thus we have to use it like that. Within bash, the clean way would be to use an array. This does not work for other shells, though. Any other thoughts on this?
> 
> And I think I have finished `_forgit_log` now. Passing files actually works when using `--` (e.g. `git forgit log -- README.md). @sandr01d Can you do a quick review of my implementation?

Your code looks good to me. I gave it a quick test and everything works as expected so far. The ability to pass in files is very neat, great work!
Regarding the shellcheck, when we change the line to this it does not seem to complain:

```bash
git log "${graph[@]}" --color=always --format="$log_format" ${FORGIT_LOG_GIT_OPTS:+$FORGIT_LOG_GIT_OPTS} "$@" ${_forgit_emojify:+$_forgit_emojify}
```

But that is probably just hiding the issue and not doing anything better. Word splitting is intentional here so from my perspective disabling shellcheck as you do is fine. I think we should probably disable globbing and restore it after the line.

bin/git-forgit
sandr01d: I've moved the `--preview` option back into the `opts` variable where applicable. I think we should also do so for the functions where it never was in the `opts` variable (`_forgit_blame`, `_forgit_revert_commit` and `_forgit_stash_push`), but I would like to keep this PR focused on removing deferred execution and will do so in a follow up PR.

sandr01d:
@carlfriedrich I noticed you changed the default of `FORGIT_LOG_GRAPH_ENABLE` to true in 521f349. I think this was by mistake, so I changed it back in 0fdbda3. Let me know in case this was intentional.

carlfriedrich:
> @carlfriedrich I noticed you changed the default of `FORGIT_LOG_GRAPH_ENABLE` to true in [521f349](https://github.com/wfxr/forgit/commit/521f3494025671deebbe7c1315c7fac1d4cec77e). I think this was by mistake, so I changed it back in [0fdbda3](https://github.com/wfxr/forgit/commit/0fdbda300fb6c5e90cd87f75587fb237b7c8d166). Let me know in case this was intentional.

Actually I did not change the default value, I just switched the logic of evaluation and transferred the default value to the new variable `_forgit_log_graph_enable` because IMO that made the code more readable and consistent. I think you changed the default value to true now?

> ```shell
> git log "${graph[@]}" --color=always --format="$log_format" ${FORGIT_LOG_GIT_OPTS:+$FORGIT_LOG_GIT_OPTS} "$@" ${_forgit_emojify:+$_forgit_emojify}
> ```
> 
> But that is probably just hiding the issue and not doing anything better. Word splitting is intentional here so from my perspective disabling shellcheck as you do is fine. I think we should probably disable globbing and restore it after the line. I ran into the same issue with `_forgit_fixup` and `_forgit_rebase` and decided to also simply disable shellcheck there until we settled on a solution.

Yes, that's definitley just a workaround and would clutter the code unnecessarily IMO. Disable globbing sounds reasonable. Maybe we could parse `FORGIT_LOG_GIT_OPTS` to an array `_forgit_log_git_opts` at the beginning of the code (and for all other git options variables as well) to have that in one single place. This would make the evaluation of user-configurable variables more consistent as well.

bin/git-forgit
```diff
@@ -386,66 +409,79 @@ _forgit_cherry_pick_from_branch() {
...
+_forgit_checkout_file_preview() {
+    git diff --color=always -- "$1" | $_forgit_diff_pager
 }
 
 # git checkout-file selector
 _forgit_checkout_file() {
     _forgit_inside_work_tree || return 1
-    local git_checkout cmd files opts
+    local git_checkout files opts
     git_checkout="git checkout $FORGIT_CHECKOUT_FILE_GIT_OPTS"
```
carlfriedrich, line 479: @sandr01d In `_forgit_checkout_file` we still have a deferred code line for the git command, which we're using for returning early as well as the action after the fzf dialog. This is common practice in other forgit functions as well. Do we want to keep this? IMO it makes sense, because it reduces code duplication. It also solves the problem with the options variable, so for consistency we could do this in every function and wouldn't need to disable the shellcheck.
Cleaner, however, would be defining a function for it. What do you think?

sandr01d:
> Actually I did not change the default value, I just switched the logic of evaluation and transferred the default value to the new variable _forgit_log_graph_enable because IMO that made the code more readable and consistent. I think you changed the default value to true now?

I did, but it was implicitly true by default before your change. Previously we only removed the `--graph` option when `FORGIT_LOG_GRAPH_ENABLE` was explicitly set to false and kept it when it was unset.

> Disable globbing sounds reasonable. Maybe we could parse FORGIT_LOG_GIT_OPTS to an array _forgit_log_git_opts at the beginning of the code (and for all other git options variables as well) to have that in one single place. This would make the evaluation of user-configurable variables more consistent as well.

That does sound like a good approach to me.

bin/git-forgit
sandr01d: I would prefer to use functions here a as well.  When we parse the options environment variables into an array as you mentioned earlier we don't even have to disable the shellcheck and I find functions more readable and less error prone to deferred execution.

carlfriedrich:
> I did, but it was implicitly true by default before your change. Previously we only removed the `--graph` option when `FORGIT_LOG_GRAPH_ENABLE` was explicitly set to false and kept it when it was unset.

Oh I see! Sorry, that was a misunderstanding. Thanks for correcting it!

bin/git-forgit
carlfriedrich: Alright, then let's go with functions here as well!

carlfriedrich:
@sandr01d I refactored most of `_forgit_diff`. Can you check if `alt+e` on a diffed file still works for you?

sandr01d:
> @sandr01d I refactored most of `_forgit_diff`. Can you check if `alt+e` on a diffed file still works for you?

Works for me

bin/git-forgit
```diff
@@ -78,42 +86,99 @@ _forgit_ignore_pager=${FORGIT_IGNORE_PAGER:-$(hash bat &>/dev/null && echo 'bat
...
+    # shellcheck disable=SC2086 # split FORGIT_DIFF_GIT_OPTS on purpose
+    git diff --color=always $FORGIT_DIFF_GIT_OPTS "$@"
+}
+
+_forgit_diff_view() {
+    local input_line=$1
+    local diff_context=$2
+    local commits="${*:3}"
+    echo "$input_line" | _forgit_get_files_from_diff_line | xargs -0 \
+        "$FORGIT" exec_diff ${commits:+$commits} -U"$diff_context" -- | $_forgit_diff_pager
```
sandr01d, line 169: @carlfriedrich Would it be possible to keep these quoted?

```suggestion
        "$FORGIT" exec_diff ${commits:+"$commits"} -U"$diff_context" -- | $_forgit_diff_pager
```

bin/git-forgit
```diff
@@ -78,42 +86,99 @@ _forgit_ignore_pager=${FORGIT_IGNORE_PAGER:-$(hash bat &>/dev/null && echo 'bat
...
     files=$(sed -nE 's/.*-- (.*)/\1/p' <<< "$*") # extract files parameters for `git show` command
-    preview_cmd="echo {} | $_forgit_extract_sha | xargs -I% git show --color=always -U$_forgit_preview_context % -- $files | $_forgit_show_pager"
-    enter_cmd="echo {} | $_forgit_extract_sha | xargs -I% ${FORGIT} diff %^! $files"
     opts="
         $FORGIT_FZF_DEFAULT_OPTS
         +s +m --tiebreak=index
-        --bind=\"enter:execute($enter_cmd)\"
-        --bind=\"ctrl-y:execute-silent(echo {} | $_forgit_extract_sha | ${FORGIT_COPY_CMD:-pbcopy})\"
-        --preview=\"$preview_cmd\"
+        --bind=\"enter:execute($FORGIT log_enter \"{}\" $files)\"
```
sandr01d, line 121: `$files` is unquoted here. I think we could use an array as we do in `_forgit_rebase` and quote it.

bin/git-forgit
carlfriedrich: No, `$commits` can contain one or two commits, and if it's two commits it has to be two separate arguments. But I thought about this before as well, and maybe we can also use an array here. I will check this.

sandr01d:
Notes regarding d23ec19:

I've exposed the new functions  `git_branch_delete` and `git_checkout_commit` as forgit commands, so they can be used with `xargs`.
I'm parsing the environment variables into arrays using `read -r -a`. The `-a` option does not exist in zsh, but from my understanding we always use bash for executing /bin/git-forgit since #241. I've tested it with bash, zsh and fish and it appears to be working fine.
@carlfriedrich could you do a quick review of these changes?

bin/git-forgit
sandr01d: With `_forgit_parse_array` added we should now be able to solve it like this

```bash
local commits=()
_forgit_parse_array commits "${*:3}"
echo "$input_line" | _forgit_get_files_from_diff_line | xargs -0 \
        "$FORGIT" exec_diff "${commits[@]}" -U"$diff_context" -- | $_forgit_diff_pager
```

sandr01d:
> This will not run on MacOS's bash, unfortunately, because MacOS uses ancient bash version 3.2.57 (I think due to licensing issues). Is there a different way to implement this?

Ah, I see thanks for the clarification. We could of course just set the IFS, parse all the arrays and unset it again like this:

```sh
# set IFS
${IFS+"false"} && unset old_IFS || old_IFS="$IFS"
# parse all the arrays here
_forgit_log_git_opts=()
IFS=" " read -r -a _forgit_log_git_opts <<< "$FORGIT_LOG_GIT_OPTS"
_forgit_diff_git_opts=()
IFS=" " read -r -a _forgit_diff_git_opts <<< "$FORGIT_DIFF_GIT_OPTS"
# ...
# reset IFS
${old_IFS+"false"} && unset IFS || IFS="$old_IFS"
```

But this is neither reusable nor pretty. I am not aware of a way to make this work in a function without a nameref, but I will do further research. I probably won't have time to work on it this week. In case you come up with something, let me know or feel free to push it to this branch.

sandr01d:
@carlfriedrich I've removed the nameref from `_forgit_parse_array` in 88d928d. The downside of this is that arrays parsed with this function now are always global variables. I would be okay with this tough. Would love to hear your thoughts on this.

sandr01d:
Note regarding ce37644:

I changed how we parse file arguments for passing them along to the preview functions. I've added `_forgit_escaped_files` for this. The only difference in behavior to the `sed` command we had previously is that the string that is printed out has each file name wrapped in single quotes, so they can be used safely without having to worry about globbing or word splitting. There is no need to quote this string when passing it to a preview function, because the individual file names inside it are already properly quoted. I changed the `files` variable names to `escaped_files` to indicate this.

bin/git-forgit
```diff
@@ -78,86 +174,131 @@ _forgit_ignore_pager=${FORGIT_IGNORE_PAGER:-$(hash bat &>/dev/null && echo 'bat
...
+    quoted_files=$(_forgit_quote_files "$@")
     opts="
         $FORGIT_FZF_DEFAULT_OPTS
         +s +m --tiebreak=index
-        --bind=\"enter:execute($enter_cmd)\"
-        --bind=\"ctrl-y:execute-silent(echo {} | $_forgit_extract_sha | ${FORGIT_COPY_CMD:-pbcopy})\"
-        --preview=\"$preview_cmd\"
+        --bind=\"enter:execute($FORGIT log_enter \"{}\" $quoted_files)\"
+        --bind=\"ctrl-y:execute-silent($FORGIT yank_sha {})\"
+        --preview=\"$FORGIT log_preview \"{}\" $quoted_files\"
```
sandr01d, line 211: @carlfriedrich Is the extra quotation of `{}` necessary here? `fzf`s documentation [states](https://github.com/junegunn/fzf#preview-window) that "{} is replaced with the single-quoted string of the focused line", so I think we should be fine without it.

carlfriedrich:
@sandr01d Thanks again for keeping up the good work on this!

> @carlfriedrich I've removed the nameref from `_forgit_parse_array` in [88d928d](https://github.com/wfxr/forgit/commit/88d928df55d1703c65955056a475bf54cf9a3705). The downside of this is that arrays parsed with this function now are always global variables. I would be okay with this tough. Would love to hear your thoughts on this.

That is totally fine for me. I don't see a problem having global variables here.

> I changed how we parse file arguments for passing them along to the preview functions. I've added `_forgit_escaped_files` for this. The only difference in behavior to the `sed` command we had previously is that the string that is printed out has each file name wrapped in single quotes, so they can be used safely without having to worry about globbing or word splitting. There is no need to quote this string when passing it to a preview function, because the individual file names inside it are already properly quoted. I changed the `files` variable names to `quoted_files` to indicate this. I make use of it in `_forgit_log` where file names were not quoted previously and in `_forgit_rebase` and `_forgit_fixup` where all file names were quoted together (e.g. `"LICENSE README.md"` instead of `"LICENSE" "README.md"`) by accident in one of my earlier commits in this PR.

Great work, and I also appreciate the rename to `forgit_quote_files`.

Are there any functions left to refactor now?

I am a little bit concerned about the size of the diff on this branch. We have changed quite a lot, which makes reviewing the code quite hard. Do you see a useful way of semantically sqashing commits? Can we maybe add an overview of all the things we did in this PR to the PR description?

I will continue to use the branch in my daily work, but I am not using git as heavily as in the past right now, so stumbling on bugs might take some time.

bin/git-forgit
carlfriedrich: That indeed reads like we don't need the quotes.

sandr01d:
> Are there any functions left to refactor now?

No, we're all set and from my side this would now be ready for review.

> I am a little bit concerned about the size of the diff on this branch. We have changed quite a lot, which makes reviewing the code quite hard. Do you see a useful way of semantically sqashing commits?

I agree with you. I did not expect it to get this large when we started out with this. We could rebase and combine all commits that belong to a single forgit function. We would loose some of the history though and I'm not sure if that's worth it. I will mark this PR ready-for-review once we decided on this.

> Can we maybe add an overview of all the things we did in this PR to the PR description?

Done.



> I will continue to use the branch in my daily work, but I am not using git as heavily as in the past right now, so stumbling on bugs might take some time.

Given the size of this PR I will definitively leave it open for review for some time. I've also been using this branch for my daily work since I've created it. There are some functions that I rarely need, e.g. `gi` and `grb` and I will give them some extra testing aswell.

carlfriedrich:
> No, we're all set and from my side this would now be ready for review.

Great!

> We could rebase and combine all commits that belong to a single forgit function. We would loose some of the history though and I'm not sure if that's worth it. I will mark this PR ready-for-review once we decided on this.

Usually we're squash-merging the PRs in this repo, so the history would be gone anyway. On such a huge change, however, it might be useful to keep the history, indeed. I am quite unsure about this. 55 commits in a single PR seems a bit overkill for sure, IMO we could at least squash the "fix shellcheck" commits into their parents, what do you think?

> > Can we maybe add an overview of all the things we did in this PR to the PR description?
> Done.

Great, thanks a lot!

> Given the size of this PR I will definitively leave it open for review for some time. I've also been using this branch for my daily work since I've created it. There are some functions that I rarely need, e.g. `gi` and `grb` and I will give them some extra testing aswell.

That definitley sounds reasonable.

sandr01d:
> Usually we're squash-merging the PRs in this repo, so the history would be gone anyway. On such a huge change, however, it might be useful to keep the history, indeed. I am quite unsure about this. 55 commits in a single PR seems a bit overkill for sure, IMO we could at least squash the "fix shellcheck" commits into their parents, what do you think?

I think there was a misunderstanding here, I thought you meant squashing in this branch before we start the review process. I would actually squash the commits when merging with the usual "squash and merge" option, but keep them as is in the current branch. This way we kind of have both, a clean history on the master branch and the full history inside this PR. Does that sound reasonable to you?

carlfriedrich:
> I think there was a misunderstanding here, I thought you meant squashing in this branch before we start the review process. I would actually squash the commits when merging with the usual "squash and merge" option, but keep them as is in the current branch. This way we kind of have both, a clean history on the master branch and the full history inside this PR. Does that sound reasonable to you?

No, you actually got it right: I would really like to review this PR in detail, but given the fact that it is so large and so many commits, it is really hard to keep an eye on everything. That's why I am wondering how we could simplify the process. Squashing commits that semantically belong together is one option IMO, then we could review commit by commit. Maybe we could also split it into multiple PRs, then we could leave this branch untouched.
However, I am not sure if this would cause all more work than necessary and we'd rather just test it in real life and then merge it as is.
What do you think?

carlfriedrich:
The longer I think and [read about it](https://ieftimov.com/posts/how-to-make-reviewers-love-your-big-pull-requests/), the more I am sure that we should squash commits semantically.
Since you already spent a lot time refactoring, I can imagine that re-organizing commits doesn't sound like a motivating task.
I can offer that I try to combine commits on a separate branch, which would also include reviewing them. That will take some time, though, since I am quite busy at the moment.
What do you think?

sandr01d:
> The longer I think and [read about it](https://ieftimov.com/posts/how-to-make-reviewers-love-your-big-pull-requests/), the more I am sure that we should squash commits semantically. Since you already spent a lot time refactoring, I can imagine that re-organizing commits doesn't sound like a motivating task. I can offer that I try to combine commits on a separate branch, which would also include reviewing them. That will take some time, though, since I am quite busy at the moment. What do you think?

My main concern was loosing the original commit history, but I am not completely opposed to squashing commits semantically and I think you're probably right, that it could ease the review process. If we do so, squashing everything together that belongs to a single function would make sense to me. Don't worry about the additional workload, I'm fine with doing it. Given that I will be quite busy the next week too, I probably won't get to it before the next weekend.

carlfriedrich:
> My main concern was loosing the original commit history, but I am not completely opposed to squashing commits semantically and I think you're probably right, that it could ease the review process. If we do so, squashing everything together that belongs to a single function would make sense to me. Don't worry about the additional workload, I'm fine with doing it. Given that I will be quite busy the next week too, I probably won't get to it before the next weekend.

Thanks for your understanding and your will to invest more time. You can still keep the original history by copying your branch before squashing anything. Or you leave the original branch as it is, modify the copied branch and open a new (or multiple) tidied-up PRs for it.

sandr01d:
@carlfriedrich I've now squashed commits together that belong to the same functions. Could you take a quick look and let me know whether that works for you? If so, we are ready start the review process.
The original history can still be found [here](https://github.com/sandr01d/forgit/tree/reduce-deferred-exec-orig).

carlfriedrich:
@sandr01d Thanks a lot for your further work on this. 22 commits is definitely a lot better than 55, and reviewing this PR commit by commit seems possible to me now. I will try to manage this during the following few weeks, okay?

sandr01d:
> @sandr01d Thanks a lot for your further work on this. 22 commits is definitely a lot better than 55, and reviewing this PR commit by commit seems possible to me now. I will try to manage this during the following few weeks, okay?

Great, looking forward to your review. In retrospect, cleaning up the commit history here was the right decision. Thanks for convincing me.

sandr01d:
This PR is now ready for review. I've been daily driving this branch since we started working on it and in the past weeks things have been perfectly stable for me.
@cjappl This branch has not seen any testing on macOS yet, so it would be especially interesting to know whether everything works as expected for you too.

cjappl:
@sandr01d downloaded and will test for a little bit. I think I would want to daily drive it for a little considering how extensive the change is.

I'll keep you posted.

cjappl:
In general, my testing environment should be assumed to be:
MacOS
Fish shell


First issue: Loss of scrolling with alt+j alt+k in the preview window of `ga`

1. Change a file
2. `ga`
3. alt-j to scroll in the preview window (option+j) on mac

Expected:
Preview window scrolls up and down

Observed:
Preview window is static

cjappl:
Taking it through the actions I use the very most: `ga`, `gcf`, `grh`, `glo`, `gct`, `gco` `gcb` everything is working A-OK! Great work wrangling this complexity, beast of a PR.

I would propose we all drive this branch for a week or so and if no one has hit anything by Feb 5th, we submit.

Thoughts @sandr01d @carlfriedrich ?

carlfriedrich:
@cjappl I would actually like to do a code review on this before it gets merged.

cjappl:
> @cjappl I would actually like to do a code review on this before it gets merged.

🤦 whoops, yeah and that! That seems important. I skipped over that pretty crucial step lol.

Code review + drive for a bit :)

carlfriedrich:
I just took a first quick look at the commit history.

@sandr01d IMO the following functions should have dedicated commits as well:

- `_forgit_extract_sha`
- `_forgit_is_file_tracked`

Right now these are replaced piece by piece with each commit. I think each function should have its own commit (just like `_forgit_emojify`), and they should be the first in the commit stack, since they are helper functions. Would you mind  extracting these changes? Otherwise I could do it. What do you think?

cjappl:
> I just took a first quick look at the commit history.
> 
> @sandr01d IMO the following functions should have dedicated commits as well:
> 
>     * `_forgit_extract_sha`
> 
>     * `_forgit_is_file_tracked`
> 
> 
> Right now these are replaced piece by piece with each commit. I think each function should have its own commit (just like `_forgit_emojify`), and they should be the first in the commit stack, since they are helper functions. Would you mind extracting these changes? Otherwise I could do it. What do you think?

I would prefer we "squash and merge" by default in this repo to keep the history clean, making the individual commits irrelevant.

carlfriedrich:
@cjappl We already discussed this earlier in this thread:

> Usually we're squash-merging the PRs in this repo, so the history would be gone anyway. On such a huge change, however, it might be useful to keep the history, indeed.

I still think that we should keep the commit history here, given the size of the change. So I would vote for either a merge commit or merge by rebase.

Either way, separating commits into semantical entities will make the review process easier.

cjappl:
Cool, works for me! sorry for missing that. In that case I agree, abstracting out the individual changes would be good.

carlfriedrich:
> @sandr01d IMO the following functions should have dedicated commits as well:
> 
>     * `_forgit_extract_sha`
>     * `_forgit_is_file_tracked`

Digging deeper into the implementation, I noticed that my suggestion is not possible because these functions are used in preview expressions, so they would be unknown to the child process. Sorry for the noise. :-)

I still have the feeling that there is potential of extracting more atomic changes out of the PR, though, so I might come up with other proposals during the review.

carlfriedrich:
So going through the single commits I see that we're still doing many things at once here. Which is good for the end result, but hard to capture all at once when reading the changes.

First of all: while this is tagged as a refactoring PR, it also fixes a few bugs along the way. These should definitely appear in the commit messages in the end, because we're using the commit history for our change log in the release notes.

To make this easier and more transparent, I would suggest to spot changes forming a logic entity and extract them into a separate PR, then rebasing this PR on top of that.

The goals of this are:
- Give forgit users the chance to understand what's happening in the changelog
- Track bug fixes
- Make this PR smaller
- Get better code

I would furthermore suggest to do this one after another, i.e.:
1. Spot one logic entity
2. Extract it to a dedicated PR
3. Rebase this PR
4. Go back to 1

We do this until this PR cannot be split to entities anymore. This requires some more rebasing effort, but I am convinced that it will make both the changelog and the code better and more readable. Does that sound like a good plan @sandr01d? I am happy to help doing all the rebasing, if you are not willing to put even more work into this. In order not to need to test all the PRs separately, I think we could stack them up and merge them all at once in the end.

The first entity I see we could extract is parsing all the `FORGIT_*_GIT_OPTS` into arrays using the new `_forgit_parse_array` function. I see this as an atomic change that is independent of the rest, and it is one that already includes a bugfix.

WDYT?

sandr01d:
Doing the review process in separate smaller chunks like this sounds reasonable to me. Do I understand it correctly, that we'd focus on a single PR and only move forward with rebasing this branch and create the next PR once we're happy with the changes in the chunk? I think that splitting this PR into multiple PRs all at once would create a lot of friction. I expect doing it this way will take some time (there will be 22+ separate PRs), but we're not in a hurry.

> In order not to need to test all the PRs separately, I think we could stack them up and merge them all at once in the end.

Could you please elaborate on this? Would we merge this PR once we've rebased upon all the smaller PRs and close down the smaller PRs?

> The first entity I see we could extract is parsing all the FORGIT_*_GIT_OPTS into arrays using the new _forgit_parse_array function.

That makes sense and I can prepare a PR for that. I don't see a way to include the usage of the arrays we create with this though. Those are used exclusively in the `forgit_git_*` functions which were created afterwards. Would a PR that only adds the `_forgit_parse_array` function and the parsing itself be ok?

carlfriedrich:
> Doing the review process in separate smaller chunks like this sounds reasonable to me.

Thanks for understanding!

> Do I understand it correctly, that we'd focus on a single PR and only move forward with rebasing this branch and create the next PR once we're happy with the changes in the chunk?

Yes, that's exactly how I meant it.

> I think that splitting this PR into multiple PRs all at once would create a lot of friction. I expect doing it this way will take some time (there will be 22+ separate PRs), but we're not in a hurry.

Yes, it will take some time and chore work indeed, and I am happy to help with this. I am not sure if we really need that much PRs in the end, though. Let's go ahead and see how this PR shrinks on the way.

> > In order not to need to test all the PRs separately, I think we could stack them up and merge them all at once in the end.
> 
> Could you please elaborate on this? Would we merge this PR once we've rebased upon all the smaller PRs and close down the smaller PRs?

Exactly. When we prepare a chunk PR and merge it right away, we have to make sure that it is tested individually, which is needless work, given the fact that we already are ahead in the code. So we just leave it open and prepare the next PR based on the first PR's branch and then merge all of them when the task is completed.

> > The first entity I see we could extract is parsing all the FORGIT_*_GIT_OPTS into arrays using the new _forgit_parse_array function.
> 
> That makes sense and I can prepare a PR for that. I don't see a way to include the usage of the arrays we create with this though. Those are used exclusively in the `forgit_git_*` functions which were created afterwards. Would a PR that only adds the `_forgit_parse_array` function and the parsing itself be ok?

Thanks, that would be a good starting point. However, is there a reason why we couldn't put the arrays in the former deferred code snippets? IMO that would make a useful atomic commit.

sandr01d:
> Thanks, that would be a good starting point. However, is there a reason why we couldn't put the arrays in the former deferred code snippets? IMO that would make a useful atomic commit.

Nevermind, that is possible.

I think we're all set then, I've created #338  and am awaiting your reviews.

carlfriedrich:
@sandr01d Thanks for submitting the patch. This branch can be rebased now on top of #338. Do you need help with that?

The next atomic chunk I see is changing all of the git calls from variables into functions, e.g.

```shell
 git_blame="git blame ${_forgit_blame_git_opts[*]}"
```

into

```shell
_forgit_git_blame() {
    git blame "${_forgit_blame_git_opts[@]}" "$@"
}
```

These functions are all called directly, so this should work independently of the rest. And as far as I can see this would be a no-op change as well, so refactoring only. Can you confirm this? Or does this change behavior in passing the command line opts?

sandr01d:
Alright, I've rebased on top of master now and we're ready to start testing. Let me know when whether you find any issues.
@cjappl Could you pease double check whether your fix from #365 is still working as intended. It should (and I can confirm it does on Linux), but I want to be absolutely sure.

In regards to merging, I would strongly recommend we merge this PR (without squashing it) once it's tested. This way, we can be absolutely certain that the final state resembles exactly what we've been testing and we avoid having to rebase and merge 11 PRs and any mistakes that might arise by doing so.

bin/git-forgit
```diff
@@ -137,27 +197,81 @@ _forgit_diff() {
...
+    for commit in "${commits[@]}"; do
+        escaped_commits+="'${commit//\{/\\\\\{}' "
+    done
     opts="
         $FORGIT_FZF_DEFAULT_OPTS
-        +m -0 --bind=\"enter:execute($enter_cmd | $_forgit_enter_pager)\"
-        --preview=\"$preview_cmd\"
-        --bind=\"alt-e:execute-silent($EDITOR \\\"\$\($get_file)\\\" >/dev/tty </dev/tty)+refresh-preview\"
+        +m -0 --bind=\"enter:execute($FORGIT diff_enter {} $escaped_commits | $_forgit_enter_pager)\"
+        --preview=\"$FORGIT diff_view {} $_forgit_preview_context $escaped_commits\"
```
sandr01d, line 266:
```suggestion
        --preview=\"$FORGIT diff_view {} $_forgit_preview_context $escaped_commits\"
        --preview=\"$FORGIT diff_view {} \"$_forgit_preview_context\" $escaped_commits\"
```

`$_forgit_preview_context` is still not escaped here. Should we fix it in this PR or in a follow up? WDYT?

cjappl:
Reporting back on mac, using fish shell. @sandr01d 


# Adding files
1. ✅ Adding file[with]brackets
2. ✅ adding "file with spaces.txt"
3. ✅ adding normal file

# Reset head
1. ✅ file[with]brackets
2. ✅ "file with spaces.txt"
3. ✅ normal file


# Git log
1. ✅ without arguments
2. ✅ with branch name as argument
3. ✅ with file name as argument

# Diffing files
1. ✅ Diffing file[with]brackets
2. ✅ Diffing "file with spaces.txt"
3. ✅ Diffing normal file

# Git ignore
1. ✅ normal operation

# Checkout file
1. ✅ file[with]brackets
2. ✅ "file with spaces.txt"
3. ✅ normal file

# Checkout branch
1. ✅ Local branch
2. ✅ Remote branch
3. ✅ Local branch as argument
4.  ✅ New branch to be created

# Delete branch
1. ✅ normal operation
2. ✅ With branch as argument

# Git clean
1. ✅ Adding dummy file, then deleting it through `glcean`

# Git stash show/git stash push
1. ✅ Normal operation adding things to the stash, dropping things from the stash
2. ❌ You are able to enter into `gsp` with no changes in the directory, I feel we should early exit/not drop into fzf in this case

> I don't know if this is a regression or not. If it is not a regression let's file a ticket and fix it later. Just basic UX stuff, nothing horrible.
> 
> Steps to repro:
> 1. Have clean directory, no changes no added files
> 2. `gsp`
> 
> Expected:
> Nothing happens, gsp exits early, just like if you `ga` with "Nothing to add"
> 
> Observed:
> Fzf pops up but the list is empty, user can do nothing.


# cherry pick
1. ✅ Normal operation with no arguments
2. ✅ WIth branch as argument

# Rebase
1. ✅ Squashing multiple commits together
2. ✅ Rewording commits
3. ✅ with branch name as argument

# Fixup
1. ❓ not sure how to test this one, open to ideas

# Checkout commit
1. ✅ With no arguments
4. ✅ WIth commit sha as argument

# Revert commit
1. ✅ With no arguments
2. ✅ WIth commit sha as argument

# Blame
1. ❌ Normal operation.

> Again, not sure if this is a regression. If it isn't, let's file a ticket and NOT fix it here.
> 
> Steps to repro:
> 1. Navigate to top level of our forgit repo
> 2. `gbl`
> 
> Expected:
> Directories do not have a blame shown (what even is a git blame for a directory??)
> 
> Observed:
> Both files AND DIRECTORIES show up as blames. I think we need to filter out the directories
> Also: It would be nice if we could syntax highlight the blames, or put them through `bat` or our pager.

# Checkout tag
1. ✅ With no arguments
2. ✅ With tag as argument

cjappl:
Overall I give this a thumbs up on mac and fish on the basis of this "quick smoke test" this was all of the tests I could think to run, and everything seems generally solid.

I will continue playing with it and using it as my daily driver. Let me know if you saw any test cases I missed or something I should spend time focusing on. Great work @sandr01d , for such an insane code review very few regressions as far as I can tell

bin/git-forgit
cjappl: Either way fine with me

sandr01d:
@cjappl First of all, thanks  for the in depth testing!

> ### Git stash show/git stash push
> ❌ You are able to enter into gsp with no changes in the directory, I feel we should early exit/not drop into fzf in this case

I can reproduce this in this branch, but so  can I in the master branch, so not a regression. Definitively a bug though, I agree that we should open a separate issue.

> ### Fixup
>    ❓ not sure how to test this one, open to ideas

I think the easiest way to test this is to create a local repository, add a few files and create a few commits, then do a few changes, add them and use `gfu` afterwards. At least that is how I tested things when I was doing changes to it.

> ### Blame
> 
> ❌ Normal operation.
> 
> Again, not sure if this is a regression. If it isn't, let's file a ticket and NOT fix it here.
> 
> Steps to repro:
> 
> Navigate to top level of our forgit repo
> gbl
> 
> Expected:
> Directories do not have a blame shown (what even is a git blame for a directory??)
> 
> Observed:
> Both files AND DIRECTORIES show up as blames. I think we need to filter out the directories

This one I can not reproduce. I've checked our implementation on how we generate the file list for this command and it is by simply executing `fzf` in the directory like this:

```sh
file=$(FZF_DEFAULT_OPTS="$opts" fzf --preview="$FORGIT blame_preview {} ${flags[*]}")
```

What happens when you execute `fzf` in the repository by itself, do you also see directories? Anything in your `$FZF_DEFAULT_OPTS` or `$FORGIT_FZF_DEFAULT_OPTS` that might cause this behavior? Would also be worth checking how this behaves in the master branch on your side. I think you might have the same behavior there too.

> Also: It would be nice if we could syntax highlight the blames, or put them through bat or our pager.

I agree. Syntax highlighting would be super useful here.

> I will continue playing with it and using it as my daily driver. Let me know if you saw any test cases I missed or something I should spend time focusing on.

Would be great if you could play around the environment variables we allow customizing the git commands with, such as `FORGIT_BLAME_GIT_OPTS`. Also passing arguments to the git commands that support it like `gbl --color-by-age`.

cjappl:
>     Git stash show/git stash push
> 
>     ❌ You are able to enter into gsp with no changes in the directory, I feel we should early exit/not drop into fzf in this case
> 
> I can reproduce this in this branch, but so can I in the master branch, so not a regression. Definitively a bug though, I agree that we should open a separate issue 

Filed here https://github.com/wfxr/forgit/issues/369 

For the future, definitely does not have to be taken on until this big review lands

> Blame

Good intuition. This was a config issue on my side. Fixing my config fixed the issue, this is not a regression or anything to worry about.

> I agree. Syntax highlighting would be super useful here.

https://github.com/wfxr/forgit/issues/370 

For the future, definitely does not have to be taken on until this big review lands

> Would be great if you could play around the environment variables we allow customizing the git commands with, such as FORGIT_BLAME_GIT_OPTS. Also passing arguments to the git commands that support it like gbl --color-by-age.

Will put some time on my calendar to do this tomorrow, same with git fixup.


I have updated my big summary comment above with this info as well.

bin/git-forgit
carlfriedrich: Since we already started testing here, let's do it in a follow up PR.

carlfriedrich:
> In regards to merging, I would strongly recommend we merge this PR (without squashing it) once it's tested. This way, we can be absolutely certain that the final state resembles exactly what we've been testing and we avoid having to rebase and merge 11 PRs and any mistakes that might arise by doing so.

@sandr01d You're probably right, that seems to be the safest way. I think GitHub should actually automatically close the 11 other PRs when we merge this, since they contain the same commits as in this branch.

@cjappl Great test work, thanks a lot! I will do some testing during the next days as well.

carlfriedrich (approved):

cjappl:
I have completed my testing and deem this good to ship on fish shell and macOS. My complete testing is in a previous comment. (including test cases I didn't do before.

Let me know if there is anything else people can think to test that I have missed.

cjappl (approved):
🚀 🎸 🔥

bin/git-forgit
sandr01d: Sure, I'll send a follow up PR once this one is merged.

sandr01d:
Since we did not find any new bugs I'm planning to merge this on Sunday when I don't hear any objections.

carlfriedrich:
@sandr01d Didn't have time for testing yet, but if you guys did not find anything then go for it! 🚀

PR #337, opened 2024-01-29
README tweak: decrease level of headers

cjappl:
Just to enable a new `##brew` subheader to explain a bit more about that installation.

I realized our top level header was H3 instead of H1. Remove some `#` to simplify.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@
 This tool is designed to help you use git more efficiently.
 It's **lightweight** and **easy to use**.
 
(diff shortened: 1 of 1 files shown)


carlfriedrich (approved):
Fine for me, thanks for that! Funny that GitHub seems to render both versions identically, at least it looks the same for me on both branches.

cjappl:
Agreed! They must strip off the leading # until they get to a common base! Cool.

Going to merge

Issue #324, opened 2023-08-15
Discussion: Reduce the amount of deferred execution


sandr01d:
We currently rely on deferred execution quite heavily. This leads to subtle bugs that are quite hard to catch (see e.g. #271 and #323). I also find that it makes the code hard to read in some places. Additionally deferred code has security implications, especially when using `eval` and/or when processing untrusted sources (e.g. other peoples git repositories). I think we should reduce the amount of deferred code. I'm aware that this would require quite some amount of work, but in my opinion it would be worth it.
My current idea would be to replace the deferred code with functions. I think we might also be able to reuse more code/make reusing code easier with this approach. We'd end up with two different kinds of functions:

1. Functions for the individual git commands
2. Helper functions that are used in functions of category 1.

I think it would make sense to move all functions of category 2 into a separate file.
Ideally this would also speed up the process of adding new features in the long run, as adding a new function of category 1 would mostly mean chaning up functions of category 2.

If we decide to do this I suggest we create a draft early, so anyone can comment on the general structure of things or even work on it together.

Would be interested in hearing your opinions on this.

cjappl:
I'm a fan of removing the deferred execution. These bugs are the ones that really stump me for a long time ("Do I need 4 quotes, or 5? How many backslashes?")

I'm agnostic about splitting things into separate files. Unless it becomes extremely verbose, I think a project as simple as this benefits from there being one file to find everything you may need to know. With not a lot of major IDE support for things that exist in other languages (go to definition) I could see splitting the source becoming a pain in development.

Pinging @carlfriedrich @wfxr

carlfriedrich:
@sandr01d Thanks for starting this topic. I actually agree with the issues you mention and had similar thoughts multiple times when searching for certain bugs. I was always scared to change more than needed, though, and funnily the reason is that the code covers so many corner cases in certain areas that I was afraid to lose any of them when doing too much refactoring. So that's a vicious circle, obviously. The code is hard to read and to maintain, but in order to not destroy any working behavior, we're adding even more hard-to-read and hard-to-maintain stuff. 🙈

So approaching this as a group effort is something I am all-in on! Do I get it right that you would be willing to create a first draft?

Concerning the additional file for helper functions: I actually don't see the need of it either since the forgit code is not that huge, but I would't bother having it split up as well. I think two files should be easy to handle even without a major IDE. 😊

Removing deferred executions should also make forgit faster, so that's a bonus IMO.

sandr01d:
[...]

PR #338, opened 2024-01-29
Refactor: Parse environment variables into arrays

sandr01d:
## Check list

## Description

Forgit allows specifying git options in environment variables that are then passed along to the individual git commands. We currently treat those as strings. This PR adds a `_forgit_parse_array` function and uses it to parse all such environment variables into arrays instead.

## Rationale

This step is necessary to get rid of deferred code. Currently, the environment variables are space separated strings, which do get split by using `eval`. Parsing these variables into arrays allows us to get rid of the `eval` usage further down the line. As a positive side effect, it avoids unintended globbing and word splitting.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [X] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [x] bash
    - [X] zsh
    - [ ] fish
- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

## Note

This PR belongs to #324 and resulted from discussions in #326.

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -71,6 +71,55 @@ hash emojify &>/dev/null && _forgit_emojify='|emojify'
 # extract the first git sha occurring in the input and strip trailing newline
 _forgit_extract_sha="grep -Eo '[a-f0-9]+' | head -1 | tr -d '[:space:]'"
 
+# parse a space separated string into an array
+# arrays parsed with this function are global
+_forgit_parse_array() {
+    ${IFS+"false"} && unset old_IFS || old_IFS="$IFS"
+    # read the value of the second argument
+    # into an array that has the name of the first argument
+    IFS=" " read -r -a "$1" <<< "$2"
+    ${old_IFS+"false"} && unset IFS || IFS="$old_IFS"
+}
+
+# parse environment variables into global arrays
+# this allows using them properly quoted without word splitting or glob expansion
+_forgit_log_git_opts=()
+_forgit_parse_array _forgit_log_git_opts "$FORGIT_LOG_GIT_OPTS"
+_forgit_add_git_opts=()
+_forgit_parse_array _forgit_add_git_opts "$FORGIT_ADD_GIT_OPTS"
+_forgit_diff_git_opts=()
+_forgit_parse_array _forgit_diff_git_opts "$FORGIT_DIFF_GIT_OPTS"
+_forgit_reset_head_git_opts=()
+_forgit_parse_array _forgit_reset_head_git_opts "$FORGIT_RESET_HEAD_GIT_OPTS"
+_forgit_rebase_git_opts=()
+_forgit_parse_array _forgit_rebase_git_opts "$FORGIT_REBASE_GIT_OPTS"
+_forgit_fixup_git_opts=()
+_forgit_parse_array _forgit_fixup_git_opts "$FORGIT_FIXUP_GIT_OPTS"
+_forgit_blame_git_opts=()
+_forgit_parse_array _forgit_blame_git_opts "$FORGIT_BLAME_GIT_OPTS"
+_forgit_revert_commit_git_opts=()
+_forgit_parse_array _forgit_revert_commit_git_opts "$FORGIT_REVERT_COMMIT_GIT_OPTS"
+_forgit_branch_delete_git_opts=()
+_forgit_parse_array _forgit_branch_delete_git_opts "$FORGIT_BRANCH_DELETE_GIT_OPTS"
+_forgit_checkout_commit_git_opts=()
+_forgit_parse_array _forgit_checkout_commit_git_opts "$FORGIT_CHECKOUT_COMMIT_GIT_OPTS"
+_forgit_checkout_tag_git_opts=()
+_forgit_parse_array _forgit_checkout_tag_git_opts "$FORGIT_CHECKOUT_TAG_GIT_OPTS"
+_forgit_checkout_branch_git_opts=()
+_forgit_parse_array _forgit_checkout_branch_git_opts "$FORGIT_CHECKOUT_BRANCH_GIT_OPTS"
+_forgit_checkout_file_git_opts=()
+_forgit_parse_array _forgit_checkout_file_git_opts "$FORGIT_CHECKOUT_FILE_GIT_OPTS"
+_forgit_cherry_pick_git_opts=()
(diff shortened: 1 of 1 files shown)


sandr01d:
FYI: I removed the empty line at 120 that did not make sense anymore. Everything else is exactly as it is in #326.

carlfriedrich:
Great, that's exactly how I meant it. 👍 Looking at this change it's obvious what it does, so approved content-wise.

One thing to improve, though: I would like to see the PR description in the commit message. Furthermore it should have a "Refactor: " prefix or something similar in the headline IMO.

And just to be sure: this PR does not change any behavior, does it? It really is pure refactoring, right?

carlfriedrich:
And for completeness: I did a quick test with `FORGIT_LOG_GIT_OPTS="--oneline --decorate" bin/git-forgit log` on this branch using bash and it did what I expected, so I'm gonna check bash in the PR description.

I also added a link to the original issue #324 above and removed the "won't merge" part, because we will actually merge this, we just wait until the review and rebase in #326 is finished.

bin/git-forgit
```diff
@@ -71,6 +71,55 @@ hash emojify &>/dev/null && _forgit_emojify='|emojify'
...
+_forgit_cherry_pick_git_opts=()
+_forgit_parse_array _forgit_cherry_pick_git_opts "$FORGIT_CHERRY_PICK_GIT_OPTS"
+_forgit_clean_git_opts=()
+_forgit_parse_array _forgit_clean_git_opts "$FORGIT_CLEAN_GIT_OPTS"
+_forgit_stash_push_git_opts=()
+_forgit_parse_array _forgit_stash_push_git_opts "$FORGIT_STASH_PUSH_GIT_OPTS"
+_forgit_stash_show_git_opts=()
+_forgit_parse_array _forgit_stash_show_git_opts "$FORGIT_STASH_SHOW_GIT_OPTS"
+_forgit_checkout_branch_branch_git_opts=()
+_forgit_parse_array _forgit_checkout_branch_branch_git_opts "$FORGIT_CHECKOUT_BRANCH_BRANCH_GIT_OPTS"
```
cjappl, line 121: Is there any performance impact of something like this? Does this mean every time we call a `forgit` command, we are populating the arrays for all commands, not just reading the variable for the one we are calling?

carlfriedrich: @cjappl That is indeed a good point!
@sandr01d Can we move the parsing of each variable into the corresponding individual function?

See, splitting out individual chunks of a large PR always creates focus and brings up new valuable input! :-)

sandr01d:
@carlfriedrich 

> One thing to improve, though: I would like to see the PR description in the commit message. Furthermore it should have a "Refactor: " prefix or something similar in the headline IMO.

Done

> And just to be sure: this PR does not change any behavior, does it? It really is pure refactoring, right?

Yes, the behavior should be identical to what is was before.

> removed the "won't merge" part, because we will actually merge this, we just wait until the review and rebase in https://github.com/wfxr/forgit/pull/326 is finished.

I'm a bit confused by that. I thought we would close down this PR and merge the commit with #326 or are we talking about the same thing here?

bin/git-forgit
sandr01d: Good point, I've pushed a commit with the change. Will have to move things around a bit when I rebase #326 onto this, as the arrays will mostly move into the `_forgit_git_*` functions. There are two arrays we'll be parsing twice: `_forgit_diff_git_opts` and `_forgit_blame_git_opts` because both are used in the main function (e.g. `_forgit_blame`) and the preview function. We do not make a sacrifice on performance here though, because the preview functions will be executed from a fzf subshell and the arrays need to be reparsed therefore anyway.

sandr01d:
FYI: I will keep changes as commits around here to make the review process easier, but will squash everything from this PR together when I rebase #326.

carlfriedrich:
> Done

@sandr01d Great, thanks a lot!

> Yes, the behavior should be identical to what is was before.

Great, then this is approved from my side.

> I'm a bit confused by that. I thought we would close down this PR and merge the commit with #326 or are we talking about the same thing here?

I just re-read our thread in the other PR and noticed that I obviously got you wrong when I said "Exactly." 🙈 Sorry for that.

I imagine in the end we'll have something like this:

- PR 1 (this one) contains commit A
- PR 2 contains commits A & B
- PR 3 contains commits A, B & C
- PR 4 (the original one) contains commits A, B, C & D

Of course we could just merge PR 4 and close the others when we're finished. But we can as well merge all PRs sequentially in the end, one after another, so that effectively each PR only contains its dedicated commit afterwards. We just don't merge every PR right away, but instead wait until we have all PRs ready, so that we don't have to test each one individually.
I'd consider that much cleaner. Got the idea?

That being said, I think in this case we actually could merge right away if we're sure that it does not change the behavior. What do you think?

sandr01d:
> I imagine in the end we'll have something like this:
>
>     PR 1 (this one) contains commit A
>     PR 2 contains commits A & B
>     PR 3 contains commits A, B & C
>     PR 4 (the original one) contains commits A, B, C & D
> 
> Of course we could just merge PR 4 and close the others when we're finished. But we can as well merge all PRs sequentially in the end, one after another, so that effectively each PR only contains its dedicated commit afterwards. We just don't merge every PR right away, but instead wait until we have all PRs ready, so that we don't have to test each one individually.
> I'd consider that much cleaner. Got the idea?

Gotcha now.

> That being said, I think in this case we actually could merge right away if we're sure that it does not change the behavior. What do you think?

I think I would prefer to wait until we're done with the review process before we start merging anything. I don't see any benefit in merging it right away and #326 has seen way more testing by now, while the changes here are still pretty new. Also in case we decide to change an implementation detail in this PR further down the review process, not having this merged already will save us from additional rebasing.

sandr01d:
@cjappl I've implemented the changes you requested, do you approve aswell?

cjappl:
Yep approved! Thanks.

carlfriedrich:
> I think I would prefer to wait until we're done with the review process before we start merging anything. I don't see any benefit in merging it right away and #326 has seen way more testing by now, while the changes here are still pretty new. Also in case we decide to change an implementation detail in this PR further down the review process, not having this merged already will save us from additional rebasing.

Alright, that sounds reasonable!

carlfriedrich (approved):

sandr01d:

PR #341, opened 2024-02-01
Add new workflow to update homebrew formula on release

cjappl:
This workflow was found here:

https://github.com/dawidd6/action-homebrew-bump-formula

This should be very similar to our AUR workflow, which can be seen here:
https://github.com/wfxr/forgit/blob/ab7400ddb37c758b8247ce0884cd2283a873e258/.github/workflows/aur.yaml#L1-L20

I am writing an email to @wfxr asking him to add a BREW_TOKEN to our secrets with the appropriate permissions so this can run :). This change should not be submitted until he has confirmed the BREW_TOKEN has been added to our secrets.

### DIFF
--- a/.github/workflows/brew.yaml
+++ b/.github/workflows/brew.yaml
@@ -1,4 +1,4 @@
-name: brew
+name: Homebrew
 
 on:
   workflow_run:
@@ -8,7 +8,7 @@ on:
       - completed
 
 jobs:
-  aur:
+  brew:
     runs-on: ubuntu-latest
     steps:
       - name: Update Homebrew formula
--- a/.github/workflows/brew.yaml
+++ b/.github/workflows/brew.yaml
@@ -1,4 +1,4 @@
-name: Homebrew
+name: brew
 
 on:
   workflow_run:
@@ -15,6 +15,6 @@ jobs:
         uses: dawidd6/action-homebrew-bump-formula@v3
         with:
           # Required, custom GitHub access token with the 'public_repo' and 'workflow' scopes
-          token: ${{secrets.TOKEN}}
+          token: ${{secrets.BREW_TOKEN}}
           # Formula name, required
           formula: forgit
--- a/.github/workflows/brew.yaml
+++ b/.github/workflows/brew.yaml
@@ -1,0 +1,20 @@
+name: Homebrew
+
+on:
+  workflow_run:
+    workflows:
+      - "Release"
+    types:
+      - completed
+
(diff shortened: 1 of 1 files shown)


.github/workflows/brew.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+on:
+  workflow_run:
+    workflows:
+      - "Release"
+    types:
+      - completed
+
+jobs:
+  aur:
```
carlfriedrich, line 11: This job should be called "brew" or something similar.

.github/workflows/brew.yaml
```diff
@@ -0,0 +1,20 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Update Homebrew formula
+        uses: dawidd6/action-homebrew-bump-formula@v3
+        with:
+          # Required, custom GitHub access token with the 'public_repo' and 'workflow' scopes
+          token: ${{secrets.TOKEN}}
```
carlfriedrich, line 18: And the token should probably be called `HOMEBREW_TOKEN` to make its purpose clear.

carlfriedrich (changes requested):
@cjappl Great stuff, thanks for checking this out!

carlfriedrich (approved):

carlfriedrich:
@cjappl Thanks for the quick fixes.
Just being curious: did you generate a personal access token in your account for this purpose, so that the automatic brew commits will be done in your name?

cjappl:
> @cjappl Thanks for the quick fixes. Just being curious: did you generate a personal access token in your account for this purpose, so that the automatic brew commits will be done in your name?

No, I could, but it's probably safer from a security side to have wfxr do it (just so I don't have to send one of my tokens in plaintext to him via email or similar). He is still the one that must update any secrets or other Settings of this project, as it's under his name.

I would expect the commits likely end up under his user account in that case!

wfxr (approved):
LGTM.

@cjappl BREW_TOKEN has been set.

cjappl:
Thanks @wfxr ! Will merge sometime tomorrow.

carlfriedrich:
@cjappl Did you merge into `FixZshCompletions` on purpose?

cjappl:
Whoa. No I did not. Thank you for catching..... Let me fix that

cjappl:
Great catch @carlfriedrich . would you please review #342 and #340 to make sure they are both as you expect them to be?

PR #343, opened 2024-02-02
Refactor: Move git commands from deferred code into functions

sandr01d:
## Check list

## Description

We often used deferred code to encapsulate git commands and make them reusable. This change removes deferred code for git commands and replaces it with functions instead.
Some of these new functions are used with xargs, which needs to execute them from a subshell. To make this possible, we now expose the concerned functions as forgit commands without advertising them in the help text.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [X] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [X] zsh
    - [ ] fish
- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

## Note

This PR belongs to #326 and resulted from discussions in #324

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -71,6 +71,16 @@ hash emojify &>/dev/null && _forgit_emojify='|emojify'
 # extract the first git sha occurring in the input and strip trailing newline
 _forgit_extract_sha="grep -Eo '[a-f0-9]+' | head -1 | tr -d '[:space:]'"
 
+# parse a space separated string into an array
+# arrays parsed with this function are global
+_forgit_parse_array() {
+    ${IFS+"false"} && unset old_IFS || old_IFS="$IFS"
+    # read the value of the second argument
+    # into an array that has the name of the first argument
+    IFS=" " read -r -a "$1" <<< "$2"
+    ${old_IFS+"false"} && unset IFS || IFS="$old_IFS"
+}
+
 _forgit_pager=${FORGIT_PAGER:-$(git config core.pager || echo 'cat')}
 _forgit_show_pager=${FORGIT_SHOW_PAGER:-$(git config pager.show || echo "$_forgit_pager")}
 _forgit_diff_pager=${FORGIT_DIFF_PAGER:-$(git config pager.diff || echo "$_forgit_pager")}
@@ -102,7 +112,9 @@ _forgit_log() {
     graph=--graph
     [[ $FORGIT_LOG_GRAPH_ENABLE == false ]] && graph=
     log_format=${FORGIT_GLO_FORMAT:-$_forgit_log_format}
-    eval "git log $graph --color=always --format='$log_format' $FORGIT_LOG_GIT_OPTS $* $_forgit_emojify" |
+    _forgit_log_git_opts=()
+    _forgit_parse_array _forgit_log_git_opts "$FORGIT_LOG_GIT_OPTS"
+    eval "git log $graph --color=always --format='$log_format' ${_forgit_log_git_opts[*]} $* $_forgit_emojify" |
         FZF_DEFAULT_OPTS="$opts" fzf
     fzf_exit_code=$?
     # exit successfully on 130 (ctrl-c/esc)
@@ -146,7 +158,9 @@ _forgit_diff() {
     # prevent fzf from interpreting this substring by escaping the opening bracket.
     # The string is evaluated a few subsequent times, so we need multiple escapes.
     escaped_commits=${commits//\{/\\\\\{}
-    git_diff="git diff --color=always $FORGIT_DIFF_GIT_OPTS $escaped_commits"
+    _forgit_diff_git_opts=()
+    _forgit_parse_array _forgit_diff_git_opts "$FORGIT_DIFF_GIT_OPTS"
+    git_diff="git diff --color=always ${_forgit_diff_git_opts[*]} $escaped_commits"
     preview_cmd="cd '$repo' && $get_files | xargs -0 $git_diff -U$_forgit_preview_context -- | $_forgit_diff_pager"
     enter_cmd="cd '$repo' && $get_files | xargs -0 $git_diff -U$_forgit_fullscreen_context -- | $_forgit_diff_pager"
     opts="
@@ -157,7 +171,7 @@ _forgit_diff() {
         $FORGIT_DIFF_FZF_OPTS
         --prompt=\"$commits > \"
     "
-    eval "git diff --name-status $FORGIT_DIFF_GIT_OPTS $commits -- ${files[*]} | sed -E 's/^([[:alnum:]]+)[[:space:]]+(.*)$/[\1]	\2/'" |
+    eval "git diff --name-status ${_forgit_diff_git_opts[*]} $commits -- ${files[*]} | sed -E 's/^([[:alnum:]]+)[[:space:]]+(.*)$/[\1]	\2/'" |
         sed 's/	/  ->  /2' | expand -t 8 |
         FZF_DEFAULT_OPTS="$opts" fzf
     fzf_exit_code=$?
@@ -170,7 +184,9 @@ _forgit_diff() {
 _forgit_add() {
     _forgit_inside_work_tree || return 1
     local git_add changed unmerged untracked files opts preview extract
-    git_add="git add $FORGIT_ADD_GIT_OPTS"
+    _forgit_add_git_opts=()
+    _forgit_parse_array _forgit_add_git_opts "$FORGIT_ADD_GIT_OPTS"
+    git_add="git add ${_forgit_add_git_opts[*]}"
     # Add files if passed as arguments
     [[ $# -ne 0 ]] && { $git_add "$@" && git status -su; return $?; }
 
@@ -210,7 +226,9 @@ _forgit_add() {
 _forgit_reset_head() {
     _forgit_inside_work_tree || return 1
     local git_reset_head cmd files opts rootdir
-    git_reset_head="git reset -q $FORGIT_RESET_HEAD_GIT_OPTS HEAD"
+    _forgit_reset_head_git_opts=()
+    _forgit_parse_array _forgit_reset_head_git_opts "$FORGIT_RESET_HEAD_GIT_OPTS"
+    git_reset_head="git reset -q ${_forgit_reset_head_git_opts[*]} HEAD"
     [[ $# -ne 0 ]] && { $git_reset_head "$@" && git status --short; return $?; }
     rootdir=$(git rev-parse --show-toplevel)
     cmd="git diff --staged --color=always -- $rootdir/{} | $_forgit_diff_pager "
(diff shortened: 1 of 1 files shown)


sandr01d:
@carlfriedrich I did not include the deferred git commands that we did not end up creating functions for in this PR. Let me know whether you would like to have them included here.

A note regarding 2d81ed9: This was not originally included in #326. When I was revisiting this I came to the conclusion that a `_forgit_git_branch_delete` function should always pass the -D flag to `git branch`.

carlfriedrich (changes requested):
@sandr01d Thanks a lot for submitting this new patch!

Haven't looked into everything in detail, yet, but first thing I noticed is that 7657444632c8d0421d63838f318e03fa4dc35168 reverts some of the changes of #338. The array parsing in the functions gets lost. Can you fix that?

sandr01d:
> Haven't looked into everything in detail, yet, but first thing I noticed is that [7657444](https://github.com/wfxr/forgit/commit/7657444632c8d0421d63838f318e03fa4dc35168) reverts some of the changes of #338. The array parsing in the functions gets lost. Can you fix that?

The array parsing was moved into the `_forgit_git_*` functions were the arrays are being used. I think it is correct as is, can you verify?

carlfriedrich:
@sandr01d Yes, they are correctly added in the first commit of this PR (1e8f79edefe17b35bad55725973ed387e910ce73), but the second commit (7657444632c8d0421d63838f318e03fa4dc35168) removes the calls from the `_forgit_git_*` functions again.

sandr01d:
> @sandr01d Yes, they are correctly added in the first commit of this PR ([1e8f79e](https://github.com/wfxr/forgit/commit/1e8f79edefe17b35bad55725973ed387e910ce73)), but the second commit ([7657444](https://github.com/wfxr/forgit/commit/7657444632c8d0421d63838f318e03fa4dc35168)) removes the calls from the `_forgit_git_*` functions again.

Could you please point me to a line were this is happening, I do not understand. The parsing is added to the `_forgit_*` functions in https://github.com/wfxr/forgit/commit/1e8f79edefe17b35bad55725973ed387e910ce73 and moved to the `_forgit_git_*` functions in https://github.com/wfxr/forgit/commit/7657444632c8d0421d63838f318e03fa4dc35168. See for example, `_git_forgit_add` after https://github.com/wfxr/forgit/commit/7657444632c8d0421d63838f318e03fa4dc35168:

https://github.com/sandr01d/forgit/blob/2d81ed909fd799566fb7c377f09e6b23a2675173/bin/git-forgit#L183C1-L187C2

carlfriedrich:
@sandr01d Oh, I'm sorry! I viewed the patch on my tiny mobile phone screen, which obviously is not a good idea, and also I obviously did not read your comment correctly. 🙈 So yes, on my desktop monitor I see now that the calls are moved and not removed. Will have a more detailed look soon.

sandr01d:
> @sandr01d Oh, I'm sorry! I viewed the patch on my tiny mobile phone screen, which obviously is not a good idea, and also I obviously did not read your comment correctly. 🙈 So yes, on my desktop monitor I see now that the calls are moved and not removed. Will have a more detailed look soon.

No worries and thanks for the clarification. I was only questioning my sanity for a brief second :laughing:

carlfriedrich:
Just browsed through all the changes and the patch looks clean, great work! Also good catch with the `-D` flag in `_forgit_git_branch_delete`. I would suggest to squash that commit into its parent.

I wonder if we could replace the four `xargs` calls with `for` loops, so that we do not have to export these git functions as private functions to the forgit CLI. That would make the code more readable IMO, because we would remove a level of indirection there and call the functions directly. WDYT?

sandr01d:
> I would suggest to squash that commit into its parent.

Done

> I wonder if we could replace the four `xargs` calls with `for` loops, so that we do not have to export these git functions as private functions to the forgit CLI. That would make the code more readable IMO, because we would remove a level of indirection there and call the functions directly. WDYT?

I think that is a great idea, but I think we should probably do so in a follow up PR once #326 has been merged. I see this as more of a general refactor and not related to deferred code and #326 is already pretty big. WDYT?

carlfriedrich:
I understand that this can be seen as a more general refactoring. I think, though, that it would make this specific commit better, so IMO it would be the right thing to do here.
Concerning the complexity of this PR, it would add the `xargs` refactoring, but instead remove the introduction of `private_commands` here, so I think this specific PR wouldn't grow in complexity.
And concerning the complexity of #326: everything we extract to these chunk-PRs will reduce the overall complexity of the final PR (what we're doing here doesn't need to be done in #326).
I see, however, that everything we're adding here which wasn't part of #326 before, increases the difficulty of rebasing #326 on this PR. I would consider this a good invest, though, and I again want to offer my help with this. I know, rebasing can be a pain, but I still enjoy doing it when I know that it's worth for a good end result.
So are you in? 🤗

sandr01d:
> I understand that this can be seen as a more general refactoring. I think, though, that it would make this specific commit better, so IMO it would be the right thing to do here.

That is totally reasonable and I agree with that. I am not against the change, but I don't really see a good way to implement it in this specific PR, let me elaborate a bit on this.
I've refactored `_forgit_reset_head`, `_forgit_checkout_commit` and `_forgit_branch_delete` and they work fine. You can take a look at it [here](https://github.com/sandr01d/forgit/commit/b4feaba86b7cd0bb3f2cb23cc3b2b67c9ad8a7e9) in case you're interested.
The issue is with `_forgit_stash_show`. The usage of `xargs` is withing deferred code that is executed from an fzf subshell:

```bash
# xargs is within this deferred code
cmd="echo {} |cut -d: -f1 |xargs -I% $FORGIT git_stash_show % |$_forgit_diff_pager"
    opts="
        $FORGIT_FZF_DEFAULT_OPTS
        +s +m -0 --tiebreak=index --bind=\"enter:execute($cmd | $_forgit_enter_pager)\"
        --bind=\"ctrl-y:execute-silent(echo {} | cut -d: -f1 | tr -d '[:space:]' | ${FORGIT_COPY_CMD:-pbcopy})\"
# we use this for the fzf preview here
        --preview=\"$cmd\"
        $FORGIT_STASH_FZF_OPTS
    "
```
We can not call `_forgit_git_stash_show` from the subshell. We've introduced the `private_commands` to deal with this particular issue for the `_forgit_*_preview` functions. The code I've pasted above is later on replaced with a preview function and then it becomes trivial to get rid of `xargs`: 

```diff
_forgit_stash_show_preview() {
-   echo "$1" | cut -d: -f1 | xargs -I% "$FORGIT" "git_stash_show" % | $_forgit_diff_pager
+   # we don't actually need a loop here because
+   # there is always just one stash previewed
+   local stash
+   stash=$(echo "$1" | cut -d: -f1)
+   _forgit_git_stash_show "$stash" | $_forgit_diff_pager
}
```

If we want to get rid of `xargs` and the `private_commands` in this PR that means we'd have to find a new solution to this. Which to me clearly indicates that we should isolate other chunks first and once the `_forgit_*_preview` functions are introduced, we could rebase this PR on top of that.

> I see, however, that everything we're adding here which wasn't part of https://github.com/wfxr/forgit/pull/326 before, increases the difficulty of rebasing https://github.com/wfxr/forgit/pull/326 on this PR. I would consider this a good invest, though, and I again want to offer my help with this.

Appreciated, thanks. I think it would be great if we could take turns rebasing every now and then. :+1:

carlfriedrich:
@sandr01d Ah, I see. Thanks for the detailed explanation. I actually didn't realize that it's used in deferred code at one place. In this case I agree, let's do this after #326. I'd consider this PR approved then. ✅

Shall I rebase #326 on this PR then?

carlfriedrich (approved):

sandr01d:
Alternatively, to doing this after #326, we could keep this here as is for the moment and continue splitting out other chunks. Once the preview functions are introduced, we can rebase this PR on top of that and get rid of `xargs` here. I would convert this PR in a draft in this case. I think that would result in the cleanest commit history, so what do you think about that?

carlfriedrich:
That sounds like a good idea! Then let's do the preview functions in the next chunk. I just took a quick look over them and think they would make a reasonable patch. WDYT?

sandr01d:
I've now rebased this PR on #349 and replaced the usage of `xargs`, as previously requested by @carlfriedrich. Re-requesting you reviews @cjappl @carlfriedrich.

bin/git-forgit
```diff
@@ -284,25 +287,36 @@ _forgit_reset_head_preview() {
...
+        files+=("$file")
+    done < <(git diff --staged --name-only | FZF_DEFAULT_OPTS="$opts" fzf)
+    if [[ ${#files} -eq 0 ]]; then
+        echo 'Nothing to unstage.'
+        return 1
+    fi
+    for file in "${files[@]}"; do
+        _forgit_git_reset_head "$rootdir/$file"
+    done
+    git status --short
```
carlfriedrich, line 319: This is so much better to read than the previous statement! 👍

However, is `IFS=''` correct here? Haven't tested it, yet, but from reading the code, I think it should be `IFS='\n'`, shouldn't it?

bin/git-forgit
```diff
@@ -677,9 +709,15 @@ _forgit_branch_delete() {
     "
 
     cmd="git branch --color=always | LC_ALL=C sort -k1.1,1.1 -rs"
```
carlfriedrich, line 711: What about this deferred `git` call here? Can we get rid of that as well?

carlfriedrich (changes requested):
Looks mostly good to me, thanks for the work. Can you check the two comments I left?

wfxr (approved):
LGTM

cjappl (approved):
Did a functional check with `fish` on MacOS and seems like everything is still working as intended :)

bin/git-forgit
sandr01d: `IFS=''` should actually be correct. The `read` command already splits the output by line and we do not want/need to do any further splitting. It's suggested to do it this way in [SC2207](https://www.shellcheck.net/wiki/SC2207).

bin/git-forgit
sandr01d: I've only included deferred `git` calls that we replaced with separate functions in #326, see my [previous comment](https://github.com/wfxr/forgit/pull/343#issuecomment-1951397215). We could replace all deferred `git` calls in this PR, which would make it a bit bigger, as there are still a few of them left that we did not create functions for. We could also do that in the next PR. I'm open for both.

bin/git-forgit
carlfriedrich: Great, thanks for the explaination.

bin/git-forgit
carlfriedrich: Ah sorry, I missed that part of the comment in the first place. I would prefer replacing all `git` calls in this PR, since that's what its title says. :-)

sandr01d: f79da92 removes all deferred git commands. I had to include converting `$_forgit_emojify` into a function and the replacement of the `$graph` variable with an array to get the commands working without deferred code and `eval`. This makes this PR quite big. Haven't squashed the commit yet, because I'm thinking creating a separate PR out of f79da92 to reduce the size of this one. WDYT?

carlfriedrich: Ah, I see, that's quite a huge additional change indeed. I did not expect that, sorry. So yes, let's move that to a new PR. I would consider this PR finished then.

Do I get it right that none of the new changes were part of the original #326? I wonder what's the best way to continue then, since we have some changes in #326 left to include. Shall we do those first?

carlfriedrich (approved):

bin/git-forgit
sandr01d: > Do I get it right that none of the new changes were part of the original https://github.com/wfxr/forgit/pull/326?

No they were mostly part of #326, I just moved them over to this branch. The only thing in this PR that is not in #326 yet are the four hunks that remove `xargs`. 
I would suggest we rebase #326 on this PR to make sure we don't digress too far and I create a new PR that converts `$_forgit_emojify` into a function and after that a PR that replaces the rest of the deferred `git` commands. Sounds good?

carlfriedrich: Yep, that sounds good, thanks a lot!

sandr01d: FYI: I started to rebase #326. I'm almost done and will finalize it tomorrow, so don't put any effort into it.

carlfriedrich: Oh wow, thanks a lot for your effort!

sandr01d:

PR #344, opened 2024-02-05
feat(completions): add completion for fish

folliehiyuki:
fish's builtin git completion automatically registers `git-forgit` completions as completions for `forgit` subcommand of git. Therefore this PR provides completions for both formats `git-forgit` and `git forgit`.

Simple subcommands get the completion list from `__fish_git_*` functions, while others requiring more than 1 `__fish_git_*` completion sources reuse the completion items from the corresponding git commands.

## Check list

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [ ] zsh
    - [x] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/completions/git-forgit.fish
+++ b/completions/git-forgit.fish
@@ -1,0 +1,59 @@
+#
+# forgit completions for fish plugin
+#
+# Place this file inside your <fish_config_dir>/completions/ directory.
+# It's usually located at ~/.config/fish/completions/. The file is lazily
+# sourced when git-forgit command or forgit subcommand of git is invoked.
+
+function __fish_forgit_needs_subcommand
+    for subcmd in add blame branch_delete checkout_branch checkout_commit checkout_file checkout_tag \
+        cherry_pick cherry_pick_from_branch clean diff fixup ignore log rebase reset_head revert_commit \
+        stash_show stash_push
+        if contains -- $subcmd (commandline -opc)
+            return 1
+        end
+    end
+    return 0
+end
+
+# Load helper functions in git completion file
+not functions -q __fish_git && source $__fish_data_dir/completions/git.fish
+
+# No file completion by default
+complete -c git-forgit -x
+
+complete -c git-forgit -n __fish_forgit_needs_subcommand -a add -d 'git add selector'
+complete -c git-forgit -n __fish_forgit_needs_subcommand -a blame -d 'git blame viewer'
+complete -c git-forgit -n __fish_forgit_needs_subcommand -a branch_delete -d 'git branch deletion selector'
+complete -c git-forgit -n __fish_forgit_needs_subcommand -a checkout_branch -d 'git checkout branch selector'
+complete -c git-forgit -n __fish_forgit_needs_subcommand -a checkout_commit -d 'git checkout commit selector'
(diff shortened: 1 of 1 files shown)


sandr01d:
I ran into the same issue with the CI previously, but the issue went away by itself before I could do anything to fix it. For some reason the fish PPA fails to install

```
Cannot add PPA: 'ppa:~fish-shell/ubuntu/release-3'.
ERROR: '~fish-shell' user or team does not exist.
Error: Process completed with exit code 1.
```

I'm suspecting an issue with the fish PPA. This is not  caused by the code changes, the CI fails before the actual checks.

folliehiyuki:
> Give me a few test cases I can run locally on my machine with explicit detail to ensure the feature is working?

Just use `git-forgit` and `git forgit` as you normally do + spamming the <TAB> key!

```bash
mkdir test-repo && cd test-repo
git init
touch file{1,2,3}

# See that the completion list includes all subcommands
git forgit <Tab>
git-forgit <Tab>

# Completion for `add`
git-forgit add <Tab><Tab>
git forgit add <Tab><Tab>
...
```

All the `__fish_git_*` functions are loaded from the completion file of git, so you need it in-place first. It comes by default from every fish-shell package in Linux distros.

cjappl:
@folliehiyuki this is great! Working as you intended.

A couple questions:
1. Do you have any thoughts on if this should live in it's own file, or these methods be run at the end of forgit.plugin.fish? I don't know the overhead or cost, but I think it would be nice for the fish completions to "just work" when someone sources forgit.
2. I'm noticing these completions are now much nicer than the ones on the normal functions/aliases (`glo <TAB` is very sad in comparision ) if you're feeling up to it, after this gets submitted one of us should go clean up those completions as well! This is a great upgrade.

Let me know what you think on point 1. Depending on that we can submit.

folliehiyuki:
On point 1, see https://fishshell.com/docs/current/completions.html#where-to-put-completions.
* Fish plugin managers like `oh-my-fish` and `fisher` are smart enough to put the completion into the right place in your `$__fish_config_dir/completions` on installation (see https://github.com/jorgebucaran/replay.fish for an example of the repo structure the plugin managers understand, which we already follow nicely).
* Furthermore, fish completion files are lazy-loaded (which means they're sourced the 1st time you press `<TAB> on the corresponding command).

So, it's better putting the completion in a separated file. It works with just the `git-forgit` command inside `$PATH` (I actually use `git-forgit` this way without the aliases). There is no need to source the fish plugin.

On point 2, as said above, I don't use the fish plugin personally, so it's hard to judge. My recommendation would be to use [abbr](https://fishshell.com/docs/current/cmds/abbr.html) instead of aliases, so on the command line, pressing `glo <TAB>` would actually result to `git forgit log <TAB>` and we'll have nice completion (still with fewer key presses).

cjappl:
Awesome. Makes sense to me. Thanks again for the contribution @folliehiyuki ! merging it now.

If you feel you have energy to tune up the completions in the non-git-plugin version, another contribution would be greatly appreciated, otherwise I'll check it out when I have time.

Cheers!

PR #310, opened 2023-05-26
Added completions for zsh

sandr01d:
## Check list

- [X] I have performed a self-review of my code
- [X] I have commented my code in hard-to-understand areas
- [X] I have made corresponding changes to the documentation

## Description

The implementation follows the existing bash completions with a few differences:

I've split the implementation into two files: _completions/\_git-forgit_ for usage of forgit as a git subcommand and _completions/git-forgit.zsh_ for usage of forgit as a shell plugin. The reason for this is that zsh's `autoload` does not load the files on shell start, but dynamically the first time a completion is needed. Having the `compdef` commands that bind the completions for the shell plugin in the main file would lead to the completions becoming available after `git forgit` has been completed for the first time. Because this behavior would be inconsistent and confusing, I've placed them in a separate file that has to be sourced explicitly. _completions/\_git-forgit_ simply has to be placed in a directory that is in zsh's `$fpath`.
I've changed the behavior of the `_git-forgit` completion function in comparison to the one for bash to allow more than one completion per command. This is useful for e.g. `forgit::add`.

[asciicast](https://asciinema.org/a/obvYlQLiF9scW5RZ2BjOwBrK6)

I've also added/updated a few of the descriptions that are used in the fish wrapper, so they are the same across shells.

Closes #291 

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Breaking change
- [X] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [X] zsh
    - [ ] fish
- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

PR #346, opened 2024-02-06
Auto enable completion

wfxr:
## Check list

## Description

We can enable completions automatically for `zsh` ~~and `bash`~~.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/completions/git-forgit.bash
+++ b/completions/git-forgit.bash
@@ -6,9 +6,6 @@
 #
 #   /usr/share/bash-completion/completions
 #   ~/.local/share/bash-completion/completions
-#
-# When using forgit via the shell plugin, source this file explicitly after
-# forgit.plugin.zsh to enable tab completion for shell functions and aliases.
 
 _git_branch_delete()
 {
--- a/completions/git-forgit.bash
+++ b/completions/git-forgit.bash
@@ -6,6 +6,9 @@
 #
 #   /usr/share/bash-completion/completions
 #   ~/.local/share/bash-completion/completions
+#
+# When using forgit via the shell plugin, source this file explicitly after
+# forgit.plugin.zsh to enable tab completion for shell functions and aliases.
 
 _git_branch_delete()
 {
--- a/completions/git-forgit.zsh
+++ b/completions/git-forgit.zsh
@@ -1,10 +1,5 @@
 #!/bin/zsh
-#
 # forgit completions for zsh plugin
-#
-# When using forgit via the shell plugin, place completions/_git-forgit in your
-# $fpath (e.g. /usr/share/zsh/site-functions) and source this file after
-# forgit.plugin.zsh to enable tab completion for shell functions and aliases.
 
 # Check if forgit plugin is loaded
 if (( $+functions[forgit::add] )); then
--- a/forgit.plugin.zsh
+++ b/forgit.plugin.zsh
@@ -11,6 +11,7 @@ if [[ -n "$ZSH_VERSION" ]]; then
     # shellcheck disable=2277,2296,2298
     0="${${(M)0:#/*}:-$PWD/$0}"
     INSTALL_DIR="${0:h}"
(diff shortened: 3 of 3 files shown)


wfxr:
This pr works for me, but fails in ci because of the lack of `_git`. And considering the reason @sandr01d explained in #310:

> The reason for this is that zsh's autoload does not load the files on shell start, but dynamically the first time a completion is needed. Having the compdef commands that bind the completions for the shell plugin in the main file would lead to the completions becoming available after git forgit has been completed for the first time. Because this behavior would be inconsistent and confusing, I've placed them in a separate file that has to be sourced explicitly.

I think it's better to sourcing the `completions/git-forgit.zsh` manually.

carlfriedrich:
@wfxr I actually like this idea and it makes perfect sense to me. @sandr01d @cjappl What do you think?

In order to fix the shellcheck IMO we would have to install the `git` completions before the tests, or use a container which already has them. I think this shouldn't be too difficult, or am I wrong?

wfxr:
> @wfxr I actually like this idea and it makes perfect sense to me. @sandr01d @cjappl What do you think?
> 
> In order to fix the shellcheck IMO we would have to install the `git` completions before the tests, or use a container which already has them. I think this shouldn't be too difficult, or am I wrong?

@carlfriedrich Thanks for the quickly testing! I'm not sure how to avoid the `_git` not found error gracefully if someone actually doesn't have git completions installed. @sandr01d Is it possible to delay this error until completion is triggered instead of when `completions/git-forgit.zsh` is loaded?

carlfriedrich:
Now that I look at it again I think this should actually be fixed in the ZSH completion file. We [rely on the git completions being installed](https://github.com/wfxr/forgit/blob/master/completions/_git-forgit#L97), while actually we should have error handling for that in case they are not. @sandr01d Can you take a look if I see that correctly? I am gonna reopen this for discussion.

sandr01d:
Only saw this PR today, sorry for the late reply. I think that the issue this PR is trying to solve has already been solved by #340, where @cjappl found a way to auto enable zsh completions whithout having to source any additional files (be it manually or automatically).

> Now that I look at it again I think this should actually be fixed in the ZSH completion file. We rely on the git completions being installed https://github.com/wfxr/forgit/blob/master/completions/_git-forgit#L97, while actually we should have error handling for that in case they are not.

We currently do rely on the git completions being present and simply fail with an exit code of 1 when a user tries to use a completion and they aren't. I guess in this case we should simply print out a user facing error message letting the user now whats going on. [`_message`](https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions-howto.org#other-functions) appears to be the correct utility function to do so. @carlfriedrich Is that what you have in mind?

cjappl:
Yeah, I think we should re-test with the newly merged file and see if there is any difference in what we intend to be the user experience vs how we have it now. 

I think the latest version is pretty streamlined and also follows the "best practices" i saw in many other zsh completions. The main thing I'd worry about with this PR is actively loading/sourcing the completions on shell startup instead of lazy loading when the completion is requested.

TLDR: check out main with the new completion file in your fpath and see if that leaves anything unaccounted for :)

carlfriedrich:
> I guess in this case we should simply print out a user facing error message letting the user now whats going on. [`_message`](https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions-howto.org#other-functions) appears to be the correct utility function to do so. @carlfriedrich Is that what you have in mind?

That sounds reasonable.

However, if #340 already provided a better implementation for this kind of problem, I'm gonna close this again. Thanks for your comments!

PR #348, opened 2024-02-06
feat(fish): use abbr instead of alias

folliehiyuki:
## Check list

## Description

* Switch from aliases to abbreviations, as I suggested in https://github.com/wfxr/forgit/pull/344#issuecomment-1930072329
* `string collect` returns 1 on empty argument, so it's used here to avoid repeating `if else end` blocks. Another benefit is that the output of `string collect` is ensured to be a single string. From `string --help`:
> `string collect` collects its input into a single output argument, without splitting the output when used in a command substitution. This is useful when trying to collect multiline output from another
command into a variable. Exit status: 0 if any output argument is non-empty, or 1 otherwise.

This PR is marked as breaking change, since abbr behaves differently from alias. It can only be used in the interactive command line (so putting `exec glo` into your scripts won't work).

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [x] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [ ] zsh
    - [x] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/conf.d/forgit.plugin.fish
+++ b/conf.d/forgit.plugin.fish
@@ -27,206 +27,27 @@ set | awk -F ' ' '{ print $1 }' | grep FORGIT_ | while read var
     end
 end
 
-function forgit::log -d "git commit viewer"
-    "$FORGIT" log $argv
-end
+# alias `git-forgit` to the full-path of the command
+alias git-forgit "$FORGIT_INSTALL_DIR/bin/git-forgit"
 
-function forgit::diff -d "git diff viewer" --argument-names arg1 arg2
-    "$FORGIT" diff $argv
-end
-
-function forgit::add -d "git add selector" --wraps "git add"
-    "$FORGIT" add $argv
-end
-
-function forgit::reset::head -d "git reset HEAD (unstage) selector"
-    "$FORGIT" reset_head $argv
-end
-
-function forgit::stash::show -d "git stash viewer"
-    "$FORGIT" stash_show $argv
-end
-
-function forgit::stash::push -d "git stash push selector" ()
-    "$FORGIT" stash_push $argv
-end
-
-function forgit::clean -d "git clean selector"
-    "$FORGIT" clean $argv
-end
-
-function forgit::cherry::pick -d "git cherry-picking" --argument-names 'target' --wraps "git cherry-pick"
-    "$FORGIT" cherry_pick $argv
-end
-
-function forgit::cherry::pick::from::branch -d "git cherry-picking with interactive branch selection" --wraps "git cherry-pick"
-    "$FORGIT" cherry_pick_from_branch $argv
-end
-
-function forgit::rebase -d "git rebase"
-    "$FORGIT" rebase $argv
-end
-
-function forgit::fixup -d "git fixup"
-    "$FORGIT" fixup $argv
-end
-
-function forgit::checkout::file -d "git checkout-file selector" --argument-names 'file_name' --wraps "git checkout --"
-    "$FORGIT" checkout_file $argv
-end
-
-function forgit::checkout::branch -d "git checkout branch selector" --argument-names 'input_branch_name' --wraps "git branch"
-    "$FORGIT" checkout_branch $argv
-end
-
-function forgit::checkout::tag -d "git checkout tag selector" --argument-names 'tag_name' --wraps "git checkout"
-    "$FORGIT" checkout_tag $argv
-end
-
-function forgit::checkout::commit -d "git checkout commit selector" --argument-names 'commit_id' --wraps "git checkout"
-    "$FORGIT" checkout_commit $argv
-end
-
-function forgit::branch::delete -d "git branch deletion selector" --wraps "git branch --delete"
-    "$FORGIT" branch_delete $argv
-end
-
-function forgit::revert::commit -d "git revert commit selector" --argument-names 'commit_hash' --wraps "git revert --"
-    "$FORGIT" revert_commit $argv
-end
-
-function forgit::blame -d "git blame viewer"
-    "$FORGIT" blame $argv
-end
-
-function forgit::ignore -d "git ignore generator"
-    "$FORGIT" ignore $argv
-end
-
-function forgit::ignore::update
-    "$FORGIT" ignore_update $argv
-end
-
-function forgit::ignore::get
-    "$FORGIT" ignore_get $argv
-end
-
-function forgit::ignore::list
-    "$FORGIT" ignore_list $argv
-end
-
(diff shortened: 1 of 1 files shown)


conf.d/forgit.plugin.fish
```diff
@@ -27,206 +27,27 @@ set | awk -F ' ' '{ print $1 }' | grep FORGIT_ | while read var
...
-
-function forgit::ignore::update
-    "$FORGIT" ignore_update $argv
-end
-
-function forgit::ignore::get
-    "$FORGIT" ignore_get $argv
-end
+# Add `git-forgit` command to PATH
+fish_add_path -ga "$FORGIT_INSTALL_DIR/bin"
```
cjappl, line 31: Is there any way to do this without affecting the user's $PATH? That seems like it could be a little invasive?

conf.d/forgit.plugin.fish
```diff
@@ -3,7 +3,7 @@
 set INSTALL_DIR (dirname (dirname (status -f)))
 set -x FORGIT_INSTALL_DIR "$INSTALL_DIR/conf.d"
 set -x FORGIT "$FORGIT_INSTALL_DIR/bin/git-forgit"
-if [ ! -e "$FORGIT" ]
+if not test -e "$FORGIT"
```
cjappl, line 6: Can you break this into a separate review, it's a great change but I want to be as focused as possible per review.

conf.d/forgit.plugin.fish
```diff
@@ -27,206 +27,27 @@ set | awk -F ' ' '{ print $1 }' | grep FORGIT_ | while read var
...
-    if test -n "$forgit_checkout_tag"
-        alias $forgit_checkout_tag 'forgit::checkout::tag'
-    else
-        alias gct 'forgit::checkout::tag'
-    end
-
+    abbr -a -- (string collect $forgit_add; or string collect "ga") git-forgit add
+    abbr -a -- (string collect $forgit_reset_head; or string collect "grh") git-forgit reset_head
+    abbr -a -- (string collect $forgit_log; or string collect "glo") git-forgit log
+    abbr -a -- (string collect $forgit_diff; or string collect "gd") git-forgit diff
```
cjappl, line 38: re: your comment about this being a breaking change, I would guess someone out there relies on this command existing in a form that can be run on a non-interactive shell.

Is there a workaround they can use? I guess just call `exec git forgit log` instead of the now broken `exec glo` ?

cjappl (changes requested):
Overall this is a MASSIVE improvement. I have not tested it yet (but will do so after some chatting).

I especially like the extreme reduction in duplication, having "one source of truth" for commands and completions etc. 

I'm worried mainly about affecting the user's $PATH in a way that is invasive. I would be surprised if I sourced a plugin and my PATH was affected.

conf.d/forgit.plugin.fish
folliehiyuki: I can replace `git-forgit` in the abbreviations with `$FORGIT`. Completion shouldn't be affected, but the expanded command line might look a bit ugly (fish expands the full path of `$FORGIT_INSTALL_DIR/bin/git-forgit`.

conf.d/forgit.plugin.fish
folliehiyuki: Yeah. That's the way to go. Usage inside automated scripts should rely on the full commands.

conf.d/forgit.plugin.fish
cjappl: I agree it's small, but considering there is a potentially breaking change in the rest of this, having this split apart makes sure if we need to revert anything we don't lose this improvement! 

I would definitely prefer a separate review for this one (even if it's added overhead) I'll approve it ASAP. :)

conf.d/forgit.plugin.fish
cjappl: Cool, that seems fine to me!

conf.d/forgit.plugin.fish
cjappl: Meaning when you expand the abbreviation? 

It would expand:
```
ga
```
becomes
```
$FORGIT_INSTALL_DIR/bin/git-forgit add
```

For the user? 

Instead of 

```
git forgit add
```

conf.d/forgit.plugin.fish
folliehiyuki: Since I find a few improvements to add, let me put it into a separated PR.

conf.d/forgit.plugin.fish
folliehiyuki: I added `git-forgit` alias to the full-path command, so hopefully it'll solve the "lengthy command line" problem.

cjappl: Great! Let me try this out for a bit and get a feel for it before merging. Overall code structure wise it's a big improvement, just going to perform some testing to make sure I understand it and there aren't any gotchas.

cjappl: Still testing this! It's definitely a change as a user, but easy to adapt to. It also de-duplicates the code so much, which is a huge positive.

conf.d/forgit.plugin.fish
```diff
@@ -2,10 +2,8 @@
 
 set INSTALL_DIR (dirname (dirname (status -f)))
 set -x FORGIT_INSTALL_DIR "$INSTALL_DIR/conf.d"
-set -x FORGIT "$FORGIT_INSTALL_DIR/bin/git-forgit"
```
cjappl, line 5: Unfortunately this line cannot change, this is auto-replaced in our homebrew script, so it should stay as it is! I would revert this entire section (5-8)

carlfriedrich: @cjappl The line has just been moved (to line 8), why can't it be auto-replaced there now?

cjappl: Perhaps it could but it would need testing with the homebrew formula to make sure it still works. If this change can be done with keeping the original structure in place it's one less thing to check.

folliehiyuki: I removed the 2nd commit modifying the variable.

cjappl (approved):
I think let's run with this for now. I think the code maintenance is substantially improved and the minor backwards compatability stuff (not being able to use abbreviations in scripts for fish) is easily fixable by clients.

I'm going to merge and we can always revert if there is a riot.

Thanks again for the fix on this, I had no idea the power of abbr !

PR #349, opened 2024-02-06
Refactor: Replace deferred code used for fzf preview with functions

sandr01d:
## Check list

## Description

Removes the deferred code that is used for creating the fzf preview functions and replaces it with `_forgit_*_preview` functions instead.
This PR changes the `flags` variable in `_forgit_blame` from a string to an array. This is necessary to allow the flags to be passed to `_forgit_blame_preview` as individual arguments.
The only implementation detail I changed in comparison to #326 is that I removed `xargs` from from `_forgit_stash_show_preview`, as discussed in #343. `xargs` was actually not necessary here because we always preview only a single stash. This way we do not have to expose `_forgit_git_stash_show` as a `private_command`.
I've included new functions we've added in #326, that are used inside the `_forgit_*_preview` functions. I regard them as part of the preview functions.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [X] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [X] zsh
    - [ ] fish
- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

## Note

This PR belongs to https://github.com/wfxr/forgit/pull/326 and resulted from discussions in https://github.com/wfxr/forgit/issues/324

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -71,6 +71,16 @@ hash emojify &>/dev/null && _forgit_emojify='|emojify'
 # extract the first git sha occurring in the input and strip trailing newline
 _forgit_extract_sha="grep -Eo '[a-f0-9]+' | head -1 | tr -d '[:space:]'"
 
+# parse a space separated string into an array
+# arrays parsed with this function are global
+_forgit_parse_array() {
+    ${IFS+"false"} && unset old_IFS || old_IFS="$IFS"
+    # read the value of the second argument
+    # into an array that has the name of the first argument
+    IFS=" " read -r -a "$1" <<< "$2"
+    ${old_IFS+"false"} && unset IFS || IFS="$old_IFS"
+}
+
 _forgit_pager=${FORGIT_PAGER:-$(git config core.pager || echo 'cat')}
 _forgit_show_pager=${FORGIT_SHOW_PAGER:-$(git config pager.show || echo "$_forgit_pager")}
 _forgit_diff_pager=${FORGIT_DIFF_PAGER:-$(git config pager.diff || echo "$_forgit_pager")}
@@ -102,7 +112,9 @@ _forgit_log() {
     graph=--graph
     [[ $FORGIT_LOG_GRAPH_ENABLE == false ]] && graph=
     log_format=${FORGIT_GLO_FORMAT:-$_forgit_log_format}
-    eval "git log $graph --color=always --format='$log_format' $FORGIT_LOG_GIT_OPTS $* $_forgit_emojify" |
+    _forgit_log_git_opts=()
+    _forgit_parse_array _forgit_log_git_opts "$FORGIT_LOG_GIT_OPTS"
+    eval "git log $graph --color=always --format='$log_format' ${_forgit_log_git_opts[*]} $* $_forgit_emojify" |
         FZF_DEFAULT_OPTS="$opts" fzf
     fzf_exit_code=$?
     # exit successfully on 130 (ctrl-c/esc)
@@ -146,7 +158,9 @@ _forgit_diff() {
     # prevent fzf from interpreting this substring by escaping the opening bracket.
     # The string is evaluated a few subsequent times, so we need multiple escapes.
     escaped_commits=${commits//\{/\\\\\{}
-    git_diff="git diff --color=always $FORGIT_DIFF_GIT_OPTS $escaped_commits"
+    _forgit_diff_git_opts=()
+    _forgit_parse_array _forgit_diff_git_opts "$FORGIT_DIFF_GIT_OPTS"
+    git_diff="git diff --color=always ${_forgit_diff_git_opts[*]} $escaped_commits"
     preview_cmd="cd '$repo' && $get_files | xargs -0 $git_diff -U$_forgit_preview_context -- | $_forgit_diff_pager"
     enter_cmd="cd '$repo' && $get_files | xargs -0 $git_diff -U$_forgit_fullscreen_context -- | $_forgit_diff_pager"
     opts="
@@ -157,7 +171,7 @@ _forgit_diff() {
         $FORGIT_DIFF_FZF_OPTS
         --prompt=\"$commits > \"
     "
-    eval "git diff --name-status $FORGIT_DIFF_GIT_OPTS $commits -- ${files[*]} | sed -E 's/^([[:alnum:]]+)[[:space:]]+(.*)$/[\1]	\2/'" |
+    eval "git diff --name-status ${_forgit_diff_git_opts[*]} $commits -- ${files[*]} | sed -E 's/^([[:alnum:]]+)[[:space:]]+(.*)$/[\1]	\2/'" |
         sed 's/	/  ->  /2' | expand -t 8 |
         FZF_DEFAULT_OPTS="$opts" fzf
     fzf_exit_code=$?
@@ -170,7 +184,9 @@ _forgit_diff() {
 _forgit_add() {
     _forgit_inside_work_tree || return 1
     local git_add changed unmerged untracked files opts preview extract
-    git_add="git add $FORGIT_ADD_GIT_OPTS"
+    _forgit_add_git_opts=()
+    _forgit_parse_array _forgit_add_git_opts "$FORGIT_ADD_GIT_OPTS"
+    git_add="git add ${_forgit_add_git_opts[*]}"
     # Add files if passed as arguments
     [[ $# -ne 0 ]] && { $git_add "$@" && git status -su; return $?; }
 
@@ -210,7 +226,9 @@ _forgit_add() {
 _forgit_reset_head() {
     _forgit_inside_work_tree || return 1
     local git_reset_head cmd files opts rootdir
-    git_reset_head="git reset -q $FORGIT_RESET_HEAD_GIT_OPTS HEAD"
+    _forgit_reset_head_git_opts=()
+    _forgit_parse_array _forgit_reset_head_git_opts "$FORGIT_RESET_HEAD_GIT_OPTS"
+    git_reset_head="git reset -q ${_forgit_reset_head_git_opts[*]} HEAD"
     [[ $# -ne 0 ]] && { $git_reset_head "$@" && git status --short; return $?; }
     rootdir=$(git rev-parse --show-toplevel)
     cmd="git diff --staged --color=always -- $rootdir/{} | $_forgit_diff_pager "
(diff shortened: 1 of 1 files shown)


bin/git-forgit
```diff
@@ -146,8 +203,10 @@ _forgit_diff() {
     # prevent fzf from interpreting this substring by escaping the opening bracket.
     # The string is evaluated a few subsequent times, so we need multiple escapes.
     escaped_commits=${commits//\{/\\\\\{}
-    git_diff="git diff --color=always $FORGIT_DIFF_GIT_OPTS $escaped_commits"
-    preview_cmd="cd '$repo' && $get_files | xargs -0 $git_diff -U$_forgit_preview_context -- | $_forgit_diff_pager"
+    _forgit_diff_git_opts=()
+    _forgit_parse_array _forgit_diff_git_opts "$FORGIT_DIFF_GIT_OPTS"
+    git_diff="git diff --color=always ${_forgit_diff_git_opts[*]} $escaped_commits"
+    preview_cmd="cd '$repo' && $FORGIT diff_view {} $_forgit_preview_context $escaped_commits"
```
sandr01d, line 209: I think we should also remove the deferred execution from this line in this PR.

sandr01d:
I did all the changes I meant to do here. Awaiting your reviews @carlfriedrich @cjappl.

cjappl:
Sounds good! A lot to review, but I'll take it through it's paces on fish as a system test today and report back on how that goes :)

carlfriedrich:
@sandr01d Thanks for the patch! I will do a review in the following days. This time I'll make sure to view it directly on my desktop monitor. 😁

bin/git-forgit
```diff
@@ -244,21 +302,30 @@ _forgit_reset_head() {
     echo 'Nothing to unstage.'
 }
 
+_forgit_stash_show_preview() {
```
carlfriedrich, line 306: LGTM

bin/git-forgit
```diff
@@ -268,6 +335,14 @@ _forgit_stash_show() {
     return $fzf_exit_code
 }
 
+_forgit_stash_push_preview() {
```
carlfriedrich, line 339: LGTM

bin/git-forgit
```diff
@@ -328,9 +396,12 @@ _forgit_clean() {
     echo 'Nothing to clean.'
 }
 
-_forgit_cherry_pick() {
-    local git_cherry_pick base target preview opts fzf_selection fzf_exitval
+_forgit_cherry_pick_preview() {
+    echo "$1" | cut -f2- | _forgit_extract_sha | xargs -I% git show --color=always % | $_forgit_show_pager
+}
 
+_forgit_cherry_pick() {
```
carlfriedrich, line 404: LGTM

bin/git-forgit
```diff
@@ -371,9 +441,13 @@ _forgit_cherry_pick() {
     $git_cherry_pick "${commits[@]}"
 }
 
+_forgit_cherry_pick_from_branch_preview() {
```
carlfriedrich, line 445: LGTM

bin/git-forgit
```diff
@@ -436,23 +508,29 @@ _forgit_rebase() {
     fi
 }
 
+_forgit_file_preview() {
```
carlfriedrich, line 512: LGTM

bin/git-forgit
```diff
@@ -462,22 +540,24 @@ _forgit_fixup() {
         # GIT_SEQUENCE_EDITOR=: is needed to skip the editor
         GIT_SEQUENCE_EDITOR=: git rebase --autostash -i --autosquash "$prev_commit"
     fi
+}
 
+_forgit_checkout_file_preview() {
```
carlfriedrich, line 546: LGTM

bin/git-forgit
```diff
@@ -532,38 +611,40 @@ _forgit_checkout_branch() {
...
-        --preview=\"$preview\"
+        --preview=\"$FORGIT branch_preview {}\"
         $FORGIT_CHECKOUT_TAG_FZF_OPTS
     "
     tag="$(eval "$cmd" | FZF_DEFAULT_OPTS="$opts" fzf)"
     [[ -z "$tag" ]] && return 1
     $git_checkout "$tag"
 }
 
+_forgit_checkout_commit_preview() {
```
carlfriedrich, line 632: LGTM

bin/git-forgit
```diff
@@ -573,19 +654,22 @@ _forgit_checkout_commit() {
         FZF_DEFAULT_OPTS="$opts" fzf | eval "$_forgit_extract_sha" | xargs -I% $git_checkout % --
 }
 
+_forgit_branch_preview() {
```
carlfriedrich, line 658: LGTM

bin/git-forgit
```diff
@@ -595,10 +679,18 @@ _forgit_branch_delete() {
     echo -n "$branches" | tr '\n' '\0' | xargs -I{} -0 $git_branch -D {}
 }
 
+_forgit_revert_preview() {
```
carlfriedrich, line 683: LGTM

bin/git-forgit
```diff
@@ -636,10 +727,20 @@ _forgit_revert_commit() {
     $git_revert "${commits[@]}"
 }
 
+_forgit_blame_preview() {
```
carlfriedrich, line 731: LGTM

bin/git-forgit
```diff
@@ -648,33 +749,35 @@ _forgit_blame() {
...
-    eval $git_blame "$file" "$flags"
+    eval $git_blame "$file" "${flags[*]}"
 }
 
 # git ignore generator
 export FORGIT_GI_REPO_REMOTE=${FORGIT_GI_REPO_REMOTE:-https://github.com/dvcs/gitignore}
 export FORGIT_GI_REPO_LOCAL="${FORGIT_GI_REPO_LOCAL:-${XDG_CACHE_HOME:-$HOME/.cache}/forgit/gi/repos/dvcs/gitignore}"
 export FORGIT_GI_TEMPLATES=${FORGIT_GI_TEMPLATES:-$FORGIT_GI_REPO_LOCAL/templates}
 
+_forgit_ignore_preview() {
```
carlfriedrich, line 771: LGTM

bin/git-forgit
```diff
@@ -736,18 +839,37 @@ valid_commands=(
     "stash_push"
 )
 
+private_commands=(
```
carlfriedrich, line 843: A note about this change should also go into the commit message IMO.

carlfriedrich (changes requested):
Great work @sandr01d, thanks for extracting this patch! Can you fix the few things I requested and then squash all the changes into one commit? Thanks a lot!

cjappl:
> squash all the changes into one commit

Is there an advantage to doing this before "squash and merge" when we merge the final one @carlfriedrich ? You keep suggesting it, which makes me think I'm overlooking some upside to the approach.

In my perspective, it seems like manually squashing and merging is unneeded extra work if the work will just be squashed-and-merged automagically by github later. 

Is there a difference in the two approaches that makes the manual way better?

carlfriedrich:
> Is there an advantage to doing this before "squash and merge" when we merge the final one @carlfriedrich ? You keep suggesting it, which makes me think I'm overlooking some upside to the approach.

@cjappl Generally speaking, it gives us the possibility to review the commit message. This is not the case when GitHub's squash and merge feature is used, because that happens after the review. And since the commit message is used for the changelog in our release notes, it should be part of the review IMO, especially given the fact that it cannot be changed once it is merged to master.

In this specific case it is also crucial that we don't want to squash all the commits, since this PR also contains the base commit which will be merged in another PR. This makes it even more important to keep the commits tidy.

cjappl:
Sweet, thanks for the explanation @carlfriedrich !

> @cjappl Generally speaking, it gives us the possibility to review the commit message. This is not the case when GitHub's squash and merge feature is used, because that happens after the review. And since the commit message is used for the changelog in our release notes, it should be part of the review IMO, especially given the fact that it cannot be changed once it is merged to master.

Nice, good point, I hadn't considered that. "Make commit messages reviewable"

> In this specific case it is also crucial that we don't want to squash all the commits, since this PR also contains the base commit which will be merged in another PR. This makes it even more important to keep the commits tidy.

Makes sense to me! I think in this specific review case (considering the complexity of it all) I'm all for anything that provides reviewability for it. I'm pro-squash for this review.


I don't necessarily think we need to do this for 100% of PRs in forgit, but we can maybe discuss on an Issue ticket if we want to implement that process change for every commit going forward, I think there are mostly "pros" for complex multi-file multi-commit branches, and mostly "cons" for simple changes.

bin/git-forgit
```diff
@@ -92,21 +96,30 @@ _forgit_log_format=${FORGIT_LOG_FORMAT:-%C(auto)%h%d %s %C(black)%C(bold)%cr%Cre
 _forgit_log_preview_options="--graph --pretty=format:'$_forgit_log_format' --color=always --abbrev-commit --date=relative"
 _forgit_fullscreen_context=${FORGIT_FULLSCREEN_CONTEXT:-10}
 _forgit_preview_context=${FORGIT_PREVIEW_CONTEXT:-3}
-_forgit_is_file_tracked="(git ls-files {} --error-unmatch) &> /dev/null"
+
+_forgit_is_file_tracked() {
```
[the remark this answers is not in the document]
sandr01d: This one is a bit tricky, because the string contains the `{}` placeholder that fzf replaces. Since this is no longer passed to fzf directly and instead used in the `_forgit_*_preview` functions now, we need to pass the file from fzf to `$_forgit_is_file_tracked` somehow. I am not aware of a way to do this without converting it into a function. Since we replace all usages of it, I think it would be the best to replace it with a function in this PR.

bin/git-forgit
```diff
@@ -70,6 +70,10 @@ hash emojify &>/dev/null && _forgit_emojify='|emojify'
 
 # extract the first git sha occurring in the input and strip trailing newline
 _forgit_extract_sha="grep -Eo '[a-f0-9]+' | head -1 | tr -d '[:space:]'"
+_forgit_extract_sha() {
```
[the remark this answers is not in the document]
sandr01d: This can be done and I removed the `_forgit_extract_sha` function in ab5a9e1. Unfortunately we need to ignore a lot of shellchecks and introduce more usage of `eval` to do so, which I am not happy about. From my perspective it would be better to keep the function in this case. I'll leave the decision to you, if you want to keep the changes from ab5a9e1 I will squash them into the previous commit, otherwise I will remove them. WDYT?

bin/git-forgit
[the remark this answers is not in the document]
carlfriedrich: Ah, I see, didn't notice that. Shure, then let's keep it as it is.

bin/git-forgit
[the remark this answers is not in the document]
carlfriedrich: Ah, I see. Well, then let's remove it again. Thanks for trying, though!

carlfriedrich (approved):
@sandr01d Thanks for the loop, changes and commit message approved. You can remove the last commit and then I'd consider this PR ready.

Shall I rebase #326 on this PR then?

carlfriedrich:
Ah wait, we have #343 next in the line. @sandr01d Will you rebase that on top of this PR then first?

sandr01d:
Yes, sure. Don't have the time today, but plan to do so tomorrow.

sandr01d:

PR #354, opened 2024-02-25
Fix diff preview and edit for files with a ']' character in their name

### LINKED ISSUE #353: `[]` in a file name brakes the preview window
## Check list

- [X] I have read through the [README](https://github.com/wfxr/forgit/blob/master/README.md)
- [X] I have the latest version of forgit
- [X] I have searched through the existing issues

## Environment info

- OS
    - [ ] Linux
    - [X] Mac OS X
    - [ ] Windows
    - [ ] Others:
- Shell
    - [X] bash
    - [ ] zsh
    - [ ] fish

* Have a file named `foo/bar/[baz]/buzz.py` in your git repo.
* Use one of the forgit command to get a fzf window
* The `[]` in the file name makes the preview window brake (picture attached).

image

Sadly this is becoming very common file names due to remix/next and other frontend frameworks with file based routing.
https://nextjs.org/docs/pages/building-your-application/routing/dynamic-routes

Thanks 🙇

sandr01d:
## Check list

## Description

Our sed command for removing the short status (e.g. [M]) from gits
output to extract the file name with \_forgit\_diff matched until the
last ']' character in the line due to sed being greedy. This created
issues with file names that contain a ']' character.
To fix this, I made sure that only the short status is removed by sed,
independently of the file name.
Fixes #353

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [X] zsh
    - [ ] fish
- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -137,10 +137,10 @@ _forgit_diff() {
     #   oldfile\0
     # We have to do a two-step sed -> tr pipe because OSX's sed implementation does
     # not support the null-character directly.
-    get_files="echo {} | sed 's/.*] *//' | sed 's/  ->  /\\\n/' | tr '\\\n' '\\\0'"
+    get_files="echo {} | sed 's/\\\\s*\\\\[.]\\\\s*//' | sed 's/  ->  /\\\n/' | tr '\\\n' '\\\0'"
     # Similar to the line above, but only gets a single file from a single line
     # Gets the new name of renamed files
-    get_file="echo {} | sed 's/.*] *//' | sed 's/.*->  //'"
+    get_file="echo {} | sed 's/\\\\s*\\\\[.]\\\\s*//' | sed 's/.*->  //'"
     # Git stashes are named "stash@{x}", which contains the fzf placeholder "{x}".
     # In order to support passing stashes as arguments to _forgit_diff, we have to
     # prevent fzf from interpreting this substring by escaping the opening bracket.


cjappl (approved):
Seems reasonable to me.

Off topic but is there any way for us to have unit tests in this project in done kind of structured way? This would be the perfect thing to have a test for to make sure we don't regress. (This is a bigger topic that 100% does not block this review)

sandr01d:
> Off topic but is there any way for us to have unit tests in this project in done kind of structured way? This would be the perfect thing to have a test for to make sure we don't regress. (This is a bigger topic that 100% does not block this review)

That would be great. Did a quick search and [this GitHub Actions](https://github.com/marketplace/actions/generate-shell-script-unit-testing) looks promising (haven't done any in depth research, so I don't know if it's any good). Maybe we should create a separate issue for this?

cjappl:
Yeah these look great! We should definitely open a ticket and we can tackle it when we are able. Good finds.

bin/git-forgit
```diff
@@ -137,10 +137,10 @@ _forgit_diff() {
     #   oldfile\0
     # We have to do a two-step sed -> tr pipe because OSX's sed implementation does
     # not support the null-character directly.
-    get_files="echo {} | sed 's/.*] *//' | sed 's/  ->  /\\\n/' | tr '\\\n' '\\\0'"
+    get_files="echo {} | sed 's/\\\\s*\\\\[.]\\\\s*//' | sed 's/  ->  /\\\n/' | tr '\\\n' '\\\0'"
```
carlfriedrich, line 140: Oh wow, respect that you fixed this on the current codebase instead of the new version without the deferred code. This looks so ugly, haha! I am so looking forward to get rid of all these escapes. 😅

carlfriedrich (approved):

bin/git-forgit
sandr01d: Yeah, I actually did the fix in the new codebase and backported it. Troubleshooting issues like this is **a lot** easier in the new code base IMO :slightly_smiling_face:

PR #355, opened 2024-02-27
Refactor: Replace deferred git commands with regular git commands

sandr01d:
## Check list

## Description

We were using deferred code in git commands in some places without any apparent reason. Each of these deferred code snippets was only executed a single time, so we can replace them with regular git commands.
This PR replaces the remaining git commands that were not removed in #343, because they were not moved into functions.

## Note

This PR belongs to #326 and resulted from discussions in #324.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [X] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -66,7 +66,13 @@ _forgit_contains_non_flags() {
 }
 
 # optional render emoji characters (https://github.com/wfxr/emoji-cli)
-hash emojify &>/dev/null && _forgit_emojify='|emojify'
+_forgit_emojify() {
+    if hash emojify &>/dev/null; then
+        emojify
+    else
+        cat
+    fi
+}
 
 # extract the first git sha occurring in the input and strip trailing newline
 _forgit_extract_sha="grep -Eo '[a-f0-9]+' | head -1 | tr -d '[:space:]'"
@@ -127,7 +133,8 @@ _forgit_log() {
     log_format=${FORGIT_GLO_FORMAT:-$_forgit_log_format}
     _forgit_log_git_opts=()
     _forgit_parse_array _forgit_log_git_opts "$FORGIT_LOG_GIT_OPTS"
-    eval "git log $graph --color=always --format='$log_format' ${_forgit_log_git_opts[*]} $* $_forgit_emojify" |
+    eval "git log $graph --color=always --format='$log_format' ${_forgit_log_git_opts[*]} $*" |
+        _forgit_emojify |
         FZF_DEFAULT_OPTS="$opts" fzf
     fzf_exit_code=$?
     # exit successfully on 130 (ctrl-c/esc)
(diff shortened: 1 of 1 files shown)


sandr01d:
This should be the last big PR. After this, only small changes are left :rocket:

bin/git-forgit
```diff
@@ -180,52 +181,41 @@ _forgit_diff_view() {
...
-    _forgit_diff_git_opts=()
-    _forgit_parse_array _forgit_diff_git_opts "$FORGIT_DIFF_GIT_OPTS"
-    git_diff="git diff --color=always ${_forgit_diff_git_opts[*]} $escaped_commits"
-    enter_cmd="cd '$repo' && $get_files | xargs -0 $git_diff -U$_forgit_fullscreen_context -- | $_forgit_diff_pager"
+    for commit in "${commits[@]}"; do
+        escaped_commits+="'${commit//\{/\\\\\{}' "
+    done
     opts="
         $FORGIT_FZF_DEFAULT_OPTS
-        +m -0 --bind=\"enter:execute($enter_cmd | $_forgit_enter_pager)\"
```
carlfriedrich, line 154: I think the `enter_cmd` change here and everything related to it does not fit with the commit description. I would suggest removing it from this commit and do it later together with the same thing in `_forgit_log`, where the `enter_cmd` still exists.

sandr01d: Sure, done now.

bin/git-forgit
```diff
@@ -716,9 +719,12 @@ _forgit_branch_delete() {
...
 
-    cmd="git branch --color=always | LC_ALL=C sort -k1.1,1.1 -rs"
-    for branch in $(eval "$cmd" | FZF_DEFAULT_OPTS="$opts" fzf | awk '{print $1}'); do
-        _forgit_git_branch_delete "$branch"
+    for branch in $(git branch --color=always |
+        LC_ALL=C sort -k1.1,1.1 -rs |
+        FZF_DEFAULT_OPTS="$opts" fzf |
+        awk '{print $1}')
+    do
+        _forgit_branch_delete "$branch"
```
carlfriedrich, line 727: I think this should stay `_forgit_git_branch_delete`, shouldn't it?

carlfriedrich (changes requested):
@sandr01d Overall this looks great for me, thanks for the work again. Only thing I would like to see is a note in the commit message about why the `FORGIT_LOG_GRAPH_ENABLE` to `_forgit_log_graph_enable` change has been necessary along with this. (I assume it is necessary for this commit? Otherwise I would prefer to have a dedicated commit for it.)

bin/git-forgit
sandr01d: Yes, good catch!

carlfriedrich (approved):

carlfriedrich:
@sandr01d Perfect, thanks for the update! Will you do the `enter_cmd` changes in a next PR?

sandr01d:

PR #356, opened 2024-03-01
Fix homebrew update action to run on tag push

cjappl:
The first time the homebrew auto update ran, it failed:

https://github.com/wfxr/forgit/actions/runs/8104493571/job/22151209325

The readme of our package we are using:
https://github.com/dawidd6/action-homebrew-bump-formula?tab=readme-ov-file

May rely on the tag push to extract some information, so I modified it to match their recommendation.

### DIFF
--- a/.github/workflows/brew.yaml
+++ b/.github/workflows/brew.yaml
@@ -1,11 +1,9 @@
 name: Homebrew
 
 on:
-  workflow_run:
-    workflows:
-      - "Release"
-    types:
-      - completed
+  push:
+    tags:
+      - '*'
 
 jobs:
   brew:


wfxr (approved):
LGTM

carlfriedrich:
@cjappl FYI: you don't have to wait for the next automatic release to test this, you can create a custom tag with the fix version increased at any time (e.g. 24.03.1).

cjappl:
@carlfriedrich cool!! I wasn't entirely sure if those were generated automatically or not.

Do we have any sort of process around tag making? Anything that needs to exist in the name of the tag or the description? 

I notice they're generally nice and uniform, so I want to make sure I don't mess anything up

carlfriedrich:
@cjappl Basically you can create tags with any name. However, the versioning scheme we decided on for forgit follows the structure `year.month.fix`, so I would suggest sticking to that by just counting up the `fix` value after the last dot.

cjappl:
Cool thanks @carlfriedrich ! Will test sometime next week when I'm back at a computer.

PR #358, opened 2024-03-03
Refactor: Replace deferred code in enter commands with functions

sandr01d:
## Check list

## Description

In `_forgit_log` and `_forgit_enter` it is possible to diff a single commit/file by pressing enter. We used to store the code that executes the diffs in variables and passed it to fzf as deferred code. This refactor reduces the amount of deferred code by using functions instead of variables.

### Note

This PR belongs to #326 and resulted from discussions in #324.
In comparison to #326 I've replaced the `$enter_cmd` variable in `_forgit_diff` with a function aswell. The part of it that `cd`s into the repo directory has been removed because this is handled by `_forgit_diff_view` already.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [X] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [X] zsh
    - [ ] fish
- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -66,7 +66,13 @@ _forgit_contains_non_flags() {
 }
 
 # optional render emoji characters (https://github.com/wfxr/emoji-cli)
-hash emojify &>/dev/null && _forgit_emojify='|emojify'
+_forgit_emojify() {
+    if hash emojify &>/dev/null; then
+        emojify
+    else
+        cat
+    fi
+}
 
 # extract the first git sha occurring in the input and strip trailing newline
 _forgit_extract_sha="grep -Eo '[a-f0-9]+' | head -1 | tr -d '[:space:]'"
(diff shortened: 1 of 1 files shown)


sandr01d:
I've noticed there is still some deferred code left in the pager variables (e.g. `$_forgit_enter_pager`). I think we should replace these with functions as well. Not in this PR though. We might want to do it after all changes in #326 have been reviewed or even after #326 has been merged.

carlfriedrich (approved):

carlfriedrich:
@sandr01d Great, thanks for the patch. I think next could be the `execute_silent` commands, right? What else have we left?

> I've noticed there is still some deferred code left in the pager variables (e.g. `$_forgit_enter_pager`). I think we should replace these with functions as well. Not in this PR though. We might want to do it after all changes in #326 have been reviewed or even after #326 has been merged.

That's a good idea, and I indeed think we can do it after we finished the ongoing review.

cjappl (approved):

sandr01d:
@carlfriedrich 
> I think next could be the execute_silent commands, right?

I think we should do `_forgit_extract_sha` first because that's used in the `execute-silent` commands a few times.

carlfriedrich:
> @carlfriedrich
> 
> > I think next could be the execute_silent commands, right?
> 
> I think we should do `_forgit_extract_sha` first because that's used in the `execute-silent` commands a few times.
> 
> > What else have we left?
> 
> Apart from those two, I think only `_forgit_quote_files` is left. I might split the `execute_silent` commands into multiple PRs though (edit, yank commit and yank stash).

Great, let's do it like that!

sandr01d:

PR #360, opened 2024-03-06
Delete brew workflow

cjappl:
It turns out we are on the brewbot "auto bump" list so we don't need this workflow at all.

What a whirlwind.

Proof is in the logs here:
https://github.com/wfxr/forgit/actions/runs/8167132514/job/22327036492

```
/home/linuxbrew/.linuxbrew/bin/brew bump-formula-pr --no-audit --no-browse --message=[`action-homebrew-bump-formula`](https://github.com/dawidd6/action-homebrew-bump-formula) --version=24.03.1 forgit
Error: Whoops, the forgit formula has its version update
pull requests automatically opened by BrewTestBot!
We'd still love your contributions, though, so try another one
that's not in the autobump list:
  https://github.com/Homebrew/homebrew-core/blob/master/.github/autobump.txt
```

If you look at the autobump list and search "forgit" we are in there:
https://github.com/Homebrew/homebrew-core/blob/master/.github/autobump.txt


cjappl:
Looks like they have a github action that every 3 hrs-ish runs through the list of the autobump formulae and auto bumps them.

https://github.com/Homebrew/homebrew-core/blob/master/.github/workflows/autobump.yml

Cool!

carlfriedrich (approved):
Great, didn't know about that Bot either. :-)

PR #345, opened 2024-02-05
fix: keep the script compatible with Busybox applets

folliehiyuki:
* Use short flags for `sort` command
* Remove `-n` flag for `nl` command. The numbers in the selection list won't be aligned anymore, but no information is lost.

The current arguments are only presented in GNU's coreutils.

## Check list

- [x] I have performed a self-review of my code
- [ ] I have commented my code in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

PR #365, opened 2024-03-11
Fix issue where diff preview was broken on mac

### LINKED ISSUE #362: diff preview not working on MacOS
## Check list

## Environment info

- OS
    - [ ] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Others:
- Shell
    - [x] bash
    - [x] zsh
    - [x] fish

Hello,

I've recently ported my forgit configuration from a Linux machine to my MacOS laptop. I've noticed previews are not working for the `gd` command. For example, using the default configuration:
Screenshot 2024-03-09 at 13 58 22

Other commands using the same pager, such as `ga`, seem to work as expected:
Screenshot 2024-03-09 at 14 00 27

Did anyone else encounter the same issue?

Cheers

cjappl:
Closes #362 

I need one of my favorite linux pals (@sandr01d , @wfxr , @carlfriedrich ) to ensure this still works on your favorite linux distro.

The test cases I would recommend:
1. File with normal name
2. File with spaces in the name 
3. File with brackets in the name (`my[test]file.txt`)
4. File that has been renamed.

I have checked on my mac and docker image and everything seems happy.

## Check list

## Type of change

- [ ] Bug fix

## Test environment

- Shell
    - [x] bash
    - [ ] zsh
    - [x] fish
- OS
    - [x] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Others:


bin/git-forgit
```diff
@@ -137,10 +137,10 @@ _forgit_diff() {
     #   oldfile\0
     # We have to do a two-step sed -> tr pipe because OSX's sed implementation does
     # not support the null-character directly.
-    get_files="echo {} | sed 's/\\\\s*\\\\[.]\\\\s*//' | sed 's/  ->  /\\\n/' | tr '\\\n' '\\\0'"
+    get_files="echo {} | sed -e 's/^[[:space:]]*\\\\[[A-Z]\\\\][[:space:]]*//' | sed 's/  ->  /\\\n/' | tr '\\\n' '\\\0'"
```
cjappl, line 140: For this, I changed it to "Remove everything that looks like a capital letter in brackets and a bunch of spaces following it"

Please sanity check if you agree that this makes sense.

I was seeing the pattern

`[M]      somefile.txt`
`[M]      another[file].txt`

carlfriedrich:
I can confirm that files with normal name, files with spaces in the name and files with brackets in the name all work on this branch.
Renamed files, however, are not correctly diffed. The diff preview displays the file as if it was added. This is because the `get_files` code does not correctly extract the two filenames from the line. We can check that by setting the `preview_cmd` to this:

```bash
preview_cmd="$get_files | xargs -0 -I% echo '%'"
```

The preview displays a list of the contained files in the change then. For a modified file, it displays just the single filename. For a renamed file, it should display both filenames, the old and the new one. In my case it displays:

```
[R100]  old_filename
new_filename
```

So the `[R100]` is not correctly removed from the beginning.

As I found out, this is the case since #354 already, so it wasn't introduced in this PR. However, I think we should fix it in this PR.

bin/git-forgit
carlfriedrich: This does the trick for me:

`get_files="echo {} | sed -e 's/^[[:space:]]*\\\\[[A-Z0-9]*\\\\][[:space:]]*//' | sed 's/  ->  /\\\n/' | tr '\\\n' '\\\0'"`

The previous `[A-Z]` did not match the `R100` for renamed files.

carlfriedrich (changes requested):

carlfriedrich:
And I am extremely looking forward to covering these cases in a unit test. :-)

cjappl:
@carlfriedrich 100% on the unit tests, so many regressions in this area..

Can you send me the exact "steps to repro" for your R100 case? I will run it on my mac and ensure your xargs suggestion works then update this PR

carlfriedrich:
@cjappl Just call `git mv old_filename new_filename` and then forgit diff on it.

cjappl:
@carlfriedrich would you please run through the test cases again. I took your suggestion, plus one extra modification that was necessary (.* around the arrow, matching get_file).

bin/git-forgit
```diff
@@ -137,10 +137,10 @@ _forgit_diff() {
     #   oldfile\0
     # We have to do a two-step sed -> tr pipe because OSX's sed implementation does
     # not support the null-character directly.
-    get_files="echo {} | sed 's/\\\\s*\\\\[.]\\\\s*//' | sed 's/  ->  /\\\n/' | tr '\\\n' '\\\0'"
+    get_files="echo {} | sed -e 's/^[[:space:]]*\\\\[[A-Z0-9]*\\\\][[:space:]]*//' | sed 's/.*->  //' | tr '\\\n' '\\\0'"
```
sandr01d, line 140: This still does not work for me when diffing renamed files that were added. To reproduce:

1. mv LICENSE LICENSE1
2. git add .
3. gd --staged

The preview will look like LICENSE1 has been added. This can be fixed like this:

```suggestion
    get_files="echo {} | sed -e 's/^[[:space:]]*\\\\[[A-Z0-9]*\\\\][[:space:]]*//' | sed 's/.*->  //' | tr '\\\n' '\\\0'"
    get_files="echo {} | sed -e 's/^[[:space:]]*\\\\[[A-Z0-9]*\\\\][[:space:]]*//' | sed 's/  ->  /\\\n/' | tr '\\\n' '\\\0'"
```

The second sed command is what has changed here. I realized that I accidentally modified it with #354. Sorry you for that and another strong yes for unit tests.

sandr01d:
I also noticed that files with a leading white space fail to get a preview. This seems to have been the case before your changes and even before #354. So we could leave this open for a follow up PR IMO. Other than this and the change I've suggested above, this is working fine for me.

bin/git-forgit
cjappl: I think I see what you're seeing, but your suggestion doesn't work on my mac. In my current code I have on this branch, I get this:

Screenshot 2024-03-12 at 2 14 54 PM


Your suggestion leads to a blank preview (on both ubuntu and mac):
Screenshot 2024-03-12 at 2 16 44 PM

That is using this line:
```
get_files="echo {} | sed -e 's/^[[:space:]]*\\\\[[A-Z0-9]*\\\\][[:space:]]*//' | sed 's/ -> /\\\n/' | tr '\\\n' '\\\0'"
```


I realize my "working" version says "file added" at the top, but I'm not sure how it is supposed to look. Can you post a screenshot of what the working version looks like on your machine?

sandr01d: Without my change it looks like the first one. With the change, it looks like this for me:

image

That the diff is empty is correct, since the content of the file did not change. Could you check how this behaved before #354 for you? My suggestion simply adds back the code that has been replaced in #354 by accident, so it surprises me a bit that this would break something.

cjappl: Alright, I cannot explain it but after going in and copy/pasting from the original pre-#345 version (which is identical to the code you pasted above) I can confirm this works on both Mac and Ubunutu. I pushed to this branch.

Please test the test cases again :) Thanks for the help as always

sandr01d (approved):
This works for me as expected now. That is, there are still some file names that do not work, but that never actually worked. The ones I've found so far are file names that start with a white space (probably because we remove leading white space in the `sed` command) and files that have a backslash in their name (probably because it being interpreted as an escape character).
I think we can only fixes those by implementing a completely new approach, so clearly out of scope for this PR. So approved from my side.
Thanks @cjappl!

carlfriedrich (approved):

carlfriedrich:
@cjappl Please add a note in the final commit message that this also fixes diff for renames, which was broken since forgit-24.03.0.

carlfriedrich:
@cjappl Didn't read my last comment? That's why I prefer squashing before merging. :-/

cjappl:
> @cjappl Didn't read my last comment? That's why I prefer squashing before merging. :-/

Sorry I missed it, I had the page loaded before merging and I didn't see it.

cjappl:
@carlfriedrich I'm going to do a release to get this out to the package managers, do you want me to mention this in the release notes: "fixes diff preview for renames"

Currently the "auto generated" release notes are:

```
## What's Changed
* Delete brew workflow by @cjappl in https://github.com/wfxr/forgit/pull/360
* Fix issue where diff preview was broken on mac by @cjappl in https://github.com/wfxr/forgit/pull/365


**Full Changelog**: https://github.com/wfxr/forgit/compare/24.03.1...24.03.2
```

Normally I haven't varied from the auto generated, but in this case I could add a line to be:
```
## What's Changed
* Delete brew workflow by @cjappl in https://github.com/wfxr/forgit/pull/360
* Fix issue where diff preview was broken on mac by @cjappl in https://github.com/wfxr/forgit/pull/365
* Fixed regression where diff preview for renames was broken by @cjappl in https://github.com/wfxr/forgit/pull/365


**Full Changelog**: https://github.com/wfxr/forgit/compare/24.03.1...24.03.2
```

carlfriedrich:
@cjappl Yep, that would be great, thanks a lot!

cjappl:
@carlfriedrich No problem. Apologies again on missing your comment. I'll do that release now.

PR #368, opened 2024-03-15
Refactor: Quote files passed to preview functions

sandr01d:
## Check list

## Description

In some cases we need to pass multiple files to preview functions. These files are treated as a single string that is evaluated by fzf and passed on to our preview functions as individual arguments.
This PR introduces a `_forgit_quote_files` function that ensures the resulting arguments are quoted properly.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [X] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [X] zsh
    - [ ] fish
- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -66,7 +66,13 @@ _forgit_contains_non_flags() {
 }
 
 # optional render emoji characters (https://github.com/wfxr/emoji-cli)
-hash emojify &>/dev/null && _forgit_emojify='|emojify'
+_forgit_emojify() {
+    if hash emojify &>/dev/null; then
+        emojify
+    else
+        cat
+    fi
+}
 
 # extract the first git sha occurring in the input and strip trailing newline
 _forgit_extract_sha="grep -Eo '[a-f0-9]+' | head -1 | tr -d '[:space:]'"
@@ -127,7 +133,8 @@ _forgit_log() {
     log_format=${FORGIT_GLO_FORMAT:-$_forgit_log_format}
     _forgit_log_git_opts=()
     _forgit_parse_array _forgit_log_git_opts "$FORGIT_LOG_GIT_OPTS"
-    eval "git log $graph --color=always --format='$log_format' ${_forgit_log_git_opts[*]} $* $_forgit_emojify" |
+    eval "git log $graph --color=always --format='$log_format' ${_forgit_log_git_opts[*]} $*" |
+        _forgit_emojify |
         FZF_DEFAULT_OPTS="$opts" fzf
     fzf_exit_code=$?
     # exit successfully on 130 (ctrl-c/esc)
@@ -511,7 +518,7 @@ _forgit_rebase() {
     git_rebase="git rebase -i ${_forgit_rebase_git_opts[*]}"
     graph=--graph
     [[ $FORGIT_LOG_GRAPH_ENABLE == false ]] && graph=
-    cmd="git log $graph --color=always --format='$_forgit_log_format' $* $_forgit_emojify"
+    cmd="git log $graph --color=always --format='$_forgit_log_format' $*"
     files=$(sed -nE 's/.* -- (.*)/\1/p' <<< "$*") # extract files parameters for `git show` command
     opts="
         $FORGIT_FZF_DEFAULT_OPTS
@@ -520,7 +527,7 @@ _forgit_rebase() {
         --preview=\"$FORGIT file_preview {} $files\"
         $FORGIT_REBASE_FZF_OPTS
     "
-    target_commit=$(eval "$cmd" | FZF_DEFAULT_OPTS="$opts" fzf | eval "$_forgit_extract_sha")
+    target_commit=$(eval "$cmd" | _forgit_emojify | FZF_DEFAULT_OPTS="$opts" fzf | eval "$_forgit_extract_sha")
     if [[ -n "$target_commit" ]]; then
         prev_commit=$(_forgit_previous_commit "$target_commit")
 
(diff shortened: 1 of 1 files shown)


sandr01d:
FYI: This is the last patch from #326 :rocket:

bin/git-forgit
```diff
@@ -104,6 +104,29 @@ _forgit_parse_array() {
     ${old_IFS+"false"} && unset IFS || IFS="$old_IFS"
 }
 
+# parse the input arguments and print only those after the "--"
+# separator as a single line of qouted arguments to stdout
```
cjappl, line 108: small typo "qouted"

cjappl (commented):
Just checking, this is only needed for fixup, rebase and log? Not any of the other file-based functions (add, reset head, for example)

bin/git-forgit
```diff
@@ -104,6 +104,29 @@ _forgit_parse_array() {
...
+    add=false
+    while (( "$#" )); do
+        case "$1" in
+            --)
+            add=true
+            shift
+            ;;
+            *)
+            if [ $add == true ]; then
+                files+=("'$1'")
```
cjappl, line 121: Forgive my ignorance on this more advanced shell syntax.

It seems like 114 our case statement is "I found a --, set add to true, and shift (meaning that positional arguments are shifted one to the left" then we add it to the array.


What happens with a list of files, aka:

`glo -- foo.txt bar.txt` Wouldn't the second one be skipped? Do we expect people to use `glo -- foo.txt -- bar.txt`?

(Completely possible I'm missing the main understanding here, just trying to understand what this means for the user)

sandr01d: No worries, bash syntax can be a bit confusing sometimes.

> It seems like 114 our case statement is "I found a --, set add to true, and shift (meaning that positional arguments are shifted one to the left" then we add it to the array.

Yes, that's correct. 

> What happens with a list of files, aka:
> glo -- foo.txt bar.txt Wouldn't the second one be skipped? 

No, let me walk you through this example:

The while loop runs as long as there are any there are any positional arguments present. In your example we have 3 of them ("--", "foo.txt" and "bar.txt").

1. Iteration:
The first positional argument (`$1`) has a value of  "--". This matches the first case: `$add` is set `true`, we shift and terminate the first iteration. Keep in mind that when using `;;`, the case statement does not "fall through",meaning the iteration ends without evaluating the other cases (this would only be the case when `;;&` is used as a terminator).
2. Iteration:
`$1` has a value of "foo.txt". This does not match the first case. It does match the second case (which simply matches everything). Since `$add` is `true`, "'foo.txt'" is added to the `$files` array and we shift again.
3. Iteration
`$1` is now "bar.txt" and the same thing happens as in the second one. "'bar'" is added to the array and we shift, causing `$#` to become 0 and terminate the while loop.

This implementation also ensures only arguments that come after -- are treated as files and will be added to the array, since everything before -- will simply be skipped over, since `$add` is still `false`.

Does that answer your question?

sandr01d:
> Just checking, this is only needed for fixup, rebase and log? Not any of the other file-based functions (add, reset head, for example)

Yes, add and reset head do exit early without showing a preview when arguments are passed. This is also only necessary for commands that allow specifying multiple files as arguments to filter what is displayed in the preview. This only relevant when the picking commits. In most other commands we pick (and therefore preview) single files, so filtering the preview is not necessary.

carlfriedrich (approved):
@sandr01d Great work! Thanks a lot for your patience and will to invest the time for splitting out the individual chunks. I think it was worth the work.

So I think next task is to rebase the branch on the current master, right? Shall we do some testing days afterwards? If so, I would suggest to freeze the master until we have merged this. Given the size of the branch I would also be fine with merging right away and fixing eventual regressions afterwards. We're in the middle of the month, so we have two weeks until the next regular release.

WDYT?

bin/git-forgit
cjappl: Thanks for the extremely in depth walkthrough! Appreciate it. Makes sense to me!

cjappl (approved):

cjappl:
> @sandr01d Great work! Thanks a lot for your patience and will to invest the time for splitting out the individual chunks. I think it was worth the work.

Big +1 on this. Good job @sandr01d !! And thank you @carlfriedrich for being the main reviewer on it. Good teamwork.

> So I think next task is to rebase the branch on the current master, right? Shall we do some testing days afterwards? If so, I would suggest to freeze the master until we have merged this. Given the size of the branch I would also be fine with merging right away and fixing eventual regressions afterwards. We're in the middle of the month, so we have two weeks until the next regular release.

1. I agree we should freeze master until after we merge this

I am split on the right approach for testing it. I think we should avoid at all costs shipping something that is broken in some way. 

My gut is telling me we should do some in branch testing on our favorite OSes and workflows for a few days, trying to really hammer it and fix any bugs that fall out obviously, then merge. I agree with "Given the size of the branch I would also be fine with merging right away and fixing eventual regressions afterwards."


Open to any and all of this. I have some time to devote to some intensive testing, so just ping me with when the thing is ready and I'll put it through it's paces.

wfxr (approved):

sandr01d:
Thank you guys, appreciate it! And thanks for your reviews. I definitively agree that it was absolutely worth it.

> So I think next task is to rebase the branch on the current master, right?

Yes, will do so shortly.

I agree that we should freeze the master branch until the refactor has been merged. I think we should take some extra time for testing. How about we all use #326 as our daily driver for a week? If we find any bugs I will try to fix them as soon as possible. In case we don't find any bugs that require big code changes we could merge once the week is over.

> I have some time to devote to some intensive testing, so just ping me with when the thing is ready and I'll put it through it's paces.

@cjappl  That would definitively be great! From my experience, when I break things it's usually for macOS users, due to slight differences in `sed`, the older bash version and my inability to test on this OS.

sandr01d:

PR #378, opened 2024-03-28
Refactor: Do not ignore shellcheck 2207

sandr01d:
## Check list

## Description

We sometimes capture multiline output in an array. We used to do so directly, while setting IFS to "\n" to control splitting. This has the downside that bashs glob expansion will be invoked.
I think the original reason we did use this technique was that it does work in both zsh and bash. Since we always execute bin/git-forgit in bash now, this is not necessary anymore. We can use `read -r` in such cases to add each line to the array without glob expansion. This is the idiomatic way to do so and recommended in [shellcheck 2207](https://www.shellcheck.net/wiki/SC2207). In my personal opinion this is also a cleaner and more straightforward approach than setting and resetting `$IFS`.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [X] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [X] bash
    - [X] zsh
    - [ ] fish
- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -499,11 +499,10 @@ _forgit_cherry_pick() {
     [[ $fzf_exitval != 0 ]] && return $fzf_exitval
     [[ -z "$fzf_selection" ]] && return $fzf_exitval
 
-    ${IFS+"false"} && unset old_IFS || old_IFS="$IFS"
-    IFS=$'\n'
-    # shellcheck disable=2207
-    commits=($(echo "$fzf_selection" | sort -n -k 1 | cut -f2 | cut -d' ' -f1 | _forgit_reverse_lines))
-    ${old_IFS+"false"} && unset IFS || IFS="$old_IFS"
+    commits=()
+    while IFS='' read -r commit; do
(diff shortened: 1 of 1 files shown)


cjappl:
Do we need to tweak our github action that runs shellcheck to enforce this now? Better than backsliding later on

wfxr (approved):
LGTM

wfxr:
> Do we need to tweak our github action that runs shellcheck to enforce this now? Better than backsliding later on

I think we'd better to keep some flexibility since many checks themselves have exceptions, like [SC1091](https://www.shellcheck.net/wiki/SC1091), [SC2145](https://www.shellcheck.net/wiki/SC2145), [SC2206](https://www.shellcheck.net/wiki/SC2206), etc. IMO we don't have to stick to all checks as long as we have a good reason.

cjappl (approved):

carlfriedrich (approved):

Issue #380, opened 2024-04-02
Forgit no longer allowing variables inside gitconfig for pager


blaineventurine:
## Check list

- [X] I have read through the [README](https://github.com/wfxr/forgit/blob/master/README.md)
- [X] I have the latest version of forgit
- [X] I have searched through the existing issues

## Environment info

- OS
    - [ ] Linux
    - [X] Mac OS X
    - [ ] Windows
    - [ ] Others:
- Shell
    - [ ] bash
    - [X] zsh
    - [x] fish

## Problem / Steps to reproduce

I have my Git pager set to `pager = delta --side-by-side --width $(expr ${COLUMNS} - 10) --wrap-max-lines unlimited`. 

The newest version of Forgit no longer allows variables to be used - it will just display `error: unexpected argument '10)' found`.
If I set it to `--width ${COLUMNS}`, it tells me `Invalid value for width: "${COLUMNS}" is not an integer`

Bpyassing Forgit and just calling `git diff` shows everything as I would expect.

cjappl:
Thanks for reporting! We had a big refactor recently and expected a little churn. 


@sandr01d I didn't have time to fully get into this tonight, but I can repro on mac/fish. I bisected and found this was the commit that introduced it:

https://github.com/wfxr/forgit/commit/bfffda68d80aae31f51f683371f2c642ba33b0e8

sandr01d:
Can reproduce and found the issue. I've created #381 to start a discussion on how to solve this with more details. This does not work for the the enter commands though (e.g. pressing enter in `gd` or `gss` to view a single file/stash) with the pager configuration provided, but this wasn't working before the last release, so I would argue this to be out of scope - at least for now.

blaineventurine:
Thanks a ton for the quick reply! I'm glad it's reproducible and not just a problem with me.

PR #323, opened 2023-08-15
Fix opening files in editor when path contains spaces or file was renamed

sandr01d:
## Check list

- [X] I have performed a self-review of my code
- [X] I have commented my code in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation

## Description

Escaped variables in _forgit_add() and _forgit_diff() to fix opening files that contain spaces.
Added a separate sed command for retrieving a file in _forgit_diff() that correctly handles renamed files (when opening the editor while diffing staged changes).

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [X] zsh
    - [ ] fish
- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

PR #381, opened 2024-04-02
Fix: Pager variables are not evaluated

sandr01d:
## Check list

## Description

Our pager variables, e.g. `$_forgit_blame_pager` used to be in deferred code blocks that were evaluated with `eval`. Since we've removed the usage of `eval` in #326, variables in the pagers no longer get expanded, as is the case with #380.
In this specific case, evaluation is what we want and deferred code is unavoidable if we do not want to cut down on existing functionality, since the pagers are defined outside forgit (either in the git config or in an environment variable). The most straight forward approach to solving this would probably be to introduce `eval` everywhere we use the pager variables, e.g.

```diff
- "$FORGIT" exec_diff "${commits[@]}" -U"$diff_context" -- | $_forgit_diff_pager
+ "$FORGIT" exec_diff "${commits[@]}" -U"$diff_context" -- | eval "$_forgit_diff_pager"
```
However, this solution is ugly in my eyes and I'm concerned it reintroduces the anti-pattern of `eval`/deferred code that we got rid of in #326.
So here is an attempt to solve this issue while keeping the amount of deferred code and usage of `eval` as low as possible.

The code here is not final by any means and not tested thoroughly. This is a draft to start a discussion on how we want to solve this issue. Would highly appreciate your thoughts on this @wfxr, @cjappl & @carlfriedrich.

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [X] bash
    - [X] zsh
    - [ ] fish
- OS
    - [ ] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -127,12 +127,53 @@ _forgit_quote_files() {
     echo "${files[*]}"
 }
 
-_forgit_pager=${FORGIT_PAGER:-$(git config core.pager || echo 'cat')}
-_forgit_show_pager=${FORGIT_SHOW_PAGER:-$(git config pager.show || echo "$_forgit_pager")}
-_forgit_diff_pager=${FORGIT_DIFF_PAGER:-$(git config pager.diff || echo "$_forgit_pager")}
-_forgit_ignore_pager=${FORGIT_IGNORE_PAGER:-$(hash bat &>/dev/null && echo 'bat -l gitignore --color=always' || echo 'cat')}
-_forgit_blame_pager=${FORGIT_BLAME_PAGER:-$(git config pager.blame || echo "$_forgit_pager")}
-_forgit_enter_pager=${FORGIT_ENTER_PAGER:-"LESS='-r' less"}
+_forgit_page() {
+    local pager
+    pager=${FORGIT_PAGER:-$(git config core.pager)}
+    if [[ -n $pager ]]; then
+        eval "$pager"
+    else
+        cat
+    fi
+}
+
+_forgit_diff_page() {
+    local pager
+    pager=${FORGIT_DIFF_PAGER:-$(git config pager.diff)}
+    if [[ -n $pager ]]; then
+        eval "$pager"
+    else
+        _forgit_page
+    fi
+}
+
+_forgit_show_page() {
+    local pager
+    pager=${FORGIT_SHOW_PAGER:-$(git config pager.show)}
+    if [[ -n $pager ]]; then
+        eval "$pager"
+    else
+        _forgit_page
+    fi
+}
+
+_forgit_blame_page() {
+    local pager
+    pager=${FORGIT_BLAME_PAGER:-$(git config pager.blame)}
+    if [[ -n $pager ]]; then
+        eval "$pager"
+    else
+        _forgit_page
+    fi
+}
+
+_forgit_enter_page() {
+    if [[ -n $FORGIT_ENTER_PAGER ]]; then
+        eval "$FORGIT_ENTER_PAGER"
+    else
+        LESS='-r' less
+    fi
+}
 
 _forgit_log_graph_enable=${FORGIT_LOG_GRAPH_ENABLE:-"true"}
 _forgit_log_format=${FORGIT_LOG_FORMAT:-%C(auto)%h%d %s %C(black)%C(bold)%cr%Creset}
@@ -148,7 +189,7 @@ _forgit_log_preview() {
     local sha
     sha=$(echo "$1" | _forgit_extract_sha)
     shift
-    echo "$sha" | xargs -I% git show --color=always -U"$_forgit_preview_context" % -- "$@" | $_forgit_show_pager
+    echo "$sha" | xargs -I% git show --color=always -U"$_forgit_preview_context" % -- "$@" | _forgit_show_page
 }
 
 _forgit_log_enter() {
@@ -223,7 +264,7 @@ _forgit_diff_view() {
         IFS=" " read -r -a commits <<< "${*:3}"
     fi
     echo "$input_line" | _forgit_get_files_from_diff_line | xargs -0 \
-        "$FORGIT" exec_diff "${commits[@]}" -U"$diff_context" -- | $_forgit_diff_pager
+        "$FORGIT" exec_diff "${commits[@]}" -U"$diff_context" -- | _forgit_diff_page
 }
 
 _forgit_edit_diffed_file() {
@@ -264,7 +305,7 @@ _forgit_diff() {
     done
     opts="
         $FORGIT_FZF_DEFAULT_OPTS
-        +m -0 --bind=\"enter:execute($FORGIT diff_enter {} $escaped_commits | $_forgit_enter_pager)\"
+        +m -0 --bind=\"enter:execute($FORGIT diff_enter {} $escaped_commits | $FORGIT enter_page)\"
         --preview=\"$FORGIT diff_view {} '$_forgit_preview_context' $escaped_commits\"
         --bind=\"alt-e:execute-silent($FORGIT edit_diffed_file {})+refresh-preview\"
         $FORGIT_DIFF_FZF_OPTS
@@ -285,9 +326,9 @@ _forgit_diff() {
 _forgit_add_preview() {
     file=$(echo "$1" | _forgit_get_single_file_from_add_line)
     if (git status -s -- "$file" | grep '^??') &>/dev/null; then  # diff with /dev/null for untracked files
-        git diff --color=always --no-index -- /dev/null "$file" | $_forgit_diff_pager | sed '2 s/added:/untracked:/'
+        git diff --color=always --no-index -- /dev/null "$file" | _forgit_diff_page | sed '2 s/added:/untracked:/'
     else
-        git diff --color=always -- "$file" | $_forgit_diff_pager
+        git diff --color=always -- "$file" | _forgit_diff_page
     fi
 }
 
@@ -339,7 +380,7 @@ _forgit_add() {
 
 _forgit_reset_head_preview() {
     file=$1
-    git diff --staged --color=always -- "$file" | $_forgit_diff_pager
+    git diff --staged --color=always -- "$file" | _forgit_diff_page
 }
 
 _forgit_git_reset_head() {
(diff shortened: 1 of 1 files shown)


bin/git-forgit
```diff
@@ -264,7 +305,7 @@ _forgit_diff() {
     done
     opts="
         $FORGIT_FZF_DEFAULT_OPTS
-        +m -0 --bind=\"enter:execute($FORGIT diff_enter {} $escaped_commits | $_forgit_enter_pager)\"
+        +m -0 --bind=\"enter:execute($FORGIT diff_enter {} $escaped_commits | $FORGIT enter_page)\"
```
cjappl, line 308: Typo, I think `_forgit_enter_pag` matches the other changes you made here

bin/git-forgit
```diff
@@ -867,7 +908,19 @@ export FORGIT_GI_REPO_LOCAL="${FORGIT_GI_REPO_LOCAL:-${XDG_CACHE_HOME:-$HOME/.ca
...
+        escaped_files=()
+        for file in "${files[@]}"; do
+            escaped_files+=("'$file'")
+        done
+        eval "$FORGIT_IGNORE_PAGER ${escaped_files[*]}" 2> /dev/null
+    elif hash bat &> /dev/null; then
+        bat -l gitignore --color=always "${files[@]}" 2>/dev/null
+    else
+        cat "${files[@]}" 2> /dev/null
+    fi
```
cjappl, line 923: just checking this change was intentional - more of a change than the rest of the files.

cjappl (commented):
Remind me again what the initial goal of striking `eval` from the program was?

I think this approach is fine, especially if we can say that we didn't "give up" that initial goal by re-introducing it here. Code seems good and encapsulated, and I'm generally for this change 👍 (interested to hear what others think)

sandr01d:
@cjappl 
> Remind me again what the initial goal of striking eval from the program was?

`eval` takes deferred code as a string and interprets it as a commandline while applying all the nasty word splitting, globbing etc. that tends to introduce all kinds of bugs. While doing so, `eval` effectively removes the outer quotes, leaving variables unquoted when you're not careful. Working around this lead us to introduce all the weird "double escaping", like we did in #323. To make things worse, the deferred code snippets do not get checked by shellcheck, making it even harder to find issues. `eval` "blindly" executes any string you pass it - combine that with untrusted input (in our case that could be a repository with a specially crafted commit message) and you gain a huge attack surface for code injection.
To sum things up, our effort wasn't about `eval` specifically but about deferred code in general - `eval` is just the most common way to execute deferred code.

bin/git-forgit
sandr01d: I think this is actually correct this way. I've added `enter_page` to our `private_commands` array, so executing `$FORGIT enter_page` will execute the main script with the known `enter_page` argument, which in turn leads to `_forgit_enter_page` being executed. We can not call `_forgit_enter_page` directly because fzf executes the code snippet in question on a subshell. We do the same with our preview functions.

bin/git-forgit
sandr01d: Intentional, yes but not final. I will try to slim it down in case we decide to go this route. The difference with this one is that we end up with two paths in the `$files` array due to the `{,.gitignore}` expansion. When using `eval`, both should be quoted separately (hence the extra for loop). Also we have two fallbacks here (bat and cat) that we do not use anywhere else.

bin/git-forgit
```diff
@@ -127,12 +127,53 @@ _forgit_quote_files() {
...
+}
+
+_forgit_enter_page() {
+    if [[ -n $FORGIT_ENTER_PAGER ]]; then
+        eval "$FORGIT_ENTER_PAGER"
+    else
+        LESS='-r' less
+    fi
+}
```
wfxr, line 176: It seems like these changes don't actually reduce the risks posed by eval?

If my understanding is correct, the potential risk should come from the output of `git config ...`, not the fallback string constants, such as `cat` or `LESS='-r' less`, etc.

wfxr:
@sandr01d Thank you for your outstanding contributions.

I see your points, and agree with most of them. My only concern is, for a terminal utils like forgit, do we need to pursue extreme security to the point of giving up the simplicity and efficiency of the code? And can we actually do it? `eval` is present in every corner of fzf, including environment variables, various `--bind` parameters, etc. we will inevitably use `eval` as long as we are using fzf. I think one of the main reasons why fzf has been so successful is also because of the flexibility brought by the `eval` mechanism. So maybe we don't need to consider security more important than fzf since security depends on the weakest link in the system. But I'm still 100% in favor of avoiding using `eval` as long as there are good alternative implementations.

sandr01d:
@wfxr

> My only concern is, for a terminal utils like forgit, do we need to pursue extreme security to the point of giving up the simplicity and efficiency of the code?

My comment regarding security was meant as one of the reasons why we've removed `eval` in #326 and not related to this PR. I do not think the changes we've made in #326 have made the code more complex or less efficient. In fact, I personally find it easier to work with the code base since we've made these changes. Does your concern regard the changes in #326 or the changes proposed here?
In this particular case `eval` is obviously unavoidable, since we need to execute code that is stored as a string outside of forgit. Since these strings are defined in the users environment (git config or environment variables), I wouldn't consider them as coming from an untrusted source, so this should be fine.
Do I understand correctly, that you would prefer not go in the direction this PR suggests? If so, what would be your preferred way to solve #318? Simply add `eval` before any usage of the pager variables as outlined in the description or do you have something different in mind? Would be interested in reading your thoughts on the matter.

bin/git-forgit
sandr01d: They do not and they don't intend to do. This was an attempt to solve #318 without spreading the usage of `eval` throughout the code base while using as little deferred code as possible. Not having the fallbacks as deferred code mainly was an attempt to improve readability.

wfxr:
@sandr01d 
> Does your concern regard the changes in https://github.com/wfxr/forgit/pull/326 or the changes proposed here?

No, I'm referring to the changes made in this PR. #326 looks good to me. As you said, solving this problem is difficult to bypass `eval`, even if we rewrite these pagers into functions. I personally think it is more concise to use `eval` in the place where the pager need to be lazy evaluated, if that's necessary.

When I talk about efficiency, I don’t just mean this part of the function, but also the loading speed of the git-forgit script. After all, we need to completely parse this script every time we use a forgit function. I'm worried that its continued expansion will make users experience more obvious delays. Of course, just this one commit will not immediately cause a problem, but we should still be aware of it.

Another point I think of is every forgit command will call the `git-forgit` script, which should have already achieved globally lazy evaluation. Do we need to lazy evaluate each pager again inside `git-forgit`?

carlfriedrich:
Just adding my two cents here: I am more conservative in this case. I like the previous implementation better, since it is really easy to read. The change adds complexitiy to the code that is not needed IMO.

What about calling `eval` directly when assigning the `_forgit_*_pager` variables? Then at least we have them all in one place instead of scattered all over the file.

sandr01d:
> What about calling eval directly when assigning the _forgit_*_pager variables? Then at least we have them all in one place instead of scattered all over the file.

That sounds like a good approach to me @carlfriedrich, not sure exactly what it would look like though. Would you be open to create a PR for this? In this case we could close this one.

wfxr:
How about moving all the pager logic into another script named like `forgit-pager`?

```bash
declare -A PAGERS

PAGERS["core"]=${FORGIT_PAGER:-$(git config core.pager || echo cat)}
PAGERS["show"]=${FORGIT_SHOW_PAGER:-$(git config pager.show || echo "${PAGERS[core]}")}
PAGERS["diff"]=${FORGIT_DIFF_PAGER:-$(git config pager.diff || echo "${PAGERS[core]}")}
PAGERS["blame"]=${FORGIT_BLAME_PAGER:-$(git config pager.blame || echo "${PAGERS[core]}")}
PAGERS["ignore"]=${FORGIT_IGNORE_PAGER:-$(hash bat &>/dev/null && echo 'bat -l gitignore --color=always' || echo cat)}
PAGERS["enter"]=${FORGIT_ENTER_PAGER:-"LESS='-r' less"}

pager="${1:core}"

if [[ -z "${PAGERS[$pager]}" ]]; then
  echo "pager not found: $pager" >&2
  exit 1
fi

eval "${PAGERS[$pager]}"
```

Then we use `forgit-pager` in `git-forgit` like this:

```diff
++FORGIT_PAGER="${FORGIT%/*}/forgit-pager"
...
--preview_cmd="echo {} | ... | $_forgit_show_pager"
++preview_cmd="echo {} | ... | $FORGIT_PAGER show"
...
```

carlfriedrich:
@wfxr I like that approach, we would even have only one single `eval` there. I do not see any advantage of putting that into a dedicated script, though, and would prefer a function.

wfxr:
> @wfxr I like that approach, we would even have only one single `eval` there. I do not see any advantage of putting that into a dedicated script, though, and would prefer a function.

@carlfriedrich Yes, it can be implemented in the existing `git-forgit` script. Perhaps one advantage of creating a dedicate script is that it can reduce some lines in `git-forgit`. But it's not that important, both ways are OK for me.

cjappl:
I would prefer keeping this in a single script, if possible. I think it simplifies things especially with regards to our packaging and making sure everything works when installed.

Not the end of the world if we split it though, just my 2cents!

carlfriedrich:
Great, @sandr01d can you pick @wfxr 's code from above and put it in a function?

sandr01d:
I definitively like @wfxr's approach. Unfortunately associative arrays are only available in bash 4.0+, so the code won't work on macOS as is. I've pushed an approach that follows the same idea without using associative arrays. Also had to adjust it a bit to make it work with variables being passed along to the ignore pager. WDYT?

wfxr (approved):
LGTM！

cjappl (approved):

bin/git-forgit
```diff
@@ -127,19 +127,31 @@ _forgit_quote_files() {
...
 
+_forgit_pager() {
+    local pager
+    pager=$(_forgit_get_pager "$1")
+    [[ -z "${pager}" ]] && exit 1
+    eval "${pager} ${*:2}"
+}
+
+_forgit_get_pager() {
+    case "${1:-core}" in
```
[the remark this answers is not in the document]
sandr01d: Sure, updated accordingly. Let me know if it works for you like this.

carlfriedrich (approved):
Great, thanks a lot!

PR #383, opened 2024-04-09
Add a preview to _forgit_clean

sandr01d:
## Check list

## Description

`gclean` is the only command that does not display a preview. I sometimes find myself exiting `gclean` just to check the content of a file before deleting it. Being able to have a last peek at a file before removing it does sound like a good idea in general to me.

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [X] bash
    - [X] zsh
    - [ ] fish
- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -471,9 +471,17 @@ _forgit_stash_push() {
 }
 
 _forgit_clean_preview() {
-    local file
-    file=$1
-    git diff --color=always /dev/null "$file" | _forgit_pager diff
+    local path
+    path=$1
+    if [[ -d "$path" ]]; then
+        if hash tree &> /dev/null; then
+            tree "$path"
+        else
+            find "$path"
+        fi
+    else
+        git diff --color=always /dev/null "$path" | _forgit_pager diff
+    fi
 }
 
 # git clean selector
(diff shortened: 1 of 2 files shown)


cjappl (approved):

wfxr (approved):

carlfriedrich (approved):
Good idea, works for me!

sandr01d:
FYI: I've added a preview for directories now as well (before this you'd see an unwanted error message when trying to preview a directory). Looks like this when `tree` is installed

image

Otherwise like this:

image

Feel free to give this another review if you like. Will keep testing this for a few days.

bin/git-forgit
```diff
@@ -470,6 +470,20 @@ _forgit_stash_push() {
     _forgit_git_stash_push ${msg:+-m "$msg"} -u "${files[@]}"
 }
 
+_forgit_clean_preview() {
+    local path
+    path=$1
+    if [[ -d "$path" ]]; then
+        if hash tree &> /dev/null; then
+            tree "$path"
```
cjappl, line 478: Should this binary be set as a variable one can control? similar to PAGER or DELTA etc?

Are there other places in the code we care about looking at directories with a tool like this? (if the answer is either or both, we could consider accepting this change as is, then making it configurable/expanding it's use in a subsequent review)

sandr01d: > Should this binary be set as a variable one can control? similar to PAGER or DELTA etc?

Good point, that does sound reasonable to me. I can see a use case here of users wanting to override the preview with something fancier and/or colorful.

> Are there other places in the code we care about looking at directories with a tool like this?

Currently no, but maybe in the future.

I've updated this PR accordingly. I think the change is small enough to have it included in this PR.

cjappl (approved):

PR #386, opened 2024-04-19
Fix: gsp does not work outside the root directory

sandr01d:
## Check list

- [X] I have performed a self-review of my code
- [ ] I have commented my code in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation

## Description

`gsp` only works when executing it from the repository's root directory. This commit fixes this by passing the root directory to the `git ls-files` command we use to determine the files that can be stashed, similarly to how we do it in \_forgit\_reset\_head.

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [X] bash
    - [X] zsh
    - [ ] fish
- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

PR #388, opened 2024-04-19
Fix special characters in file names

### LINKED ISSUE #325: Issues with special characters in filenames
## Check list

## Environment info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish

    git clone https://github.com/wfxr/forgit.git
    cd forgit
    echo "hello world" > my_filé

Git add with `ga` does not work

    fatal: pathspec 'my_fil\303\251' did not match any files
    Nothing to add.

Add the file manually and commit:

    git add my_filé
    git commit -m "add file"

Try to view log with `glo`, it works. Then press "enter" in this commit, it show empty screen for the file "my_fil\303\251".

sandr01d:
## Check list

## Description

Git escapes special characters in it's output when `core.quotePath` is true or unset. Git always expects unquoted file paths as input. This leads to issues when we consume output from git and use it to build input for other git commands. This commit ensures we always feed unqoted paths to git commands. The `_forgit_list_files` function is introduced to handle usage of git ls-files with the -z flag, which ensures unquoted paths. It replaces the direct calls to git ls-files in `_forgit_reset_head`, `_forgit_stash_push` and `_forgit_checkout_file`.
In `_forgit_add` the current approach of using colors to separate unstaged from staged files is replaced with a call to `_forgit_ls_files` with the appropriate flags to only list unstaged files.
In `_git_reset_head` the -z option is added to the git diff command to ensure unqoted paths.
Since `git clean` does not support the -z flag, we disable `core.quotePath` by passing a configuration parameter in `_forgit_clean`.
This fixes #325

## Note

This PR is based on #386. I will request reviews once #386 has been merged.

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [X] bash
    - [X] zsh
    - [ ] fish
- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -171,7 +171,7 @@ _forgit_list_files() {
     # merge conflict.
     # With the -z flag, git also uses \0 line termination, so we 
     # have to replace the terminators.
-    git ls-files -z "$@" "$rootdir" | uniq | tr '\0' '\n'
+    git ls-files -z "$@" "$rootdir" | tr '\0' '\n' | uniq
 }
 
 _forgit_log_preview() {
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -344,10 +344,13 @@ _forgit_edit_add_file() {
 # git add selector
 _forgit_add() {
     _forgit_inside_work_tree || return 1
-    local files opts
+    local changed unmerged untracked files opts
     # Add files if passed as arguments
     [[ $# -ne 0 ]] && { _forgit_git_add "$@" && git status -su; return $?; }
 
+    changed=$(git config --get-color color.status.changed red)
+    unmerged=$(git config --get-color color.status.unmerged red)
+    untracked=$(git config --get-color color.status.untracked red)
     opts="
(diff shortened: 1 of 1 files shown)


wfxr (approved):
LGTM

cjappl (approved):

sandr01d:
I found out that my changes introduced a bug where `ga` would no longer display the status indicator for the files (e.g., [M] for a modified file). This also lead to an issue where a file could not be added if it matched the regex in the `sed` command and therefore brackets got added around part of the path. I mainly reverted changes in `_forgit_add` back to the state of the master branch, but feel free to rereview this if you like. I'll give this a few days of extra testing to see if anything else comes up, so please don't merge yet.

PR #389, opened 2024-04-21
Add FORGIT_STASH_SHOW_PREVIEW_GIT_OPTS

ccoVeille:
## Check list

## Description

FORGIT_STASH_SHOW_PREVIEW_GIT_OPTS variable can be used to tune the rendering of a stash preview.

The following example allows listing the content of the stash before opening it.

FORGIT_STASH_SHOW_PREVIEW_GIT_OPTS="--patch-with-stat --stat-count=10"

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [ ] zsh
    - [X] fish
- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,10 +163,18 @@ For linux users `FORGIT_COPY_CMD` should be set to make copy work. Example: `FOR
 Options can be set via environment variables. They have to be **exported** in
 order to be recognized by `forgit`.
 
-For instance, if you want to order branches in `gcb` by the last committed date you could:
+For instance:
+
+- if you want to order branches in `gcb` by the last committed date, you could:
 
 ```shell
 export FORGIT_CHECKOUT_BRANCH_BRANCH_GIT_OPTS='--sort=-committerdate'
+```
+
+- if you want to see a preview of the stashed items in `gss` preview, you could:
+
+```shell
+export FORGIT_STASH_SHOW_PREVIEW_GIT_OPTS="--patch-with-stat --stat-count=10"
 ```
 
 ## shell aliases
@@ -243,7 +251,7 @@ These are passed to the according `git` calls.
 | `gct`    | `FORGIT_CHECKOUT_TAG_GIT_OPTS`                                              |
 | `gco`    | `FORGIT_CHECKOUT_COMMIT_GIT_OPTS`                                           |
 | `grc`    | `FORGIT_REVERT_COMMIT_GIT_OPTS`                                             |
-| `gss`    | `FORGIT_STASH_SHOW_GIT_OPTS`                                                |
+| `gss`    | `FORGIT_STASH_SHOW_GIT_OPTS`, `FORGIT_STASH_SHOW_PREVIEW_GIT_OPTS`          |
 | `gsp`    | `FORGIT_STASH_PUSH_GIT_OPTS`                                                |
 | `gclean` | `FORGIT_CLEAN_GIT_OPTS`                                                     |
 | `grb`    | `FORGIT_REBASE_GIT_OPTS`                                                    |
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -395,7 +395,9 @@ _forgit_reset_head() {
 _forgit_stash_show_preview() {
     local stash
     stash=$(echo "$1" | cut -d: -f1)
-    _forgit_git_stash_show "$stash" | _forgit_pager diff
+    _forgit_stash_show_preview_git_opts=()
+    _forgit_parse_array _forgit_stash_show_preview_git_opts "$FORGIT_STASH_SHOW_PREVIEW_GIT_OPTS"
+    _forgit_git_stash_show "${_forgit_stash_show_preview_git_opts[@]}" -- "$stash" | _forgit_pager diff
 }
 
 _forgit_git_stash_show() {


README.md
```diff
@@ -156,7 +156,7 @@ Install `forgit` in just one click.
 | <kbd>Alt</kbd> - <kbd>E</kbd>                 | Open file in default editor (when possible) |
 
 \* Available when the selection contains a commit hash or a stash ID.
-For linux users `FORGIT_COPY_CMD` should be set to make copy work. Example: `FORGIT_COPY_CMD='xclip -selection clipboard'`.
+For Linux users `FORGIT_COPY_CMD` should be set to make copy work. Example: `FORGIT_COPY_CMD='xclip -selection clipboard'`.
```
sandr01d, line 159: I think this change would better go in a follow up PR.

README.md
```diff
@@ -339,7 +339,7 @@ export FORGIT_LOG_FZF_OPTS='
 
 # 📦 Optional dependencies
 
-- [`delta`](https://github.com/dandavison/delta) / [`diff-so-fancy`](https://github.com/so-fancy/diff-so-fancy): For better human readable diffs.
+- [`delta`](https://github.com/dandavison/delta) / [`diff-so-fancy`](https://github.com/so-fancy/diff-so-fancy): For better human-readable diffs.
```
sandr01d, line 342: Same here. Would prefer to have this in a follow up PR with the change I've commented on above.

sandr01d (changes requested):
Thanks for your contribution @ccoVeille! Just some minor changes to do here. The code itself works fine for me and I find the ability of having the diff stat on top of the preview very neat. The example you provided will go straight to my dotfiles once this is merged :wink:

README.md
ccoVeille: I removed the changes from this PR and opened #390

ccoVeille:
> Thanks for your contribution @ccoVeille! Just some minor changes to do here. The code itself works fine for me and I find the ability of having the diff stat on top of the preview very neat. The example you provided will go straight to my dotfiles once this is merged 😉

Do you think, l I should add it to the documentation the same way FORGIT_CHECKOUT_BRANCH_BRANCH_GIT_OPTS is explained ?

FORGIT_STASH_SHOW_PREVIEW_GIT_OPTS="--patch-with-stat --stat-count=10"

sandr01d:
> Do you think, l I should add it to the documentation the same way FORGIT_CHECKOUT_BRANCH_BRANCH_GIT_OPTS is explained ?

Yes, that sounds useful :+1:

ccoVeille:
Please relaunch the CI as it looks like a false positive due to `apt`

carlfriedrich:
I would suggest to wait for #391 before merging this (see my comment there).

carlfriedrich:
@sandr01d @ccoVeille For your information (because that seems to be similar to what you want to achieve here): you can override preview functions by defining your own and then include it in the according `FORGIT_*_FZF_OPTS` variable.

I am using this by defining a [custom preview function for forgit logs](https://github.com/carlfriedrich/dotfiles/blob/main/bash_functions.d/forgit.sh#L22-L35) and including that in my [`FORGIT_STASH_FZF_OPTS` variable](https://github.com/carlfriedrich/dotfiles/blob/main/bash_functions.d/forgit.sh#L92). It includes the diff stat in a dimmed representation, see a screenshot here:

grafik

This works already without introducing new variables, so might this eventually be an alternative way to go for you?

ccoVeille:
My issue is that I cannot achieve what I want with the _FZF variables. These are for fzf, BTW.

I don't want to affect the git command that is launched of the left, but the git command that are displayed on the right, the preview.

Right now, there is no way to do it.

ccoVeille:
Oh you are referring to a code that is not merged yet,but present in #391

I'll have a look. I let maintainer choose the best option. I might comment your PR

ccoVeille:
> @sandr01d @ccoVeille For your information (because that seems to be similar to what you want to achieve here): you can override preview functions by defining your own and then include it in the according `FORGIT_*_FZF_OPTS` variable.
> 
> I am using this by defining a [custom preview function for forgit logs](https://github.com/carlfriedrich/dotfiles/blob/main/bash_functions.d/forgit.sh#L22-L35) and including that in my [`FORGIT_STASH_FZF_OPTS` variable](https://github.com/carlfriedrich/dotfiles/blob/main/bash_functions.d/forgit.sh#L92). It includes the diff stat in a dimmed representation, see a screenshot here:
> 
> grafik
> 
> This works already without introducing new variables, so might this eventually be an alternative way to go for you?

I find your solution too complex. I mean you reimplemented somehow the internals of forgit.

It's OK for an advanced shell user, but for a random dev who want to add a parameter to a command. Your code works, someone could copy paste it, but I don't think they would be able to maintain it.

carlfriedrich:
> I don't want to affect the git command that is launched of the left, but the git command that are displayed on the right, the preview.

@ccoVeille Yes, I got that. Take a look at the links in my previous comment, those point to my dotfiles where I configure forgit similarly to what (I think) you want to achieve here. It works with the current forgit code base.

The point to understand is: instead of *configuring* the preview function, I *replace* it with my own (using fzf's `--preview` option). This gives you full freedom over any forgit preview, without having to add more variables to the forgit code.

carlfriedrich:
@ccoVeille Ah sorry, our comments overlapped. I agree that it's not a simple solution for everyone. However, maybe we could add an examples section to the README for use cases like this? Might be better than introducing more variables for things that already can be configured. Not sure about that myself, yet, though. 

Curious to hear opinions from @wfxr @sandr01d @cjappl.

sandr01d:
Thanks for the insight @carlfriedrich. While the solution you provided is certainly an interesting way to solve this, I think replacing preview functions with injected code goes beyond configuration since it requires familiarity with the inner workings of forgit – at least to a certain extend. I think adding environment variables to handle this would simplify the process of customizing the preview quite a bit (while of course adding complexity to the project). I lean more towards adding such variables if we can manage to reduce boiler plate to a minimum.
I agree that the changes from this PR and those proposed in #391 should go into a single PR.

ccoVeille:
For records, I don't think that I'll be able to work  on it in the next days.

ccoVeille:
> For records, I don't think that I'll be able to work on it in the next days.

For records, I'm still alive, but we got a new kid, so well, I'm pretty busy :smile: for now

ccoVeille:
Replaced by 

- #396

PR #394, opened 2024-05-01
ci: add dependabot update for GitHub actions

boidolr:
## Check list

## Description

Currently there still are actions that need to be updated due to the deprecation of Node.js 16 on runners: https://github.com/wfxr/forgit/actions/runs/8903375722
By adding a dependabot configuration the GitHub actions used by workflows of this repository will stay up to date.
All configuration options are documented on [https://docs.github.com](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file).

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [ ] zsh
    - [ ] fish
- OS
    - [ ] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [x] Others: GitHub Action Runners

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


cjappl:
Can we not just fix this by moving our action to v2? I am personally somewhat opposed to an auto-updater. I don't think it's overly onerous to update things when they need to be updated. "automatic" can mean "breaks unexpectedly and hard to track down"

boidolr:
Dependabot is only automatic in the sense that it will create pull requests with updates. These still have to be merged, so there should be no hidden breakage. Why create the prs manually when it can be done for you?

carlfriedrich:
@boidolr Thanks for your contribution!

Automating dependency PRs is definitely a good idea. I would take care of the automatically created PRs if nobody else wants to do it, so I approve this. If there are no objections from @wfxr or @sandr01d, I am gonna merge this within the next days.

sandr01d:
I'm also fine with that and would be willing to take care of PRs created from it as well.

cjappl:
Sounds good, happy to be overruled. :)

PR #396, opened 2024-07-17
Add *_PREVIEW_GIT_OPTS variables

ccoVeille:
## Check list

- [ ] I have performed a self-review of my code
- [X] I have commented my code in hard-to-understand areas
- [X] I have made corresponding changes to the documentation

## Description

These variables can be used to tune the rendering of the previews

The following example allows listing the content of the stash before
opening it.

FORGIT_STASH_SHOW_PREVIEW_GIT_OPTS="--patch-with-stat --stat-count=10"

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Breaking change
- [X] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [ ] zsh
    - [ ] fish
- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

Fixes #391

PR #402, opened 2024-11-03
feat: add interactive git worktree operations

suft:
## Check list

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Breaking change
- [X] Documentation change

## Test environment

- Shell
    - [X] bash
    - [ ] zsh
    - [X] fish
- OS
    - [ ] Linux
    - [X] Mac OS X
    - [ ] Windows
    - [ ] Others:


carlfriedrich:
@suft Thanks for your contribution! Is this ready for review or did you make it a draft on purpose?

suft:
@carlfriedrich I made it draft on purpose. Still have a few things to adjust.

bin/git-forgit
```diff
@@ -1002,6 +1003,95 @@ _forgit_ignore_clean() {
...
 }
 
+_forgit_worktree_preview() {
+    local sha
+    # trailing space in  grep to avoid matching worktrees with a common path
+    sha=$(git worktree list | grep "$1 " | awk '{print $2}')
+    [[ "$sha" == "(bare)" ]] && return
+    # the trailing '--' ensures that this works for branches that have a name
+    # that is identical to a file
+    git log "$sha" "${_forgit_log_preview_options[@]}" --
```
suft, line 1013: - I should probably add preview options `FORGIT_WORKTREE_PREVIEW_GIT_OPTS` (like #396)
- I've noticed this can run a bit slow if you have a repo with a long history because it attempts to get the entire history for the branch checked out in that worktree
- Would be quicker if I limit the log entries in the worktree preview options

bin/git-forgit
```diff
@@ -1002,6 +1003,95 @@ _forgit_ignore_clean() {
     [[ -d "$FORGIT_GI_REPO_LOCAL" ]] && rm -rf "$FORGIT_GI_REPO_LOCAL"
 }
 
+_forgit_worktree_preview() {
+    local sha
+    # trailing space in  grep to avoid matching worktrees with a common path
+    sha=$(git worktree list | grep "$1 " | awk '{print $2}')
+    [[ "$sha" == "(bare)" ]] && return
```
suft, line 1010: The preview is blank if its the bare `git-dir` (since it has no history)

sandr01d (changes requested):
Thanks for you work @suft! I think this is going to be a nice improvement. There are a few things that need to be adjusted. In some of my comments I've pinged the other maintainers for their opinions. Please hold back on implementing those, as those comments are opinionated and I'd like to have feedback from the others first before deciding in which direction to go.

bin/git-forgit
```diff
@@ -1002,6 +1003,98 @@ _forgit_ignore_clean() {
...
+    git worktree lock "${_forgit_worktree_lock_git_opts[@]}" "$@"
+}
+
+_forgit_worktree_lock() {
+    _forgit_inside_work_tree || _forgit_inside_git_dir || return 1
+    local tree opts
+
+    opts="
+        $FORGIT_FZF_DEFAULT_OPTS
+        +s -m --tiebreak=index
```
sandr01d, line 1050: Multi selection does not seem to work. We'd need to parse the output of `git worktree list` into an array to have proper word splitting. But AFAIK the git worktree subcommands do not support specifying multiple paths anyway, so we should probably simply not allow multi selection. This also affects the other functions.

bin/git-forgit
```diff
@@ -1002,6 +1003,98 @@ _forgit_ignore_clean() {
...
+        --preview=\"$FORGIT worktree_preview {1}\"
+        $FORGIT_WORKTREE_LOCK_FZF_OPTS
+    "
+
+    tree=$(git worktree list | awk '{print $1}' | grep -v "(bare)" | FZF_DEFAULT_OPTS="$opts" fzf)
+    [[ -z "$tree" ]] && return 1
+    _forgit_git_worktree_lock "$tree"
```
sandr01d, line 1057: It would be nice to allow passing in additional arguments as we do with other functions by propagating the function arguments, so that `gwl --reason <STRING>` works as expected.
This also applies to the other functions.

sandr01d (changes requested):

bin/git-forgit
```diff
@@ -1002,6 +1003,98 @@ _forgit_ignore_clean() {
...
+    [[ "$sha" == "(bare)" ]] && return
+    _forgit_worktree_preview_git_opts=()
+    _forgit_parse_array _forgit_worktree_preview_git_opts "$FORGIT_WORKTREE_PREVIEW_GIT_OPTS"
+    # the trailing '--' ensures that this works for branches that have a name
+    # that is identical to a file
+    git log "$sha" "${_forgit_worktree_preview_git_opts[@]}" --
+}
+
+_forgit_worktree_jump() {
+    _forgit_inside_work_tree || _forgit_inside_git_dir || return 1
```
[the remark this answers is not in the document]
suft: When using worktrees with a bare clone. In the following example `~/github/bat.git` is the `git-dir` and the worktrees are sub directories (`one`, ..., `five`)

Screenshot 2024-11-12 at 8 58 37 PM

bin/git-forgit
```diff
@@ -1002,6 +1003,98 @@ _forgit_ignore_clean() {
...
+    opts="
+        $FORGIT_FZF_DEFAULT_OPTS
+        +s +m --tiebreak=index
+        --preview=\"$FORGIT worktree_preview {1}\"
+        $FORGIT_WORKTREE_JUMP_FZF_OPTS
+    "
+
+    tree=$(git worktree list | awk '{print $1}' | FZF_DEFAULT_OPTS="$opts" fzf)
+    [[ -z "$tree" ]] && return 1
+    echo "$tree"
```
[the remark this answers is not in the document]
suft: - Using `cd` directly in the function only changes the current directory in the script child process, not in the terminal process where the script is called
- Any preference between `cd` and `pushd`?

bin/git-forgit
```diff
@@ -1002,6 +1003,98 @@ _forgit_ignore_clean() {
     [[ -d "$FORGIT_GI_REPO_LOCAL" ]] && rm -rf "$FORGIT_GI_REPO_LOCAL"
 }
 
+_forgit_worktree_preview() {
+    local sha
+    # trailing space in  grep to avoid matching worktrees with a common path
+    sha=$(git worktree list | grep "$1 " | awk '{print $2}')
+    # bare git-dir has no history
+    [[ "$sha" == "(bare)" ]] && return
```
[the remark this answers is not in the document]
suft: - Adding a no history message to the preview is a good idea 🤓 
- The case is when you are working with a bare clone like so (and the preview is on the bare `git-dir`)
Screenshot 2024-11-12 at 9 15 15 PM

bin/git-forgit
```diff
@@ -1002,6 +1003,98 @@ _forgit_ignore_clean() {
...
+}
+
+_forgit_git_worktree_lock() {
+    _forgit_worktree_lock_git_opts=()
+    _forgit_parse_array _forgit_worktree_lock_git_opts "$FORGIT_WORKTREE_LOCK_GIT_OPTS"
+    git worktree lock "${_forgit_worktree_lock_git_opts[@]}" "$@"
+}
+
+_forgit_worktree_lock() {
+    _forgit_inside_work_tree || _forgit_inside_git_dir || return 1
```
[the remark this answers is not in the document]
suft: I guess I never thought of that. I wanted to add other commands like `git worktree add` as `gwa` but wasn't sure where to include the fzf selecting since it has several optional parts

```sh
git worktree add [-f] [--detach] [--checkout] [-b <new-branch>] <path> [<branch>]
```

bin/git-forgit
```diff
@@ -1002,6 +1003,98 @@ _forgit_ignore_clean() {
...
+    local tree opts
+
+    opts="
+        $FORGIT_FZF_DEFAULT_OPTS
+        +s -m --tiebreak=index
+        --preview=\"$FORGIT worktree_preview {1}\"
+        $FORGIT_WORKTREE_LOCK_FZF_OPTS
+    "
+
+    tree=$(git worktree list | awk '{print $1}' | grep -v "(bare)" | FZF_DEFAULT_OPTS="$opts" fzf)
```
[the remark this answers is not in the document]
suft: When using worktrees with a bare clone, you automatically get the `git-dir` as an entry in the
`git worktree list` output (even though its not really a worktree).
So for certain operations we want to filter it out (like locking, unlocking, removing, ...).
Good catch on filtering already locked (totally slipped my mind).

bin/git-forgit
```diff
@@ -1002,6 +1003,98 @@ _forgit_ignore_clean() {
...
+    [[ $count -eq 1 ]] && return 1
+
+    opts="
+        $FORGIT_FZF_DEFAULT_OPTS
+        +s +m --tiebreak=index
+        --preview=\"$FORGIT worktree_preview {1}\"
+        $FORGIT_WORKTREE_JUMP_FZF_OPTS
+    "
+
+    tree=$(git worktree list | awk '{print $1}' | FZF_DEFAULT_OPTS="$opts" fzf)
```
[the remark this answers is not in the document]
suft: Another case that is harder to catch:

1. Create a new worktree: `git worktree add test`
2. Lock the worktree: `git worktree lock test`
3. Delete the directory: `rm -r test`
4. List worktrees: `git worktree list` (`test` will appear **locked** but not **prunable**)
5. Try to jump to the test worktree with `gwj`

bin/git-forgit
```diff
@@ -1002,6 +1003,98 @@ _forgit_ignore_clean() {
...
+    opts="
+        $FORGIT_FZF_DEFAULT_OPTS
+        +s -m --tiebreak=index --header-lines=1
+        --preview=\"$FORGIT worktree_preview {1}\"
+        $FORGIT_WORKTREE_UNLOCK_FZF_OPTS
+    "
+
+    tree=$(git worktree list | awk '{print $1}' | grep -v "(bare)" | FZF_DEFAULT_OPTS="$opts" fzf)
+    [[ -z "$tree" ]] && return 1
+    git worktree unlock "$tree"
```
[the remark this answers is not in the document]
suft: `git worktree unlock <worktree>` doesn't take any options. https://git-scm.com/docs/git-worktree

bin/git-forgit
```diff
@@ -1002,6 +1003,98 @@ _forgit_ignore_clean() {
...
+    local tree opts
+
+    opts="
+        $FORGIT_FZF_DEFAULT_OPTS
+        +s -m --tiebreak=index --header-lines=1
+        --preview=\"$FORGIT worktree_preview {1}\"
+        $FORGIT_WORKTREE_UNLOCK_FZF_OPTS
+    "
+
+    tree=$(git worktree list | awk '{print $1}' | grep -v "(bare)" | FZF_DEFAULT_OPTS="$opts" fzf)
```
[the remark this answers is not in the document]
suft: Locked worktrees can be removed with `-f -f` options
Screenshot 2024-11-13 at 7 25 08 PM

bin/git-forgit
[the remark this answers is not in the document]
carlfriedrich: This is tricky, because the actual change of the directory is currently implemented in the shell plugin. Some users (e.g. me) do not use forgit with the plugin but in the form of a subcommand of `git`. Calling `git forgit worktree_jump` won't work at the moment, though. IMO we should find a way to make this work correctly in both usecases, or otherwise rename the function (because right now it actually does not jump).

bin/git-forgit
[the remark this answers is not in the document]
sandr01d: Understood, thanks!

bin/git-forgit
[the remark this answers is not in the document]
sandr01d: > I wanted to add other commands like git worktree add as gwa but wasn't sure where to include the fzf selecting since it has several optional parts

Not all git command fit forgits ergonomics, I think the selection you made is sensible. We can always add other commands later on if we come up with something.

bin/git-forgit
```diff
@@ -1002,6 +1003,98 @@ _forgit_ignore_clean() {
...
+    git worktree remove "${_forgit_worktree_remove_git_opts[@]}" "$@"
+}
+
+_forgit_worktree_remove() {
+    _forgit_inside_work_tree || _forgit_inside_git_dir || return 1
+    local tree opts
+
+    opts="
+        $FORGIT_FZF_DEFAULT_OPTS
+        +s -m --tiebreak=index --header-lines=1
```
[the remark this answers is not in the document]
sandr01d: > This is the way we do it with all other forgit functions as well

 Weird, I've never noticed. In this case I'm fine with keeping it in this PR as well.

sandr01d (commented):
Thanks for the changes @suft, looks good to me so far. Let me summarize the things that still need to be addressed:

* The default preview should look the same as for other commands that use `git log` (https://github.com/wfxr/forgit/pull/402#discussion_r1838114506)
* `_forgit_worktree_jump` does not change the directory when forgit is used as a git subcommand (https://github.com/wfxr/forgit/pull/402#discussion_r1843629352)
* The root worktree should not be selectable with `gwl`. We can make it non-interactable with `--header-lines=1` and also bring this back in the other functions that you used this with. Don't mind my earlier comment regarding this. (https://github.com/wfxr/forgit/pull/402#discussion_r1838138545)
* We should add completions for the new commands (https://github.com/wfxr/forgit/pull/402#discussion_r1838138545)
* We should allow passing additional arguments to the underlying git commands (https://github.com/wfxr/forgit/pull/402#discussion_r1838195565). The logic for detecting whether the git command should be executed immediately without using fzf needs to be adjusted to do so. Instead of testing if any arguments were passed, we need to test whether non flag arguments were passed. You can use `_forgit_contains_non_flags` to do so.

bin/git-forgit
```diff
@@ -1002,6 +1003,130 @@ _forgit_ignore_clean() {
...
+    if [[ $# -ne 0 ]]; then
+        git worktree unlock "$@"
+        worktree_unlock_status=$?
+        return $worktree_unlock_status
+    fi
+    local worktree_list tree opts
+
+    worktree_list=$(git worktree list | grep -v "(bare)" | grep -E "locked$")
+    count=$(echo "$worktree_list" | wc -l)
+    [[ $count -eq 1 ]] && return 1
```
sandr01d, line 1116:
```suggestion
    [[ $count -eq 0 ]] && return 1
```
`$count` being 1 is actually fine here, because we filter to only include locked workspaces. Without this change it's impossible to unlock a workspace when there is only one locked worspace.

cjappl:
(it looks like there is very thorough reviewing going on here, please ping me if you need another pair of eyes, otherwise I completely defer to @sandr01d and @carlfriedrich )

bin/git-forgit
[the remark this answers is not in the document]
suft: We could rename the function to `worktree_select` - `gws`
(but also offer the `gwj` alias which combines `cd` and `worktree_select`)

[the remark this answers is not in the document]
sandr01d: Given that there is no way to manipulate the directory of the parent process, I think this is the best compromise we can achieve. @carlfriedrich WDYT?

suft:
> Thanks for the changes @suft, looks good to me so far. Let me summarize the things that still need to be addressed:
> 
> * The default preview should look the same as for other commands that use `git log` ([feat: add interactive git worktree operations #402 (comment)](https://github.com/wfxr/forgit/pull/402#discussion_r1838114506))
> * `_forgit_worktree_jump` does not change the directory when forgit is used as a git subcommand ([feat: add interactive git worktree operations #402 (comment)](https://github.com/wfxr/forgit/pull/402#discussion_r1843629352))
> * The root worktree should not be selectable with `gwl`. We can make it non-interactable with `--header-lines=1` and also bring this back in the other functions that you used this with. Don't mind my earlier comment regarding this. ([feat: add interactive git worktree operations #402 (comment)](https://github.com/wfxr/forgit/pull/402#discussion_r1838138545))
> * We should add completions for the new commands ([feat: add interactive git worktree operations #402 (comment)](https://github.com/wfxr/forgit/pull/402#discussion_r1838138545))
> * We should allow passing additional arguments to the underlying git commands ([feat: add interactive git worktree operations #402 (comment)](https://github.com/wfxr/forgit/pull/402#discussion_r1838195565)). The logic for detecting whether the git command should be executed immediately without using fzf needs to be adjusted to do so. Instead of testing if any arguments were passed, we need to test whether non flag arguments were passed. You can use `_forgit_contains_non_flags` to do so.

The root worktree doesn't show up as the first selection, so `--header-lines=1` will affect a different worktree (which I'm guessing is not the behaviour we want), unless we reverse the fzf item list.

Screenshot 2024-12-18 at 3 02 14 PM

sandr01d (changes requested):
Thanks again for your work @suft. There are a few more things that need to be addressed, but I think we're heading in the right direction :slightly_smiling_face: I think there might have been a misunderstanding regarding argument parsing, sorry if I didn't make it clear enough what I meant previously. I provided an example this time. In case any questions come up, please don't hesitate to ping me.

sandr01d:
> The root worktree doesn't show up as the first selection, so `--header-lines=1` will affect a different worktree (which I'm guessing is not the behaviour we want), unless we reverse the fzf item list.
> Screenshot 2024-12-18 at 3 02 14 PM

I'm not exactly sure what you mean. `git worktree list` always seems to give me the root worktree as the first line and `git worktree list | fzf --header-lines 1` does not allow me to select the root worktree as I'd expect. Could you please provide an example on how to reproduce this?

carlfriedrich:
@suft @sandr01d Are you planning to finish this PR to get it merged anytime?

suft:
> @suft @sandr01d Are you planning to finish this PR to get it merged anytime?

Sorry for the delay — things have been a bit hectic on my end. I’m hoping to take another look at it sometime this week, but totally understand if it makes sense to close the PR in the meantime.

completions/git-forgit.fish
```diff
@@ -42,6 +42,10 @@ complete -c git-forgit -n __fish_forgit_needs_subcommand -a reset_head -d 'git r
 complete -c git-forgit -n __fish_forgit_needs_subcommand -a revert_commit -d 'git revert commit selector'
 complete -c git-forgit -n __fish_forgit_needs_subcommand -a stash_show -d 'git stash viewer'
 complete -c git-forgit -n __fish_forgit_needs_subcommand -a stash_push -d 'git stash push selector'
+complete -c git-forgit -n __fish_forgit_needs_subcommand -a worktree_select -d 'git worktree selector'
```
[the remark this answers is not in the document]
suft: Screenshot 2025-06-29 at 10 17 31 PM

bin/git-forgit
```diff
@@ -1002,6 +1003,162 @@ _forgit_ignore_clean() {
...
+    "
+
+    tree=$(git worktree list | grep -vE "\(bare\)|locked" | awk '{print $1}' | FZF_DEFAULT_OPTS="$opts" fzf)
+    [[ -z "$tree" ]] && return 1
+
+    if [[ -z "$reason" ]]; then
+        _forgit_git_worktree_lock "$tree"
+    else
+        _forgit_git_worktree_lock "$tree" --reason "$reason"
+    fi
```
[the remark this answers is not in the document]
suft: Thanks for clarifying. The logic will need to be adjusted to propagate the arguments and also consider positional arguments like in the example of `git-forgit worktree_lock --reason "whatever"` the `<worktree>` is a positional argument and so propagating in that situation with `git worktree lock "$@"` will cause an issue.

bin/git-forgit
```diff
@@ -1002,6 +1003,162 @@ _forgit_ignore_clean() {
...
+    _forgit_worktree_remove_git_opts=()
+    _forgit_parse_array _forgit_worktree_remove_git_opts "$FORGIT_WORKTREE_REMOVE_GIT_OPTS"
+    git worktree remove "${_forgit_worktree_remove_git_opts[@]}" "$@"
+}
+
+_forgit_worktree_remove() {
+    _forgit_inside_work_tree || _forgit_inside_git_dir || return 1
+    local worktree_list tree opts force_flags
+
+    force_flags=()
```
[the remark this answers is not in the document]
suft: Propagating the args should be straightforward for `remove` since it has no positional arguments compared to the `lock` (example above).

bin/git-forgit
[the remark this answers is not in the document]
suft: We could use the `--help` option for the git subcommand to determine if it requires a positional argument. But then we will need to go through the args to determine if the positional argument is passed.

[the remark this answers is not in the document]
suft: Or we could simply propagate in the scenario like `--reason "whatever"` and have it error out because the `<worktree> is missing.

bin/git-forgit
```diff
@@ -1176,6 +1189,141 @@ _forgit_paths_list() {
...
+_forgit_worktree_lock() {
+    _forgit_inside_work_tree || _forgit_inside_git_dir || return 1
+
+    local tree opts max_args
+    max_args=$(_forgit_worktree_lock_max_args)
+
+    if [[ $# -ge "$max_args" ]] || { [[ $# -ne 0 ]] && _forgit_all_non_flags "$@"; }; then
+        git worktree lock "$@"
+        return $?
+    fi
```
sandr01d: I don't like the idea of checking for the number of allowed arguments for the command using `--help`. If the user provides invalid arguments it is fine to just pass them to git like we do everywhere else. Git will print a descriptive error message. Please try to keep things as simple as possible. Depending on the output of `--help` is not a good idea in general either, as this could easily change between different git version.

bin/git-forgit
```diff
@@ -1176,6 +1189,141 @@ _forgit_paths_list() {
...
+    [[ -z "$tree" ]] && return 1
+    echo "$tree"
+}
+
+_forgit_worktree_lock_max_args() {
+    local usage
+    usage=$(git worktree lock --help 2>/dev/null | grep -E 'usage: git worktree lock' | head -n1)
+    usage=${usage#*:}
+     rest=$(echo "$usage" | awk '{for(i=4;i<=NF;++i)printf "%s ",$i; print ""}')
+    # shellcheck disable=SC2206
```
sandr01d: This function should be removed anyway, but as a general note, please don't just disable shellchecks. If you're having issues with something, feel free to ask, but don't just turn them off.

bin/git-forgit
```diff
@@ -1176,6 +1189,141 @@ _forgit_paths_list() {
...
+    _forgit_git_worktree_remove "$tree" "$@"
+}
+
+_forgit_worktree_unlock() {
+    _forgit_inside_work_tree || _forgit_inside_git_dir || return 1
+    if [[ $# -ne 0 ]]; then
+        git worktree unlock "$@"
+        worktree_unlock_status=$?
+        return $worktree_unlock_status
+    fi
```
sandr01d: This can be simplified

```suggestion
    [[ $# -ne 0 ]] && {
        git worktree unlock "$@"
        return $?
    }
```

bin/git-forgit
```diff
@@ -1176,6 +1189,141 @@ _forgit_paths_list() {
...
+        return
+    fi
+    # the trailing '--' ensures that this works for branches that have a name
+    # that is identical to a file
+    git log "$sha" "${_forgit_log_preview_options[@]}" --
+}
+
+_forgit_worktree_select() {
+    _forgit_inside_work_tree || _forgit_inside_git_dir || return 1
+    local worktree_list count tree opts
```
sandr01d: To keep things consistent with the rest of the code, I would prefer to name this variable to `worktrees`. This applies to all functions.

```suggestion
    local worktrees count tree opts
```

bin/git-forgit
```diff
@@ -1176,6 +1189,141 @@ _forgit_paths_list() {
...
+    worktree_list=$(git worktree list | grep -vE "prunable$" | awk '{print $1}' | _forgit_filter_existing_paths)
+
+    count=$(echo "$worktree_list" | wc -l)
+    [[ $count -eq 1 ]] && return 1
```
sandr01d: We should let the user know why we refuse to run a function, e.g. `echo "No worktrees to select"` in this specific case. This applies to all your functions.

bin/git-forgit
```diff
@@ -64,6 +65,18 @@ _forgit_previous_commit() {
     fi
 }
 
+_forgit_all_non_flags() {
```
sandr01d:
```suggestion
_forgit_only_non_flags() {
```

completions/_git-forgit
```diff
@@ -18,6 +18,11 @@ _git-stash-show() {
     _alternative "files:filename:($(git stash list | sed -n -e 's/:.*//p'))"
 }
 
+_git-worktrees() {
+    # --porcelain usually paired with -z but not needed since we use awk
+    _alternative "worktrees:worktree:($(git worktree list --porcelain | awk '/worktree/ {print $2}'))"
```
sandr01d: Please use `-z` here, as this allows handling worktrees that contain a newline character.

```suggestion
    __gitcomp_nl "$(git worktree list --porcelain -z | awk 'BEGIN {RS="\0"} /worktree/ {print $2}')"
```

The same applies to the zsh completions for this command.

sandr01d (changes requested):
There are a few more things we need to resolve. I'll do another functional review once done. The issue with the main worktree being present in the list of `gwl` is also not resoved.
Please do not close my comments, just let me know once you're done and I will review and close them. Closing them from your side makes it difficult for me to know where we last left off and what needs to be addressed.

README.md
```diff
@@ -176,6 +176,20 @@ If you're having issues after updating, and commands such as `forgit::add` or al
 
 - **Interactive `git commit --fixup=reword && git rebase -i --autosquash` selector** (`grw`)
 
+- **Interactive `git commit --squash && git rebase -i --autosquash` selector** (`gsq`)
+
+- **Interactive `git commit --fixup=reword && git rebase -i --autosquash` selector** (`grw`)
+
+- **Interactive `git worktree list` selector** (`gws`/`gwj`)
+
+  + `gwj` jumps to the worktree using `cd` and can only be used via the alias, no equivalent behavior using forgit as a git subcommand
```
sandr01d:
```suggestion
  - `gwj` jumps to the worktree using `cd` and can only be used via the alias, no equivalent behavior using forgit as a git subcommand
```

README.md
```diff
@@ -176,6 +176,20 @@ If you're having issues after updating, and commands such as `forgit::add` or al
 
 - **Interactive `git commit --fixup=reword && git rebase -i --autosquash` selector** (`grw`)
 
+- **Interactive `git commit --squash && git rebase -i --autosquash` selector** (`gsq`)
+
+- **Interactive `git commit --fixup=reword && git rebase -i --autosquash` selector** (`grw`)
+
```
sandr01d: This does not belong here

README.md
sandr01d:
```suggestion
  + `gwj` jumps to the worktree using `cd` and can only be used via the alias, there is no equivalent behavior when using forgit as a git subcommand.
```

README.md
```diff
@@ -275,29 +293,33 @@ git cf
...
+| `grb`       | `FORGIT_REBASE_GIT_OPTS`                                                    |
+| `gbl`       | `FORGIT_BLAME_GIT_OPTS`                                                     |
+| `gfu`       | `FORGIT_FIXUP_GIT_OPTS`                                                     |
+| `gsq`       | `FORGIT_SQUASH_GIT_OPTS`                                                    |
+| `grw`       | `FORGIT_REWORD_GIT_OPTS`                                                    |
+| `gcp`       | `FORGIT_CHERRY_PICK_GIT_OPTS`                                               |
+| `gws`/`gwj` | `FORGIT_WORKTREE_PREVIEW_GIT_OPTS`                                          |
+| `gwl`       | `FORGIT_WORKTREE_LOCK_GIT_OPTS`, `FORGIT_WORKTREE_PREVIEW_GIT_OPTS`         |
+| `gwr`       | `FORGIT_WORKTREE_REMOVE_GIT_OPTS`, `FORGIT_WORKTREE_PREVIEW_GIT_OPTS`       |
+| `gwu`       | `FORGIT_WORKTREE_PREVIEW_GIT_OPTS`                                          |
```
sandr01d: The `*_PREVIEW_GIT_OPTS` do not exist and don't belong here.

sandr01d (changes requested):

bin/git-forgit
```diff
@@ -1176,6 +1189,141 @@ _forgit_paths_list() {
...
+    _forgit_inside_work_tree || _forgit_inside_git_dir || return 1
+    if [[ $# -ne 0 ]]; then
+        git worktree unlock "$@"
+        worktree_unlock_status=$?
+        return $worktree_unlock_status
+    fi
+    local worktree_list tree opts
+
+    worktree_list=$(git worktree list | grep -v "(bare)" | grep -E "locked$")
+    count=$(echo "$worktree_list" | wc -l)
```
sandr01d: `$count` is 1 when there are actually no locked workspaces, because `echo` adds a newline (which is counted by `wc`). We can work around it using `grep` instead:

```suggestion
    count=$(echo "$worktree_list" | grep -c .)
```

This applies to multiple functions.

bin/git-forgit
```diff
@@ -1176,6 +1189,141 @@ _forgit_paths_list() {
...
+    # shellcheck disable=SC2206
+    local args=($rest)
+    echo "${#args[@]}"
+}
+
+_forgit_git_worktree_lock() {
+    _forgit_worktree_lock_git_opts=()
+    _forgit_parse_array _forgit_worktree_lock_git_opts "$FORGIT_WORKTREE_LOCK_GIT_OPTS"
+    git worktree lock "${_forgit_worktree_lock_git_opts[@]}" "$@"
+}
```
sandr01d: Would be nice to allow operating on multiple worktrees at the same time. Should be easy to implement by allowing multiselect in fzf and modifiying the `_forgit_git_worktree_*` functions like this:

```suggestion
_forgit_git_worktree_lock() {
    local trees
    trees=$1
    shift
    _forgit_worktree_lock_git_opts=()
    _forgit_parse_array _forgit_worktree_lock_git_opts "$FORGIT_WORKTREE_LOCK_GIT_OPTS"
    echo "$trees" | xargs -I% git worktree lock "${_forgit_worktree_lock_git_opts[@]}" %
}
```

bin/git-forgit
```diff
@@ -151,6 +152,33 @@ _forgit_is_file_tracked() {
...
+# Note: We explicitly print the current branch first rather than using
+# `LC_ALL=C sort -k1.1,1.1 -rs` because git branch output has three possible
+# prefixes: '*' (current), '+' (checked out in a worktree), and ' ' (other).
+# Sorting by the first character doesn't reliably place '*' first when '+'
+# is present, since their ASCII order (* < +) conflicts with the desired order.
+_forgit_branch_list() {
+    local current
+    current=$(git branch --show-current)
+    if [[ -n "$current" ]]; then
+        printf '\e[90m%s\e[0m\n' "* $current"
```
sandr01d: Might be worth adding an extra function for this, similar to `_forgit_warn`. Maybe `_forgit_dim`?

bin/git-forgit
```diff
@@ -703,18 +733,15 @@ _forgit_cherry_pick_from_branch() {
     opts="
         $FORGIT_FZF_DEFAULT_OPTS
         +s +m --tiebreak=index --header-lines=1
-        --preview=\"$FORGIT cherry_pick_from_branch_preview '$base' {1}\"
+        --preview=\"$FORGIT cherry_pick_from_branch_preview '$base' {}\"
```
sandr01d, line 739: Could you explain why this line was changed?

bin/git-forgit
```diff
@@ -172,7 +174,7 @@ _forgit_branch_list() {
 }
 
 # Extract branch name from git branch output
-# Handles ANSI escape codes, prefix characters (* + ' '), and symbolic ref (->)
+# Handles ANSI escape codes, prefix characters (* + ' '), and symbolic refs (->)
```
sandr01d, line 178: This comment was added in 0dd8946 and modified in b7382d9, but I think both changes should be in 0dd8946.

bin/git-forgit
```diff
@@ -1259,6 +1261,146 @@ _forgit_paths_list() {
...
+
+# Return deletable worktrees (exclude main worktree which is the first one)
+_forgit_worktree_list_deletable() {
+    _forgit_worktree_list | tail -n +2
+}
+
+# Extract worktree path from formatted line (strip ANSI codes, skip 5-char prefix '[XY] ')
+# TODO: awk '{print $1}' breaks on paths containing spaces or parentheses.
+# Consider switching _forgit_worktree_list to a tab-delimited format so we can
+# use 'cut -f1' (or awk -F'\t') for reliable path extraction.
```
sandr01d, line 1317: Is this something we should do before merging?

bin/git-forgit
```diff
@@ -1259,6 +1261,146 @@ _forgit_paths_list() {
...
+    [[ ! -d "$worktree" ]] && echo "Worktree directory not found: $worktree" && return 1
+
+    local status_output
+    status_output=$(git -c color.status=always -C "$worktree" status -s 2>/dev/null)
+    [[ -n "$status_output" ]] && echo "$status_output" && echo ""
+    git -C "$worktree" log --oneline -n 200 --color=always 2>/dev/null
+}
+
+# Git worktree delete wrapper
+_forgit_git_worktree_delete() {
```
sandr01d, line 1352: I'm wondering if we should change the terminology from `delete` to `remove` to stay consistent with what git is using. I'm also fine with keeping the current naming should you prefer that.

bin/git-forgit
```diff
@@ -1259,6 +1261,146 @@ _forgit_paths_list() {
...
+        +s --multi --tiebreak=index
+        --preview=\"$FORGIT worktree_preview {}\"
+        --bind=\"ctrl-y:execute-silent($FORGIT worktree_yank_path {})\"
+        $FORGIT_WORKTREE_DELETE_FZF_OPTS
+    "
+
+    worktrees=()
+    while IFS='' read -r line; do
+        [[ -n "$line" ]] && worktrees+=("$(echo "$line" | _forgit_extract_worktree_path)")
+    done < <(_forgit_worktree_list_deletable | FZF_DEFAULT_OPTS="$opts" fzf)
```
sandr01d: When there are no deletable worktrees, this opens fzf with an empty list. I think we should add an early out for this case, similar as we do with other functions.

completions/_git-forgit
```diff
@@ -18,6 +18,10 @@ _git-stash-show() {
     _alternative "files:filename:($(git stash list | sed -n -e 's/:.*//p'))"
 }
 
+_git-worktrees() {
+    _alternative "worktrees:worktree:($(git worktree list --porcelain 2>/dev/null | grep '^worktree ' | cut -d' ' -f2-))"
```
sandr01d: I think we should use `-z` together with `--porcelain`. From the git man page:

> --porcelain
With list, output in an easy-to-parse format for scripts. This format will remain stable
across Git versions and regardless of user configuration. It is recommended to combine this
with -z. See below for
-z
Terminate each line with a NUL rather than a newline when --porcelain is specified with
list. This makes it possible to parse the output when a worktree path contains a newline
character.

Same for the bash variant.

completions/_git-forgit
```diff
@@ -102,6 +108,8 @@ _git-forgit() {
     squash) _git-log ;;
     stash_show) _git-stash-show ;;
     show) _git-show ;;
+    worktree) ;;
```
sandr01d: Should this be

```suggestion
    worktree) _git-worktree ;;
```

Which would suggest the `git worktree` subcommands, which is probably what we want here.

completions/git-forgit.bash
```diff
@@ -108,6 +115,7 @@ _git_forgit()
                 show) _git_show ;;
                 squash) _git_log ;;
                 stash_show) _git_stash_show ;;
```
sandr01d, line 121: `worktree` is missing here.

bin/git-forgit
```diff
@@ -1259,6 +1261,146 @@ _forgit_paths_list() {
...
+_forgit_worktree_delete() {
+    _forgit_inside_git_repo || return 1
+    local opts worktrees
+    [[ $# -ne 0 ]] && { _forgit_git_worktree_delete "$@"; return $?; }
+
+    opts="
+        $FORGIT_FZF_DEFAULT_OPTS
+        +s --multi --tiebreak=index
+        --preview=\"$FORGIT worktree_preview {}\"
+        --bind=\"ctrl-y:execute-silent($FORGIT worktree_yank_path {})\"
```
sandr01d, line 1396: I think we should include the `worktree_toggle_lock` binding here too. This would allow to unlock and delete a worktree in one go, which I would find useful.

bin/git-forgit
wfxr: Good point. I agree this is a limitation of the current parser. Fixing it properly would require changing the display/parse contract and would add quite a bit of complexity to this PR. For now, I’d prefer to keep the TODO and defer this until we either have a concrete use case or find a cleaner approach. My preference is to keep this PR scoped and avoid making the worktree parsing logic more complex unless we really need it.

wfxr: I created an issue (#493) to track this.

bin/git-forgit
wfxr: This was changed intentionally. The preview input is now passed through _forgit_extract_branch_name(), because the selected line can include prefixes and ANSI formatting from _forgit_branch_list() such as *  for the current branch or +  for branches checked out in a worktree. Using {} ensures the parser receives the full selected line, whereas {1} would only pass the first token and break in those cases.

bin/git-forgit
wfxr: You're right, this comment change would fit better into 0dd8946. Since this is only a very small follow-up tweak to the comment text, I was not planning to rewrite the commit history just for that. What do you think?

bin/git-forgit
wfxr: I see the consistency argument, but I would prefer to keep `delete` here for now. forgit already uses `delete` in other public command names such as `branch_delete`, so switching only the worktree variant to `remove` would make the naming less consistent within forgit itself. If you think we should align all public naming with Git terminology, I’m open to doing that more systematically, but I would rather not change just this one in isolation.

bin/git-forgit
wfxr: Good suggestion! Extracted it into a helper function, but named it `_forgit_print_dim` instead — since `_forgit_warn`/`_forgit_info` are logging functions (they write to stderr with a prefix), while this one is purely for formatting output to stdout without any prefix. The `print_` prefix helps make that distinction clear.

completions/_git-forgit
wfxr: done

completions/_git-forgit
wfxr: done

completions/git-forgit.bash
wfxr: done

bin/git-forgit
wfxr: done.

bin/git-forgit
sandr01d: Makes sense, thanks for the explanation :+1:

bin/git-forgit
sandr01d: I'd prefer to have the change in https://github.com/wfxr/forgit/commit/0dd8946990b4b1719e332c5576095324293afa4e, which should be possible with a quick rebase. That being said, this is of course a very minor thing, so feel free to close this thread and move ahead should you decide to not go this route.

bin/git-forgit
sandr01d: Fine with me

bin/git-forgit
sandr01d: In this case I think it is best to keep as is. I don't think aligning all functions with gits terminology is worth it, especially because that would be a breaking change.

bin/git-forgit
wfxr: Yes, it is indeed a very simple fix, and I've completed it.

PR #401, opened 2024-11-02
Meta: do not create release when there are no public changes

carlfriedrich:
## Check list

- [x] I have performed a self-review of my code
- [x] I have commented my code in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation

## Description

The latest forgit release did not contain any release notes, because the only commit since the previous release had the `Meta` prefix in its git summary, and we do not include these commits in the release notes.

In such a case we should not create a release at all, since we do not have any code changes. I fixed this in the GHA jobs and tested it in my private fork. Will merge this next week if no complaints come in.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [ ] zsh
    - [ ] fish
- OS
    - [ ] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

PR #414, opened 2024-12-18
Meta: do not create tag when there are no public changes

### LINKED ISSUE #413: missing `forgit-24.12.0.tar.gz` for 24.12.0 tag
👋 looks like `forgit-24.12.0.tar.gz` release artifact is missing for 24.12.0 tag.

cc @suft

suft:
## Check list

## Description

- Related to https://github.com/wfxr/forgit/pull/401
- Closes #413

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [ ] zsh
    - [ ] fish
- OS
    - [ ] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:


suft:
Do we want to check all the commit messages between the last tag and head for the `Meta` prefix?

sandr01d:
Hey @suft, thanks for helping out with this!

> Do we want to check all the commit messages between the last tag and head for the Meta prefix?
If all are Meta we exit else we create a new tag.

> If that's the case I have the fix stashed and I can update this PR with that approach.

Yes, I think that is the behavior we'd want. Ideally we'd match the Meta prefix case insensitive.

carlfriedrich:
Thanks @suft for your take on this. While this is definitely a quick fix for the problem, I do not find it optimal because it duplicates the logic we already have in [`changelog.sh`](https://github.com/wfxr/forgit/blob/main/.github/scripts/changelog.sh#L17).
In order to keep the code [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) I would prefer combining the two scripts into one, since tagging and releasing should always go hand in hand anyway.
I can offer to implement this, as I already have the idea in mind, but I think I won't have time for this until after Christmas.

carlfriedrich:
@suft Thanks again for your update. The new version won't work, unfortunately, because the output of the tag creation goes to `stdout`, which means that it will land in the `CHANGELOG.md` file, which is not what we want.

Furthermore, the name of the script `changelog.sh` does not reflect its extended functionality anymore.

Now that I think more about it, though, I think the two scripts might still be separate, but we should only reduce the two workflows into one. The `Tag` step then should have the same `if: steps.changelog.outcome == 'success'` conditional as the `Release` step in order to execute it only when there are public changes. This will keep the code clean and DRY while still sticking to the [single responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle).

My offer is still valid, I can implement this myself during the next days. If you @suft are motivated, though, feel free to continue on this. Your effort is very much appreciated! 🙏

carlfriedrich:
Another thing just came to my mind: there was a reason we had the tag and release actions separate, and that is we want to have the possibility to trigger custom releases by creating a tag manually. So the new combined workflow should also support this (i.e. triggered on tag creation and create tag only if it does not exist, yet).

PR #415, opened 2024-12-21
fix: fix edit file key binding (alt-e) in diff and add

wfxr:
## Check list

## Description

The key binding `alt-e` for editing a file in diff and add was not working properly on recent versions of fzf. As its documentation suggests, we should replace `execute-silent` with `execute` since we need to switch to a new screen and handle both input and output.

> fzf switches to the alternate screen when executing a command. However, if the command is expected to complete quickly, and you are not interested in its output, you might want to use execute-silent instead, which silently executes the command without the switching.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:


sandr01d (approved):
Can confirm the issue and the fix

cjappl (approved):

Issue #328, opened 2023-10-27
forgit::stash::push failing with error: unknown option `pathspec-file-nul'


Halfwalker:
## Check list

- [x] I have read through the [README](https://github.com/wfxr/forgit/blob/master/README.md)
- [x] I have the latest version of forgit
- [x] I have searched through the existing issues

## Environment info

- OS
    - [x] Linux (Ubuntu 20.04.6)
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish

## Problem / Steps to reproduce

Running `forgit::stash::push` via an alias `gsp=forgit::stash::push` results in the following error
```
❯ gsp
error: unknown option `pathspec-file-nul'
usage: git stash [push [-p|--patch] [-k|--[no-]keep-index] [-q|--quiet]
          [-u|--include-untracked] [-a|--all] [-m|--message <message>]
          [--] [<pathspec>...]]

    -k, --keep-index      keep index
    -p, --patch           stash in patch mode
    -q, --quiet           quiet mode
    -u, --include-untracked
                          include untracked files in stash
    -a, --all             include ignore files
    -m, --message <message>
                          stash message
```
This is on Ubuntu 20.04.6 with **zsh** and **git** versions as follows
* git version 2.25.1
* zsh 5.8 (x86_64-ubuntu-linux-gnu)
```
❯ dpkg -l git\* | rg ii
ii  git                 1:2.25.1-1ubuntu3.11 amd64        fast, scalable, distributed revision control system
ii  git-extras          5.1.0-1              all          Extra commands for git
ii  git-lfs             2.9.2-1              amd64        Git Large File Support
ii  git-man             1:2.25.1-1ubuntu3.11 all          fast, scalable, distributed revision control system (manual pages)
❯ dpkg -l zsh\* | rg ii
ii  zsh            5.8-3ubuntu1.1 amd64        shell with lots of features
ii  zsh-common     5.8-3ubuntu1.1 all          architecture independent files for Zsh
```

Git repo status is
```
❯ g status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   docker/vault/keys/apinext_token.yaml
        modified:   docker/vault/keys/login_token.yaml

no changes added to commit (use "git add" and/or "git commit -a")
```

[...]

Issue #259, opened 2022-12-12
Discussion: keybindings to support more functions?


carlfriedrich:
## Check list

- [x] I have read through the [README](https://github.com/wfxr/forgit/blob/master/README.md)
- [x] I have the latest version of forgit
- [x] I have searched through the existing issues

## Environment info

- OS
    - [x] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Others:
- Shell
    - [x] bash
    - [x] zsh
    - [x] fish

## Problem / Steps to reproduce

This came up during a discussion in #251.

`fzf` supports a [list of possible hotkeys](https://www.mankier.com/1/fzf#Key/Event_Bindings) that can be bound to custom commands and then be called on list items while browsing them. I personally make use of this in my `gss` selector by defining these options:

```shell
# Use forgit diff on stash enter
export FORGIT_STASH_ENTER_COMMAND='echo {} | cut -d: -f1 |
	xargs -I % bash -ic "git forgit diff %^1 %"'

# Pop stash on alt-enter
export FORGIT_STASH_POP_COMMAND='echo {} | cut -d: -f1 | xargs -I % git stash pop %'

# Drop stash on alt-backspace
export FORGIT_STASH_DROP_COMMAND='echo {} | cut -d: -f1 | xargs -I % git stash drop %'

export FORGIT_STASH_FZF_OPTS="
	--preview-window=top:50%
	--bind='enter:execute($FORGIT_STASH_ENTER_COMMAND)'
	--bind='alt-enter:execute($FORGIT_STASH_POP_COMMAND)+accept'
	--bind='alt-bspace:execute($FORGIT_STASH_DROP_COMMAND)+reload(git stash list)'
	--prompt='[ENTER] show   [ALT+ENTER] pop   [ALT+BACKSPACE] drop > '
"
```

This makes the stash selector look as follows:

![grafik](https://user-images.githubusercontent.com/45259958/207112533-9d61fb28-9d8c-45ab-a925-398b85cab9ba.png)

I can now view, apply or drop stashes all from within the same stash list.

I kind of "hijacked" the fzf prompt to display the available keybindings. Something similar could be achieved passing `--header "[ENTER] show   [ALT+ENTER] pop   [ALT+BACKSPACE] drop" --header-first` to `fzf`, which might be a bit cleaner. I just found putting it into the prompt visually more appealing.

@cjappl @sandr01d @wfxr
[...]

PR #417, opened 2024-12-29
Feature: add forgit show command

### LINKED ISSUE #416: `forgit log` cannot diff merge commits
## Check list

- [x] I have read through the [README](https://github.com/wfxr/forgit/blob/main/README.md)

## Environment info

- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:
- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish

Merge commits cannot be diffed from within `forgit log`.

Steps to reproduce:

```shell
git init
touch 1.txt
git add 1.txt
git commit -m "Add file 1"
git checkout -b branch
touch 2.txt
git add 2.txt
git commit -m "Add file 2"
git checkout main
git merge --no-ff branch -m "Merge branch"
```

Afterwards call `git forgit log` and try to view the merge commit by pressing `enter`. It does not work (immediately returns).

The reason is that our [current implementation](https://github.com/wfxr/forgit/blob/df5191157f552abee3176e0ff760f42bd4cc6d6d/bin/git-forgit#L188C47-L188C49) using the `^!` operator on the commit does not work for merge commits. According to [this stack overflow answer](https://stackoverflow.com/a/40986893/3018229) there are only three ways to correctly show the diff for a merge commit:

```shell
git show -m c05f017
git show --first-parent c05f017
git diff c05f017^ c05f017
```

A quick fix would be to change our implementation to the third option. The cleaner solution IMO, though, would be to implement a `forgit show` command. I am planning to do the latter.

carlfriedrich:
Use the new command in git log enter. This fixes the display of diffs for merge commits.

Fixes #416.

## Check list

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,8 @@ Then add the following to your shell's config file:
 
 - **Interactive `git diff` viewer** (`gd`)
 
+- **Interactive `git show` viewer** (`gso`)
+
 - **Interactive `git reset HEAD <file>` selector** (`grh`)
 
 - **Interactive `git checkout <file>` selector** (`gcf`)
@@ -172,6 +174,7 @@ You can change the default aliases by defining these variables below.
 forgit_log=glo
 forgit_reflog=grl
 forgit_diff=gd
+forgit_show=gso
 forgit_add=ga
 forgit_reset_head=grh
 forgit_ignore=gi
@@ -232,6 +235,7 @@ These are passed to the according `git` calls.
 | `glo`    | `FORGIT_LOG_GIT_OPTS`                                                       |
 | `grl`    | `FORGIT_REFLOG_GIT_OPTS`                                                    |
 | `gd`     | `FORGIT_DIFF_GIT_OPTS`                                                      |
+| `gso`     | `FORGIT_SHOW_GIT_OPTS`                                                      |
 | `grh`    | `FORGIT_RESET_HEAD_GIT_OPTS`                                                |
 | `gcf`    | `FORGIT_CHECKOUT_FILE_GIT_OPTS`                                             |
 | `gcb`    | `FORGIT_CHECKOUT_BRANCH_GIT_OPTS`, `FORGIT_CHECKOUT_BRANCH_BRANCH_GIT_OPTS` |
@@ -286,6 +290,7 @@ Customizing fzf options for each command individually is also supported:
 | `grl`    | `FORGIT_REFLOG_FZF_OPTS`          |
 | `gi`     | `FORGIT_IGNORE_FZF_OPTS`          |
 | `gd`     | `FORGIT_DIFF_FZF_OPTS`            |
+| `gso`    | `FORGIT_SHOW_FZF_OPTS`            |
 | `grh`    | `FORGIT_RESET_HEAD_FZF_OPTS`      |
 | `gcf`    | `FORGIT_CHECKOUT_FILE_FZF_OPTS`   |
 | `gcb`    | `FORGIT_CHECKOUT_BRANCH_FZF_OPTS` |
--- a/README.md
+++ b/README.md
@@ -29,7 +29,13 @@ It's **lightweight** and **easy to use**.
 
 # 📥 Installation
 
-*Make sure you have [`fzf`](https://github.com/junegunn/fzf) installed.*
+## Requirements
+
+- [`fzf`](https://github.com/junegunn/fzf) version `0.49.0` or higher
+
+  If your OS package manager bundles an older version of `fzf`, you might install it using [`fzf`'s own install script'](https://github.com/junegunn/fzf?tab=readme-ov-file#using-git).
+
+## Shell package managers
 
 ``` zsh
 # for zplug
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -353,7 +353,18 @@ _forgit_show_view() {
     repo=$(git rev-parse --show-toplevel)
     cd "$repo" || return 1
     echo "$input_line" | _forgit_get_files_from_diff_line | xargs -0 \
-        "$FORGIT" exec_show "${commit}" -U"$diff_context" -- | _forgit_pager diff
+        "$FORGIT" exec_show "${commit}^{commit}" -U"$diff_context" -- | _forgit_pager diff
+}
+
+_forgit_show_preview() {
+    local input_line=$1
+    local diff_context=$2
+    local commit=$3
+    if [[ "$FZF_PREVIEW_LABEL" =~ "Diff" ]]; then
+        _forgit_show_view "${input_line}" "${diff_context}" "${commit}"
+    else
+        git show --quiet --color=always "${FZF_PROMPT%% *}"
+    fi
 }
 
 _forgit_show_enter() {
@@ -381,8 +392,12 @@ _forgit_show() {
     opts="
         $FORGIT_FZF_DEFAULT_OPTS
         +m -0 --bind=\"enter:execute($FORGIT show_enter {} $escaped_commit | $FORGIT pager enter)\"
-        --preview=\"$FORGIT show_view {} '$_forgit_preview_context' $escaped_commit\"
+        --preview=\"$FORGIT show_preview {} '$_forgit_preview_context' $escaped_commit\"
+        --preview-label=\" Diff \"
         --bind=\"alt-e:execute($FORGIT edit_diffed_file {})+refresh-preview\"
+        --bind=\"alt-t:transform:[[ ! \\\"\$FZF_PREVIEW_LABEL\\\" =~ 'Diff' ]] &&
+                echo 'change-preview-label( Diff )+refresh-preview' ||
+                echo 'change-preview-label( Commit Message )+refresh-preview'\"
         $FORGIT_DIFF_FZF_OPTS
         --prompt=\"${commit} > \"
     "
@@ -1104,7 +1119,7 @@ private_commands=(
     "revert_preview"
     "reset_head_preview"
     "show_enter"
-    "show_view"
+    "show_preview"
     "stash_push_preview"
     "stash_show_preview"
     "yank_sha"
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -12,6 +12,21 @@
 # This gives users the choice to set aliases inside of their git config instead
 # of their shell config if they prefer.
 
+# Check if fzf is installed
+installed_fzf_version=$(fzf --version 2>/dev/null | awk '{print $1}')
+if [[ -z "$installed_fzf_version" ]]; then
+    echo "fzf is not installed. Please install fzf first."
+    exit 1
+fi
+
+# Check fzf version
+required_fzf_version="0.49.0"
+higher_fzf_version=$(printf '%s\n' "$required_fzf_version" "$installed_fzf_version" | sort -V | tail -n1)
+if [[ "$higher_fzf_version" != "$installed_fzf_version" ]]; then
+    echo "fzf version $required_fzf_version or higher is required. You have $installed_fzf_version."
+    exit 1
+fi
+
 # Set shell for fzf preview commands
 # Disable shellcheck for "which", because it suggests "command -v xxx" instead,
 # which is not a working replacement.
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -185,7 +185,7 @@ _forgit_log_enter() {
     local sha
     sha=$(echo "$1" | _forgit_extract_sha)
     shift
-    echo "$sha" | xargs -I% "${FORGIT}" diff %^! "$@"
+    echo "$sha" | xargs -I% "${FORGIT}" show % "$@"
 }
 
 # git commit viewer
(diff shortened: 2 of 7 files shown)


carlfriedrich:
Not sure about the fish completions, we don't have them for `forgit diff` either. @cjappl or @sandr01d do you know more about this?

cjappl:
> Not sure about the fish completions, we don't have them for `forgit diff` either. @cjappl or @sandr01d do you know more about this?

Hmm, good find. Would you file a ticket for this and assign it to me and I can fix it when I get time? I can confirm git diff does not behave with completions on fish.

cjappl (approved):

wfxr (approved):
LGTM.

sandr01d (changes requested):
Looking forward to see this feature merged, I definitely need this :smiley:
In addition to things I've pointed out in the code, there are two things I've noticed:

1. The show command has not been added to _conf.d/forgit.plugin.fish_ and as a result does not work in fish. Might the issue you were having with fish completions be related to this? 
2. Completions for zsh are missing. If I remember correctly, you're not a zsh user @carlfriedrich, so if you don't want to go through the hassle of setting up zsh on your system I'd offer to implement a zsh completion and add it to this PR. Just let me know.

bin/git-forgit
```diff
@@ -339,6 +339,65 @@ _forgit_diff() {
...
+    else
+        commit="HEAD"
+    fi
+    # Escape opening brackets to support stashes (see comment in _forgit_diff)
+    escaped_commit=${commit//\{/\\\\\{}
+    opts="
+        $FORGIT_FZF_DEFAULT_OPTS
+        +m -0 --bind=\"enter:execute($FORGIT show_enter {} $escaped_commit | $FORGIT pager enter)\"
+        --preview=\"$FORGIT show_view {} '$_forgit_preview_context' $escaped_commit\"
+        --bind=\"alt-e:execute-silent($FORGIT edit_diffed_file {})+refresh-preview\"
```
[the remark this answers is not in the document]
carlfriedrich: Good catch, thank you! I updated this accordingly.

README.md
```diff
@@ -101,6 +101,8 @@ Then add the following to your shell's config file:
 
 - **Interactive `git diff` viewer** (`gd`)
 
+- **Interactive `git show` viewer** (`gs`)
```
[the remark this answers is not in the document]
carlfriedrich: Hm, I see. Not sure about this, though. Is ghostscript actually commonly installed? I don't have it installed, and it does not ship with Ubuntu by default for example. And furthermore, if you have it installed, do you actually use it by calling the command interactively, or isn't it more a kind of util used by scripts?
IMO we could stick with `gs`, since I don't particularly like the alternatives (and would override it to `gs` anyway). Users can still override it to something different in their shell config if it interferes with ghostscript.
However, since I am not using the shell aliases myself anyway, I am open to any other suggestion.

carlfriedrich:
Thanks for your reviews everyone, especially for the good finds @sandr01d.

>     1. The show command has not been added to _conf.d/forgit.plugin.fish_ and as a result does not work in fish. Might the issue you were having with fish completions be related to this?

I am sorry, I simply forgot this. Updated it in the new commit version. However, I did not test the fish completions, I just did not add anything in the fish completions file. I was looking for something to copy from `forgit diff`, but then noticed that that's missing as well.

>     2. Completions for zsh are missing. If I remember correctly, you're not a zsh user @carlfriedrich, so if you don't want to go through the hassle of setting up zsh on your system I'd offer to implement a zsh completion and add it to this PR. Just let me know.

Good point, I added them in the new commit version, but haven't tested them. Can you give them a quick test?

> Hmm, good find. Would you file a ticket for this and assign it to me and I can fix it when I get time? I can confirm git diff does not behave with completions on fish.

@cjappl Yep, thanks a lot, I will do so.

sandr01d:
Thanks for the updates @carlfriedrich. Fish is working now (except completions, but we can handle this in #418, so no objections from my side). Zsh completions are working fine too.
There is one additional thing I've noticed while testing: the current implementation does not work well with tags. The reason for this is that git provides some additional information when showing a tag. E.g. `git show --pretty="" --name-status 24.10.0` produces the following output:

```
tag 24.10.0
Tagger: github-actions <github-actions@users.noreply.github.com>

Release 24.10.0
M	README.md
M	bin/git-forgit
M	completions/_git-forgit
M	completions/git-forgit.bash
M	completions/git-forgit.fish
M	conf.d/forgit.plugin.fish
M	forgit.plugin.zsh
```

While this probably is an edge case, we should be able to filter out the additional information quite easily.

README.md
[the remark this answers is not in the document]
sandr01d: > Is ghostscript actually commonly installed?

I think it is, at least as a dependency. On my system it is a dependency of my printers drivers and LaTeX.

> And furthermore, if you have it installed, do you actually use it by calling the command interactively, or isn't it more a kind of util used by scripts?

That is true though, I never use it interactively.

Thinking more about this, it came to my mind that a lot of people probably use `gs` as an alias for `git status`. At least that's what I've seen in the past.  So I would still favor using a different alias to not break peoples workflows.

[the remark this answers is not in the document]
carlfriedrich: `gs` being a common alias for `git status` is a good point. I went with `gso` (for "git show object") now. Deal?

carlfriedrich:
> Thanks for the updates @carlfriedrich. Fish is working now (except completions, but we can handle this in #418, so no objections from my side). Zsh completions are working fine too. There is one additional thing I've noticed while testing: the current implementation does not work well with tags. The reason for this is that git provides some additional information when showing a tag. E.g. `git show --pretty="" --name-status 24.10.0` produces the following output:

@sandr01d Good catch, thank you! I updated the call producing the file list to use the commit object instead of the tag. I kept the original call for the preview and enter commands, though. IMO if you call `git show` on a tag you actually might want to see the tag information, so now it is displayed above the diff for each file. Are you fine with that?

README.md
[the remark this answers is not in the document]
sandr01d: Works for me, thanks!

sandr01d:
>  I kept the original call for the preview and enter commands, though. IMO if you call git show on a tag you actually might want to see the tag information, so now it is displayed above the diff for each file. Are you fine with that?

Yes, absolutely. Having the tag information displayed is a big plus IMO.

> Thinking more about this I am asking myself whether people might be interested in the commit message as well when calling forgit show. Should we display it somehow? Maybe implement a preview toggle so that we can switch between the commit message and the file diff (using [transform action](https://github.com/junegunn/fzf/releases/tag/0.45.0))?

I think this is a great idea! Wasn't aware of the `transform` action — what a neat feature. We'd have to check for the installed fzf version tough. I suspect a lot of users running versions older than 0.45.0.

carlfriedrich:
@sandr01d So I gave it a shot and added a keybinding to toggle between the commit message and the commit diff. Works pretty well IMO and I already see myself using it a lot.

I also added a border to the preview window indicating what is currently shown. And furthermore I added a border displaying the keybindings.

I think maybe the latter might trigger a more general discussion (as I already suggested someting similar in #259) and maybe we might split that part out to a new MR.

I did not yet check the fzf version, yet (we need at least 0.49.0 now). I wanted to discuss how we continue on this, first. Maybe we should merge the show feature as it was before the new commit (because it already worked before) and add everything else in one or more follow-ups.

Let me know what you think and how you like the proposal.

bin/git-forgit
```diff
@@ -339,6 +339,77 @@ _forgit_diff() {
...
+        $FORGIT_FZF_DEFAULT_OPTS
+        +m -0 --bind=\"enter:execute($FORGIT show_enter {} $escaped_commit | $FORGIT pager enter)\"
+        --preview=\"$FORGIT show_view {} '$_forgit_preview_context' $escaped_commit\"
+        --bind=\"alt-e:execute($FORGIT edit_diffed_file {})+refresh-preview\"
+        --bind 'ctrl-t:transform:[[ ! \$FZF_PREVIEW_LABEL =~ Diff ]] &&
+                echo \"change-preview-label( Diff )+refresh-preview\" ||
+                echo \"change-preview-label( Commit Message )+refresh-preview\"'
+        --preview '[[ \$FZF_PREVIEW_LABEL =~ Diff ]] &&
+                $FORGIT show_view {} '$_forgit_preview_context' $escaped_commit ||
+                git show --quiet --color=always \${FZF_PROMPT%% *}'
```
sandr01d, line 391: I don't like that this introduces more deferred execution again. In this case we'd actually also have to do weird escaped quoting:

```suggestion
        --bind 'ctrl-t:transform:[[ ! \"\$FZF_PREVIEW_LABEL\" =~ Diff ]] &&
                echo \"change-preview-label( Diff )+refresh-preview\" ||
                echo \"change-preview-label( Commit Message )+refresh-preview\"'
        --preview '[[ \"\$FZF_PREVIEW_LABEL\" =~ Diff ]] &&
                $FORGIT show_view {} '$_forgit_preview_context' $escaped_commit ||
                git show --quiet --color=always \"\${FZF_PROMPT%% *}\"'
```

How about we move the part for the preview into `_forgit_show_view` or create a separate function for it? I think this would make it easier to maintain.

sandr01d (changes requested):
I find this super useful, well done @carlfriedrich! See my comments below regarding implementation details.
I think moving the second commit to a new PR would probably be best. E.g., I'm not a big fan of the bottom border and would prefer to have the key bindings displayed without it - but don't know if fzf supports this. With a new PR, we could also implement showing the existing keybindings of other functions (editing files with diff and add).

bin/git-forgit
carlfriedrich: Moving the preview part to a new function was a very good idea, thanks for that! Implemented that in the new version. I left the `--bind` argument as deferred code, though, as this is exactly how `fzf` shows it in their examples and the echoed strings would not make sense without the context of `--bind`. Do you agree?

sandr01d: Yes, I think there is now way to move the `--bind` arguments out of deferred code, which is fine for me. We should probably double quote `$FZF_PREVIEW_LABEL` in the deferred code.

sandr01d:
> I also changed the keybinding from ctrl-t to alt-t, because we also use alt-e for edit. Ok for you?

Yeah, makes sense to keep the keybinding consistent with what we already have.

Once the opened thread above is resolved the only thing missing would be checking for the fzf version, so we don't break the function for users of older fzf versions.

bin/git-forgit
carlfriedrich: I added the double quotes like you suggested. Looks a bit ugly, but works. I wanted to keep the double quotes for the complete `--bind` argument to be consistent with all the other args, so I had to double escape the quotes now.

carlfriedrich:
@sandr01d Thanks again. Please check if the quotes are fine for your and resolve the thread in case.

Concerning the fzf version check: do you have an opinion what we do when the version is below 0.49.0? Should we exit forgit with an error immediately or should we just disable the features we cannot use with older versions?
I would prefer the former, as partially disabling features would bloat the code, and basically everybody should be able to install a recent fzf version. There might be lazy people relying on old versions in their Ubuntu's apt repo ([Ubuntu 24.04 packages fzf 0.46.1](https://launchpad.net/ubuntu/oracular/+source/fzf)), but is that a reason not to move on with forgit? I mean, these people still can install old versions of forgit working with their fzf version as well.
WDYT?

Also pinging @cjappl and @wfxr for an opinion.

sandr01d:
> Concerning the fzf version check: do you have an opinion what we do when the version is below 0.49.0? Should we exit forgit with an error immediately or should we just disable the features we cannot use with older versions?
I would prefer the former, as partially disabling features would bloat the code, and basically everybody should be able to install a recent fzf version. There might be lazy people relying on old versions in their Ubuntu's apt repo ([Ubuntu 24.04 packages fzf 0.46.1](https://launchpad.net/ubuntu/oracular/+source/fzf)), but is that a reason not to move on with forgit? I mean, these people still can install old versions of forgit working with their fzf version as well.
WDYT?

This is a tough one. I agree that introducing version checks for specific features of fzf would make maintenance difficult if we decide to implement similar features like the one in question for other functions (which is something I'd like to see). My only concern is that there might be some users in restricted environments that are not able to install a different version of fzf than their distributions package manager provides. We had a similar case with #328. I wonder how many users would be negatively affected by such a change. Maybe it would be worth to open a discussion and start a poll on which distributions/versions people use? Not sure whether we'd see many responses, but might be worth a try.
I am currently leaning more towards requiring a newer version of fzf for forgit in general, because I think we should not hold back development. Would be interested to know what the other maintainers think.

cjappl:
Sorry for the delay! Thanks for pinging :) 

> Concerning the fzf version check: do you have an opinion what we do when the version is below 0.49.0? Should we exit forgit with an error immediately or should we just disable the features we cannot use with older versions?
I would prefer the former, as partially disabling features would bloat the code, and basically everybody should be able to install a recent fzf version. There might be lazy people relying on old versions in their Ubuntu's apt repo ([Ubuntu 24.04 packages fzf 0.46.1](https://launchpad.net/ubuntu/oracular/+source/fzf)), but is that a reason not to move on with forgit? I mean, these people still can install old versions of forgit working with their fzf version as well.
WDYT?

Generally in agreement with everything said here. I think we should do a (possibly major) version bump of forgit, and take in the new fzf.

> Should we exit forgit with an error immediately

This is my preference. I am against bloating the code with fzf version check

> I mean, these people still can install old versions of forgit working with their fzf version as well.

This is the key I think. This is why we version our code, so people can fall back to an older one. **Do we have to do a major version number bump as this is a backwards incompatible change?  (That would be my first thought)**

I think one of the possible good things here is that forgit is not really used as a CI/CD tool, but on devs machines directly where they have more control. I think this probably means we can be more aggressive moving forward than something crucial to many people's build pipelines.

If we get many many reports we can always roll back and remove the feature, and I will eat my words :)

carlfriedrich:
@sandr01d @cjappl Thanks for your opinions!

> My concern is that there might be some users in restricted environments that are not able to install a different version of fzf than their distributions package manager provides.

I don't see that, actually. Linux is flexible and `fzf` is distributed as a single binary. Everybody should be able to put it in their home directory under something like `~/.local/bin` and add that to their `PATH` (that's how I have it installed, it's actually part of my [dotfiles](https://github.com/carlfriedrich/dotfiles/tree/main/local/bin)).
The most user-friendly way is possibly to use `fzf`'s own `install` script, which is documented in their [README](https://github.com/junegunn/fzf?tab=readme-ov-file#using-git). We could point to that in our `README` as well.

> We had a similar case with https://github.com/wfxr/forgit/issues/328. I wonder how many users would be negatively affected by such a change.

This was actually about `git`, which is a more common tool to have installed via the package manager than `fzf` I would argue.

> Do we have to do a major version number bump as this is a backwards incompatible change? (That would be my first thought)

Since we don't use semantic versioning, there is no such thing as major versions in `forgit`. I think we should highlight this in the release notes and add it to the `README`, and we should make sure that the new version dependency is correctly modeled in the AUR and brew recipes.

If that's fine with you then, I would implement the hard version check.

sandr01d:
Hey @carlfriedrich, sorry for the delay.

> If that's fine with you then, I would implement the hard version check.

Yes, please go ahead.

carlfriedrich:
@sandr01d @cjappl I have implemented the version check and added it to the README as well as a separate commit, which will appear in the release notes. Can you check if that works for you? Any suggestions for improvements?

cjappl (approved):
LGTM - good work

bin/git-forgit
```diff
@@ -12,6 +12,21 @@
...
+    exit 1
+fi
+
+# Check fzf version
+required_fzf_version="0.49.0"
+higher_fzf_version=$(printf '%s\n' "$required_fzf_version" "$installed_fzf_version" | sort -V | head -n1)
+if [[ "$higher_fzf_version" != "$required_fzf_version" ]]; then
+    echo "fzf version $required_fzf_version or higher is required. You have $installed_fzf_version."
+    exit 1
+fi
```
sandr01d, line 28: This works fine, but I think what we store in `$higher_fzf_version` is actually the lower version. `sort -V` sorts lowest →highest and we store the first line. The check is fine, but I think we should rename the variable.

```suggestion
# Check fzf version
required_fzf_version="0.49.0"
lower_fzf_version=$(printf '%s\n' "$required_fzf_version" "$installed_fzf_version" | sort -V | head -n1)
if [[ "$lower_fzf_version" != "$required_fzf_version" ]]; then
    echo "fzf version $required_fzf_version or higher is required. You have $installed_fzf_version."
    exit 1
fi
```

sandr01d (changes requested):
Just one minor thing

sandr01d:
I prepared restricting fzf to version >= 0.49.0 for the AUR packages. @cjappl can you handle brew?

bin/git-forgit
carlfriedrich: @sandr01d Good catch, thanks a lot. I kept the variable name but changed the implementation to using `tail` instead of `head` and checking against `installed_fzf_version`. Makes the code easier to read IMO. Works just like before. Just to be sure, can you check again? I want to make sure to get this right, as people might be upset when the check does not work correctly.

cjappl:
> I prepared restricting fzf to version >= 0.49.0 for the AUR packages. @cjappl can you handle brew?

Yes, but I'm traveling until 1/29. I can handle it then, or someone else can dive in if they want to before then! I'll set a reminder for myself, just ping me if anyone gets there first.

bin/git-forgit
sandr01d: Works for me, thanks!

sandr01d (approved):
All set from my side, thanks for all the work @carlfriedrich!

cjappl:
> I prepared restricting fzf to version >= 0.49.0 for the AUR packages. @cjappl can you handle brew?

This is prepared for a PR and ready to go. When this review is submitted I will put it up for review in the homebrew repo. They are usually very prompt in approving and merging

carlfriedrich:
@sandr01d @cjappl Thanks for preparing AUR and brew. This will land in forgit 25.02.0, which will be automatically released on 2025-02-01.

cjappl:
Put up for review here for brew:

https://github.com/Homebrew/homebrew-core/pull/205866

PR #427, opened 2025-03-10
Meta: Do not tag release without public changes

### LINKED ISSUE #413: missing `forgit-24.12.0.tar.gz` for 24.12.0 tag

carlfriedrich:
This is mostly a refactoring of our GitHub actions. The tag workflow and script have been removed completely, since the release action will implicitly create the release tag if it does not exit, yet. This makes sure that we do not create a tag when there will be no release.

We have only one workflow now, which contains two jobs: one for generating the changelog and one for creating the release.

The changelog job will generate the changelog and upload it as an artifact if there are public changes. The release version, which previously was determined in the tag script, is now determined within this job, so that we can pass it to the release job.

The release job will run only if the changelog job was successful, i.e. if there are public changes. It will create a tag and a GitHub release with the given version number.

The workflow will run on the first day of every month for our regular automatic monthly releases.
It will also run on every tag push, so that we still can create a release manually if necessary. The release action will implicitly use the existing tag for the release then.

Fixes #413
Closes #414

## Check list

## Type of change

- [x] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [ ] zsh
    - [ ] fish
- OS
    - [ ] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:


carlfriedrich:
FYI: I have tested this new workflow in my fork of the repo.

.github/workflows/release.yaml
```diff
@@ -1,32 +1,78 @@
...
-      - name: Create Package
-        run: tar -czf forgit-${{github.ref_name}}.tar.gz --exclude LICENSE --exclude README.md *
-
       - name: Generate Changelog
         id: changelog
         run: .github/scripts/changelog.sh > CHANGELOG.md
+        # When the script returns with an error, it indicates that we don't have
+        # any public changes to be released. This is expected behavior, so we
+        # quit the workflow successfully in this case.
         continue-on-error: true
```
cjappl, line 27: The comment seems at odds with the message here 

> This is expected behavior, so we quit the workflow successfully in this case.

Wouldn't this imply `continue-on-error: false`?

.github/workflows/release.yaml
```diff
@@ -1,32 +1,78 @@
...
 
+      - name: Upload Changelog
+        uses: actions/upload-artifact@v4
+        if: steps.changelog.outcome == 'success'
+        with:
+          name: changelog
+          path: CHANGELOG.md
```
cjappl, line 34: Is the uploading and downloading the changelog necessary for not tagging without public changes?

If not, could we separate it into it's own PR? If it's necessary, all good to keep it here.

cjappl (commented):
Seems reasonable enough to me, few questions then LGTM

.github/workflows/release.yaml
sandr01d: I think this is fine. This ensures that the job does not fail ungracefully (sending out notifications to us etc.). Since the following steps all use `if: steps.changelog.outcome == 'success'` they're simply skipped over and the workflow ends successfully without doing any further steps.

carlfriedrich: What @sandr01d says. `continue-on-error: false` is the default behavior and would quit the workflow with an error, which is not what we want.

.github/workflows/release.yaml
carlfriedrich: Changelog creation and release publishing are two different jobs now. I find this useful to make visible at first glance that only the changelog has run and release was skipped in the GitHub actions view.
Uploading and downloading artifacts is the way to share files between jobs, so it is required for this change. Branching this out into a dedicated PR does not make sense to me, because we do the complete change for the single purpose of preventing the tag.

carlfriedrich:
@sandr01d @cjappl Thanks for your reviews!

PR #433, opened 2025-04-20
Meta: use latest ubuntu image for ci

sandr01d:
## Check list

## Description

GitHub has deprecated the `ubuntu-20.04` runnner as outlined [here](https://github.com/actions/runner-images/issues/11101). This PR updates the CI to always include the latest available Ubuntu image (currently `ubuntu-24.04`) and removes `ubuntu-20.04`.

We should probably merge this PR before anything else, to prevent the CI from failing when merging new PRs.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change
- [X] CI

## Test environment

- Shell
    - [ ] bash
    - [ ] zsh
    - [ ] fish
- OS
    - [ ] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-latest, ubuntu-20.04, ubuntu-22.04]
+        os: [macos-latest, ubuntu-latest]
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:


cjappl (approved):

wfxr (approved):
LGTM

carlfriedrich:
Thanks for the update @sandr01d. While we're at it: is there a reason why we're using two different Ubuntu versions for the CI run? For MacOS we're using only the latest version, so why not drop the static ubuntu-22.04 version as well and use only ubuntu-latest?

sandr01d:
> Thanks for the update @sandr01d. While we're at it: is there a reason why we're using two different Ubuntu versions for the CI run? For MacOS we're using only the latest version, so why not drop the static ubuntu-22.04 version as well and use only ubuntu-latest?

True, there is nothing in the CI that should differentiate between varying version of Ubuntu. I've moved ahead and removed `ubuntu-22.04`, keeping only `ubuntu-latest`.

carlfriedrich (approved):

PR #439, opened 2025-05-06
Fix: prefix all commands in forgit.plugin.zsh

### LINKED ISSUE #435: Seemingly broken when `grep` is aliased for default options
The loader script generates warnings about non-exported settings, despite the listed variables being marked for export. When I edit the file to prefix invocations of `grep` with the `command` builtin keyword, the warnings disappear. I have `grep` aliased to `grep --perl-regexp --initial-tab --binary-files=without-match --color=auto`, while the patterns used probably expect grep's default "basic" regex flavour. I suspect this problem will also occur within the `git-forgit` script itself if certain commands are aliased by the user.

sandr01d:
## Check list

## Description

Closes #435

Fish does not seem to be affected by this issue, so no changes necessary there.

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [X] bash
    - [X] zsh
    - [ ] fish
- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/forgit.plugin.zsh
+++ b/forgit.plugin.zsh
@@ -1,8 +1,11 @@
 #!/usr/bin/env bash
 # MIT (c) Wenxuan Zhang
 
-forgit::error() { printf "%b[Error]%b %s\n" '\e[0;31m' '\e[0m' "$@" >&2; return 1; }
-forgit::warn() { printf "%b[Warn]%b %s\n" '\e[0;33m' '\e[0m' "$@" >&2; }
+# all commands are prefixed with the command built-in. This prevents the wrong
+# commands getting executed in case a user added a shell alias with the same name.
+
+forgit::error() { command printf "%b[Error]%b %s\n" '\e[0;31m' '\e[0m' "$@" >&2; return 1; }
+forgit::warn() { command printf "%b[Warn]%b %s\n" '\e[0;33m' '\e[0m' "$@" >&2; }
 
 # determine installation path
 if [[ -n "$ZSH_VERSION" ]]; then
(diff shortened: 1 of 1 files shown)


carlfriedrich (changes requested):
Thanks for the fix @sandr01d!

I must confess, I actually don't like the `\<command>` syntax.

First (very subjective) argument is that I have never heard of this before. In contrast, using `command` as a prefix is a very common thing to do for me, even interactively, and in my perception this is the common way to do handle this (I might be wrong, though).

Second (kind of objective) argument is that people who don't know this syntax will have a hard time figuring out what it does, as this is difficult to google.

So I would still vote for prefixing everything with `command`.

A third approach (which I still like more than the backslash syntax) would be to call every command via its absolute path. The path could be found with `which`, so for example `printf` would become `$(which printf)`.
This would not work for shell built-ins, though (`set`, `unset`, `alias`, `export`). However, I would consider it very unlikely that someone has aliased a shell built-in. We're not escaping `if`, `then` or `while` either, so I would vote against escaping built-ins in general.

Either way I would vote for adding a describing comment to the plugin on why we do this.

sandr01d:
@carlfriedrich I've made changes accordingly. `command` does not support shell built-ins either, so I removed the prefix for those.

wfxr (approved):

carlfriedrich (approved):
Thanks a lot for the update @sandr01d. I like this a lot more! LGTM.

Issue #441, opened 2025-05-23
ZSH completions seem not to work with Oh-My-ZSH


wolfwood:
~~I see that there are completions for bash and fish, it would be great to have them for zsh as well.~~

I just found #419

wolfwood:
```shell
wolfwood@hotchip ~ $ for line in $fpath; do grep _git-add "$line/"**/*(-.); done
/home/wolfwood/.oh-my-zsh/custom/plugins/forgit/completions/_git-forgit:    add) _git-add ;;
/home/wolfwood/.oh-my-zsh/custom/plugins/forgit/completions/_git-forgit:(( $+functions[_git-add] )) || _git
/home/wolfwood/.oh-my-zsh/custom/plugins/forgit/completions/_git-forgit:(( $+functions[_git-add] )) || return 1
/home/wolfwood/.oh-my-zsh/custom/plugins/forgit/completions/_git-forgit:compdef _git-add forgit::add
zsh: no matches found: /home/wolfwood/.oh-my-zsh/functions/**/*(-.)
zsh: no matches found: /home/wolfwood/.oh-my-zsh/completions/**/*(-.)
zsh: no matches found: /home/wolfwood/.oh-my-zsh/custom/functions/**/*(-.)
zsh: no matches found: /home/wolfwood/.oh-my-zsh/custom/completions/**/*(-.)
zsh: no matches found: /home/wolfwood/.oh-my-zsh/cache/completions/**/*(-.)
zsh: no matches found: /usr/local/share/zsh/site-functions/**/*(-.)
/usr/share/zsh/5.9/functions/Completion/Unix/_git:(( $+functions[_git-add] )) ||
/usr/share/zsh/5.9/functions/Completion/Unix/_git:_git-add () {
/usr/share/zsh/5.9/functions/Completion/Unix/_git:(( $+functions[_git-add] )) ||
/usr/share/zsh/5.9/functions/Completion/Unix/_git:_git-add () {
```

wolfwood:
I would expect, for example, `glo <tab>` to complete with branch names, but instead I just get files.

wolfwood:
```shell
wolfwood@hotchip ~/repos/tryadactyl (try-1*?) $ which _git
_git () {
	# undefined
	builtin autoload -XUz
}
```

```git forgit <tab>``` doesn't work either

sandr01d:
[...]

Issue #419, opened 2025-01-17
is it possible to get branch completions when writing 'gd'?


yuvals1:

sandr01d:
We essentially make use of the completions provided by git, which include branches for `diff`. Which shell are you using? If you're using fish, this might be related to #418.

yuvals1:
how exactly is it supposed to work?
i have this line in my zshrc:
`[ -f $HOMEBREW_PREFIX/share/forgit/forgit.plugin.zsh ] && source $HOMEBREW_PREFIX/share/forgit/forgit.plugin.zsh`

but unless i explicitly add `compdef _git_diff forgit::diff` (for `gd` for example), the completions don't work.

i also have `zinit snippet OMZP::git`. maybe it interferes with forgit completions somehow?

sandr01d:
I'm assuming you're using zsh and installed fzf through brew from your previous response. Is that correct?

> how exactly are forgit completions supposed to work?

The only step necessary for working completions with zsh is that [completions/\_git-forgit](https://github.com/wfxr/forgit/blob/main/completions/_git-forgit) has to be placed in a directory that is in your `$fpath`. The brew installation script handles that for you. completions/\_git-forgit handles the necessary compdef bindings.

To make sure that completions/\_git-forgit is present in you `$fpath` could you please run the following in your zsh and send me the results?

```sh
for line in $fpath; do grep _git-add "$line/"**/*(-.); done
```

> i also have zinit snippet OMZP::git. maybe it interferes with forgit completions somehow?

We had a user that had similar issues with zinit previously. Please take a look at #330 and see if something in there helps.

yuvals1:
```
❯ for line in $fpath; do grep _git-add "$line/"**/*(-.); done
/usr/local/share/zsh/site-functions/_git-forgit:    add) _git-add ;;
/usr/local/share/zsh/site-functions/_git-forgit:(( $+functions[_git-add] )) || _git
/usr/local/share/zsh/site-functions/_git-forgit:(( $+functions[_git-add] )) || return 1
/usr/local/share/zsh/site-functions/_git-forgit:compdef _git-add forgit::add
/usr/local/share/zsh/site-functions/_git-forgit:    add) _git-add ;;
/usr/local/share/zsh/site-functions/_git-forgit:(( $+functions[_git-add] )) || _git
/usr/local/share/zsh/site-functions/_git-forgit:(( $+functions[_git-add] )) || return 1
[...]

PR #445, opened 2025-05-29
Docs: improve completion docs and move them to the top

sandr01d:
## Check list

## Description

There have been a few issues opened recently regarding completions (e.g., #419, #441). I see two common things that are causing issues that we could improve:

* Users not being aware when manual configuration is necessary
* Users not being aware of our docs regarding completions at all

This PR tries to solve both by adding documentation when configuration is necessary and moving the completion section to up in the README. They are currently at the bottom of the README, where a lot of people probably don't find them and having them next the installation section makes sense to me.
While I was at it, I've also added instructions for setting up completions with fish.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [X] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [ ] zsh
    - [ ] fish
- OS
    - [ ] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:


wfxr (approved):
LGTM

PR #450, opened 2025-07-02
Feat: Use forgit show in stash_show enter

carlfriedrich:
Instead of using `git stash show` in the `enter` binding of the `stash_show` command, use `forgit show`. This allows forgit to show the stash in the same interactive way it shows other commits (e.g. on `forgit log`), with the same context and formatting options.

## Check list

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -525,24 +525,26 @@ _forgit_reset_head() {
 
 _forgit_stash_show_preview() {
     local stash
-    stash=$(echo "$1" | cut -d: -f1)
-    _forgit_git_stash_show "$stash" | _forgit_pager diff
+    stash=$(echo "$1" | _forgit_extract_stash_name)
+    echo "$stash" | xargs -I% git show --color=always -U"$_forgit_preview_context" % | _forgit_pager show
 }
 
-_forgit_git_stash_show() {
-    git stash show --color=always --ext-diff "$@"
+_forgit_stash_show_enter() {
+    local stash
+    stash=$(echo "$1" | _forgit_extract_stash_name)
+    echo "$stash" | xargs -I% "${FORGIT}" show %
 }
 
 # git stash viewer
 _forgit_stash_show() {
     _forgit_inside_work_tree || return 1
     local opts
-    [[ $# -ne 0 ]] && { _forgit_git_stash_show "$@"; return $?; }
+    [[ $# -ne 0 ]] && { "${FORGIT}" show "$@"; return $?; }
     _forgit_stash_show_git_opts=()
     _forgit_parse_array _forgit_stash_show_git_opts "$FORGIT_STASH_SHOW_GIT_OPTS"
     opts="
         $FORGIT_FZF_DEFAULT_OPTS
-        +s +m -0 --tiebreak=index --bind=\"enter:execute($FORGIT stash_show_preview {} | $FORGIT pager enter)\"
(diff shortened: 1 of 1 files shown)


bin/git-forgit
```diff
@@ -525,24 +525,28 @@ _forgit_reset_head() {
 
 _forgit_stash_show_preview() {
     local stash
-    stash=$(echo "$1" | cut -d: -f1)
-    _forgit_git_stash_show "$stash" | _forgit_pager diff
+    stash=$(echo "$1" | _forgit_extract_stash_name)
+    shift
+    echo "$stash" | xargs -I% git show --color=always -U"$_forgit_preview_context" % -- "$@" | _forgit_pager show
```
sandr01d: Is `xargs` really needed here? Seems to be working fine without it...
From what I can see, there will also always be only a single argument passed to this function, so `-- "$@"` would not do anything or am I missing something?
This version seems to have the same behavior.
```bash
git show --color=always -U"$_forgit_preview_context" "$stash" | _forgit_pager show
```

bin/git-forgit
```diff
@@ -525,24 +525,28 @@ _forgit_reset_head() {
...
 }
 
-_forgit_git_stash_show() {
-    git stash show --color=always --ext-diff "$@"
+_forgit_stash_show_enter() {
+    local stash
+    stash=$(echo "$1" | _forgit_extract_stash_name)
+    shift
+    echo "$stash" | xargs -I% "${FORGIT}" show % "$@"
```
sandr01d: Same questions regarding `xargs` and the additional arguments as noted above with `_forgit_stash_show_preview`.

sandr01d (changes requested):
Very nice improvement! Was considering adding something like this just recently. Just two question from my side.

bin/git-forgit
carlfriedrich: @sandr01d Thanks for your review. That is a really good question indeed. I actually copied this from `_forgit_log_preview`, where we use [the exact same command sequence including `xargs`](https://github.com/wfxr/forgit/blob/main/bin/git-forgit#L193). I assume that it is obsolete there as well then?
At least I have no idea why it is there, and I agree that it shouldn't be necessary. I will add a second commit removing the obsolete `xargs`.

carlfriedrich: Okay, there were actually some more places where we used this random `xargs` expression. I cleaned up all of them and rewrote them without `xargs`.
@sandr01d Can you have a look? Should we extract this into a dedicated PR?

sandr01d: Thanks for the update. I think you might have missed my comment regarding the function arguments that are passed to `git show` with `-- "$@"`. I think this is not need either.
I'm fine with keeping the cleanup in this PR, as long as we don't squash the commits.

carlfriedrich: @sandr01d Thanks for getting back to this. I indeed missed your comment regarding `-- "$@"`, and you are totally right: it is not needed, since we do not pass any files to the function. This was also a copy&paste relic from `_forgit_log_preview`. Good catch, thanks for reporting. I removed the part from the commits and also checked all other occurences, but it is actually needed everywhere else.

sandr01d (approved):
Thanks for the update again, looks good to me now.

PR #459, opened 2025-07-18
Fix: files in untracked directories are not shown in git add

### LINKED ISSUE #458: Files in untracked directories broken in `git add`
Turns out the `-u` option we removed in fd373ba6c2081f7fedce77d332014b20fb5af201 actually did have a purpose. Without `-u`, `git status` does not show files in untracked directories so we don't get these as options anymore (only the entire directory) and the preview for an untracked directory now displays something like `error: Could not access 'new_dir/null'`.

From the git manual:

> The mode parameter is used to specify the handling of untracked files. It is
> optional: it defaults to all, and if specified, it must be stuck to the
> option (e.g.  -uno, but not -u no).
>
> The possible options are:
>
> •   no - Show no untracked files.
>
> •   normal - Shows untracked files and directories.
>
> •   all - Also shows individual files in untracked directories.
>
> **When -u option is not used, untracked files and directories are shown (i.e.
> the same as specifying normal), to help you avoid forgetting to add newly
> created files.**

I thought about reading the config file with `git config status.showUntrackedFiles`, but unfortunately this is not so easy, because

> All usual spellings for Boolean value true are taken as normal and false as no.

So, all of these would be valid configurations

```
status.showUntrackedFiles = no # mode: no
status.showUntrackedFiles = yes # mode: normal
status.showUntrackedFiles = false # mode: no
status.showUntrackedFiles = 0 # mode: no
status.showUntrackedFiles = 2 # mode: normal
```

I currently see no other option than reverting fd373ba6c2081f7fedce77d332014b20fb5af201, but wanted to check in with you first in the hope of there being a better solution.

sandr01d:
## Check list

## Description

Thanks @carlfriedrich for your insights provided in https://github.com/wfxr/forgit/issues/458#issuecommentj3089130866. I've implemented this as suggested, including the preview for normal mode, for which I simply used our existing directory view:

Pass the value of `git config status.showUntrackedFiles` to the `git status` command we run in `_forgitladd`, defaulting to `all` when unset.
Additionally correctly handle directories in `forgit_add_preview` for the case when `status.showUntracked` is explicitly set to `normal`.

You can see the different modes in action here:

<a href="https://asciinema.org/a/iO5purJ2PFkRjcl5CoBwILSXj" target="_blank"></a>

## Type of change

- [X] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [X] zsh
    - [ ] fish
- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

Fixes #458

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -430,6 +430,10 @@ _forgit_show() {
 
 _forgit_add_preview() {
     file=$(echo "$1" | _forgit_get_single_file_from_add_line)
+    # $file can be a directory when status.showUntrackedFiles is explicitly set to 'normal'
+    # In this case, show the content of the directory and return
+    [ -d "$file" ] && eval "$_forgit_dir_view \"$file\"" && return 0
+
     if (git status -s -- "$file" | grep '^??') &>/dev/null; then  # diff with /dev/null for untracked files
         git diff --color=always --no-index -- /dev/null "$file" | _forgit_pager diff | sed '2 s/added:/untracked:/'
     else
@@ -460,7 +464,7 @@ _forgit_edit_add_file() {
 # git add selector
 _forgit_add() {
     _forgit_inside_work_tree || return 1
-    local changed unmerged untracked files opts
+    local changed unmerged untracked files opts show_untracked
     # Add files if passed as arguments
     _forgit_contains_non_flags "$@" && { _forgit_git_add "$@" && git status -s; return $?; }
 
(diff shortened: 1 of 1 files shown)


carlfriedrich:
@sandr01d Thanks for the quick implementation. Works great for me! Gonna merge this directly.

PR #464, opened 2025-09-01
Meta: Add bashunit tests to CI workflow and implement a first test case

### LINKED ISSUE #363: Add unit testing framework
We have had a decent number of regressions and issues between shell/OS versions.

It would be great to set up a unit testing framework and eventually github actions to test that things are staying consistent and prevent further regressions.

Things like
- Files with spaces
- Files with brackets
- Files not previewing on one operating system

Have all been fixed/broken/rebroken. 

@sandr01d found this, which seems pretty promising! https://github.com/kward/shunit2

I also think this may be easier once #326 lands, as it breaks the code into a bunch of small simple functions ripe for unit testing.

carlfriedrich:
This PR is an attempt to close #363. It adds the bash unit test framework [`bashunit`](https://github.com/TypedDevs/bashunit) to the CI pipeline, along with a first simple test file.

Some background information: for another project I had to evaluate bash unit test frameworks (which was easier than I thought, as someone already created a really good in-depth comparison [here](https://github.com/dodie/testing-in-bash), which I took as a starting point). `bashunit` was the clear winner. It is feature-rich and yet simple, user-friendly and has a really good [documentation](https://bashunit.typeddevs.com/quickstart).

## Check list

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Test
- [ ] Documentation change

## Test environment

- Shell
    - [x] bash
    - [ ] zsh
    - [ ] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:


cjappl (approved):
Awesome!

carlfriedrich:
@cjappl @sandr01d @wfxr Looking forward to your opinions on this.

*question*: I updated the PR template to include a checkbox for having implemented test cases. Should we add some documentation about how to do this somewhere? If so, where would be a good place for this? We do not have a "Contributing" section or file anywhere, yet, but this might seem like the place for stuff like this. WDYT?

bin/git-forgit
```diff
@@ -1243,6 +1243,11 @@ private_commands=(
     "pager"
 )
 
+# Check if the script is being sourced
```
sandr01d: I think we should mention that this is necessary for the unit tests in the comment.

sandr01d (changes requested):
Looks very neat!

> question: I updated the PR template to include a checkbox for having implemented test cases. Should we add some documentation about how to do this somewhere? If so, where would be a good place for this? We do not have a "Contributing" section or file anywhere, yet, but this might seem like the place for stuff like this. WDYT?

Forgot to reply to this earlier. Yes, good idea. Adding a [CONTRIBUTING.md](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors#about-contributing-guidelines) file would be best IMO.

> We already have extracted lots of code into this kind of small helper functions, and we should continue to do so in order to make them unit-testable.

Yes, absolutely :rocket: 

> Let me know if this example test is convincing enough for you in order to move forward with bashunit. If you have already experience with another bash testing framework, though, I would like to hear your opinion on that and what might be better than with bashunit.

You convinced me. I looked into [bats-core](https://github.com/bats-core/bats-core) previously, but it falls short to bashunit IMO, especially in regards to those really handy mocking features.

Once the (very minor) change I've proposed is addressed I think we could go ahead and merge this PR. It's probably best to introduce the rest of the unit tests in several smaller PRs. WDYT?

tests/helper-functions.test.sh
```diff
@@ -0,0 +1,34 @@
...
+    for i in "${!expected[@]}"; do
+        echo "${actual[i]}"
+        assert_same "${expected[i]}" "${actual[i]}"
+    done
+}
+
+function provider_diff_lines() {
+    data_set "[A]     newfile" "newfile"
+    data_set "[D]     oldfile with spaces" "oldfile with spaces"
+    data_set "[R100]  file  ->  another file" "file" "another file"
```
sandr01d, line 31: We should probably add a few extra data sets for esoteric file names. Files with backslashes or line breaks come to my mind, e.g. see #388 and #465.

bin/git-forgit
carlfriedrich: Good catch, I have updated that.

tests/helper-functions.test.sh
carlfriedrich: Good idea! I have added a testcase with a tab-containing file in its escaped form, just like it is output from `git diff --name-status` (which is the input to `_forgit_get_files_from_diff_line`, after being manipulated slightly, but that would be a different unit test). The unit test for this function passes. However, I noticed that the preview window and on pressing enter the file diff is not shown (most probably due to what you already said in #465 concerning git's output and input formats). This should not be handled by this function, though, so IMO this unit test is complete.

sandr01d: Thanks for the update. Could we add a file name with a literal backslash in it too? I think this would be a perfect candidate to prevent future regressions.

carlfriedrich: We could totally do that. It's not the expected (and actual) type of input for this particular function, though, because it consumes output of `git status`, and that prints, as you already stated, escaped file names. That's why I don't think that's a useful test case for this function.
For other functions we will definitely need literal tabs in the test data, but here I don't see why that would make sense. Maybe this function was not a good choice for the first test case. :-/

sandr01d: Ah, makes sense, no need to add it then. Thanks for the explanation :+1:

sandr01d (approved):
I'd say that this is ready to be merged. Thank you @carlfriedrich!

PR #465, opened 2025-09-06
Fix: preview for add and clean when file names contain backslashes

sandr01d:
## Check list

## Description

While playing around with #464, I noticed that the preview of `forgit add` and `forgit clean` do not work with files that have backslashes in their names. The preview just shows something like

> error: Could not access 'file\\.txt'

This is due to the same issue we tried to address with #388 (git escapes special characters in its output, but does not expect them to be escaped when used for input). Since #388 we use the `-z` option with most git commands that have this "issue" to prevent them from escaping their output. In `forgit add` and `forgit clean` we use the `core.quotePath false` config option instead. In `forgit add` because we make use of the colors of the output of `git status` and in `forgit clean` because `git clean` does not have a `-z` option.
I noticed that (in contrast to my previous belief), `-z` does not remove the colors from `git status`, so it can be used in `forgit add`. However, I could not find a way to get things working with our current implementation, but I found a simpler approach to what we're trying to achieve that actually does work. We try to filter out added files using `grep` by matching lines with changed, unmerged and untracked files. Instead we can just do an inverse match for added files.
For `forgit clean` I've modified the command to use `_forgit_list_files` instead of `git clean -n`. This comes with a minor breaking change: flag arguments provided to `forgit clean` are now passed to `git ls-files` instead of `git clean -n`. However, git ls-files is way more flexible than git clean, so all use cases covered by git clean (and more) should still be possible. This will even allow us to introduce an environment variable in the future to modify which files are included when cleaning (I'll explain my use case in another PR, this text is already way too long :wink:).

## Type of change

- [X] Bug fix
- [ ] New feature
- [X] Refactor
- [ ] Breaking change
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [X] zsh
    - [ ] fish
- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -464,13 +464,11 @@ _forgit_edit_add_file() {
 # git add selector
 _forgit_add() {
     _forgit_inside_work_tree || return 1
-    local changed unmerged untracked files opts show_untracked
+    local added files opts show_untracked
     # Add files if passed as arguments
     _forgit_contains_non_flags "$@" && { _forgit_git_add "$@" && git status -s; return $?; }
 
-    changed=$(git config --get-color color.status.changed red)
-    unmerged=$(git config --get-color color.status.unmerged red)
-    untracked=$(git config --get-color color.status.untracked red)
+    added=$(git config --get-color color.status.added green)
     opts="
         $FORGIT_FZF_DEFAULT_OPTS
         -0 -m --nth 2..,..
@@ -482,8 +480,9 @@ _forgit_add() {
     show_untracked=$(git config status.showUntrackedFiles)
     while IFS='' read -r file; do
         files+=("$file")
-    done < <(git -c color.status=always -c status.relativePaths=true -c core.quotePath=false status -s --untracked="${show_untracked:-all}" |
-        grep -F -e "$changed" -e "$unmerged" -e "$untracked" |
+    done < <(git -c color.status=always -c status.relativePaths=true status -s -z --untracked="${show_untracked:-all}" |
+        tr '\0' '\n' |
+        grep -F --invert-match "$added" |
         sed -E 's/^(..[^[:space:]]*)[[:space:]]+(.*)$/[\1]  \2/' |
         FZF_DEFAULT_OPTS="$opts" fzf |
         _forgit_get_single_file_from_add_line)
@@ -634,7 +633,7 @@ _forgit_clean() {
         $FORGIT_CLEAN_FZF_OPTS
     "
     # Note: Postfix '/' in directory path should be removed. Otherwise the directory itself will not be removed.
-    files=$(git -c core.quotePath=false clean -xdffn "$@"| sed 's/^Would remove //' | FZF_DEFAULT_OPTS="$opts" fzf |sed 's#/$##')
+    files=$(_forgit_list_files --others "$@" | sed 's/^Would remove //' | FZF_DEFAULT_OPTS="$opts" fzf |sed 's#/$##')
     [[ -n "$files" ]] && echo "$files" | tr '\n' '\0' | xargs -0 -I% git clean "${_forgit_clean_git_opts[@]}" -xdff '%' && git status --short && return
     echo 'Nothing to clean.'
 }


sandr01d:
> I would like to wait until https://github.com/wfxr/forgit/pull/464 is merged and then approach this in a TDD-way,

Yes, makes sense

sandr01d:
Further testing showed that my attempt to fix `git add` was not feasible. Inversely matching for the color of added files is not an option because the color will also be present for files that have been partially staged, which would result in such files being filtered out. I don't know yet how we can fix this. To keep things going, I've removed the code in question from this PR and created #467 for it, so we can focus on the fix for `git clean` here first.
@carlfriedrich Coming up with unit tests for the remaining fix is tricky. `forgit_list_files` requires actual files to be present. We could mock the `git ls-files` part away, but this would result in us mostly testing the mock IMO. On top of that, adding a test for `forgit_list_files` with this PR would not be a proper regression test, because it wouldn't test that the behavior stays the same as with our previous implementation using `git clean -n`.

carlfriedrich:
@sandr01d Thanks for keeping this going and the good issue analysis. Good idea to split this into two separate issues!

The issue about coming up with a useful unit test here is that we don't actually have a testable unit. IMO that's the main advantage of introducing a unit test system: that we have to think more about how to split the code into testable units, which makes the code more readable and robust in the end.

This is the original line:

```bash
files=$(git -c core.quotePath=false clean -xdffn "$@"| sed 's/^Would remove //' | FZF_DEFAULT_OPTS="$opts" fzf |sed 's#/$##')
```

In testable units this could be refactored to:

```bash
files=$(_forgit_list_clean_files | FZF_DEFAULT_OPTS="$opts" fzf | _forgit_strip_trailing_slash)
```

The two new functions `_forgit_list_clean_files` and `_forgit_strip_trailing_slash` can then be tested individually.

So a useful TDD approach would be:

1. Refactor the code to have testable units
2. Write a failing unit test
3. Fix the code
4. See the unit test pass

In addition to unit tests, in this case it would also make sense to have integration tests. `bashunit` supports creating temporary directories via `temp_dir()` function during `set_up` or `set_up_before_script` functions, or in the test function itself. An integration test function could create a temporaty directory, call `git init` and create some files and commits in there. Then we can call `_forgit_list_clean_files` on an actual git tree (without mocking anything) and feed the output to `_forgit_clean_preview` (which is the chain which shows the issue).

IMO this fix is not urgent, as it is a rare and unlikely corner case, so this is a good bug candidate to start coming up with good test cases.

sandr01d:
> Providing reusable code is not the only reason for introducing a function.

I know, no need to be petty.

> I would argue for the exact opposite. _forgit_list_clean_files and _forgit_strip_trailing_slash are really self-explaining names which make perfectly readable code IMO. A code base consisting of these kind of functions do not need any comments or further explaination. These function names are understandable at the very first sight, even for people who have never implemented anyhting in bash script before.

Easier to read, yes, but harder to understand IMO. What I mean is that splitting the code out into such small units requires you to do constant context switches when trying to build a mental model of what the code precisely does (not on an abstract level). It makes it harder to see at a glance how the different commands interact with each other.  My concern isn't for this single code change, but more of a general one: once we start splitting each line up into multiple functions, this would make navigating the code base harder IMO.

I tried for several hours now to introduce a test, but could not find a way to do so without changing the code for the worse. I think it would be best to just move forward without it.

tests/clean-select-files.test.sh
```diff
@@ -0,0 +1,36 @@
...
+    dir=$(temp_dir)
+    cd "$dir" || return 1
+    git init --quiet
+
+    # create files to test against
+    touch file.txt
+    touch file_with\\backslashes\\.txt
+    touch "file with spaces.txt"
+    touch "file_with\ttab.txt"
+}
```
carlfriedrich: I think setting up this git repo with the files has to be done only once for all four test runs, so it can go into `set_up_before_script` (`set_up` is executed before *each* executed test function, which is four times in this case, as we have four data sets).

sandr01d: I tried, but changing the directory in `set_up_before_script` for some reason breaks the data providers. I've created an [upstream issue](https://github.com/TypedDevs/bashunit/issues/529) and will update this PR once I get a response.

Chemaclass: I will take a look into it

Chemaclass: Fix done https://github.com/TypedDevs/bashunit/pull/531 Already available in the [beta](https://bashunit.typeddevs.com/installation#define-custom-tag-and-folder) 🏗️ 
Feel free to test it and let me know your feedback! I plan to create a new release this weekend or next week.

sandr01d: Thanks for the quick fix @Chemaclass! :rocket:  Can confirm that the test now succeeds, but it seems subsequent tests do not run. This is not a regression btw. I've opened another [issue](https://github.com/TypedDevs/bashunit/issues/532).

Chemaclass: @sandr01d there you go https://github.com/TypedDevs/bashunit/pull/533

sandr01d: Massive thanks for the quick fixes @Chemaclass! Can confirm this is now working nicely.

Issue #447, opened 2025-06-11
Add support for git restore


LittleHaku:
It would be very nice to have the option of `git restore` with fuzzy find, since some times you modify a file and then decide that before it was better, so an option with `git restore` like `grs` would be very handy.

sandr01d:
This sounds very similar to what we already provide with `gcf` (`git checkout file`). Could you please explain what `grs` would provide that `gcf` does not?

sandr01d:
> Also restore lets you restore to the index instead of HEAD.
>
> So case you have a file that says Hello then you modify it and says Hello world, you add it (stage it) and modify it again to Hello world!, here restore would take it back to Hello world, checkout would take it back to HEAD, i.e., Hello.

`checkout` also restores from the index and not HEAD. From the `git checkout` man page:

>  checkout [-f|--ours|--theirs|-m|--conflict=<style>] \[<tree-ish\>] [--] <pathspec>..., git checkout [-f|--ours|--theirs|-m|--conflict=\<style\>] [\<tree-ish\>] --pathspec-from-file=\<file\> [--pathspec-file-nul]
>
>Overwrite the contents of the files that match the pathspec. When the \<tree-ish\> (most often a commit) is not given, overwrite working tree with the contents in the index. When the \<tree-ish\> is given, overwrite both the index and the working tree with the contents at the \<tree-ish\>.

AFAIK the functionality of `checkout` got split up into `switch` (`git checkout <branch>`) and `restore` (`git checkout <pathspec>`) with git 2.23, but those new commands serve the same purpose and behave the same as the corresponding functionality of `checkout` (see [this](https://stackoverflow.com/questions/61130412/what-is-the-difference-between-git-checkout-vs-git-restore-for-reverting-un)).



`gcf` got implemented (0d34ff5) before `git restore` was introduced in git, which is why it is using `checkout`). I'm open to have a discussion on whether we should transition `gcf` to use `restore` (and `gcb` to use `switch`) instead of `checkout`, but I don't see any benefit of having a function for `restore` _and_ `checkout`.

LittleHaku:
@carlfriedrich I think it makes sense to deprecate `gcb` too.

@sandr01d What do you think about it?

If I can I would like to try and help, I have never participated in a big open source software.

sandr01d:
[...]

PR #472, opened 2025-11-17
Implemented `gsw` so it uses `git switch`

LittleHaku:
As discussed #447, we should use the new `git switch` and `git restore` instead of using `git checkout` for everything.

In this PR I implemented the `gsw` with `git switch` and `gcb` now calls this new function and echoes a deprecated warning.

I haven't done the git restore yet since I want to be sure that this is properly done before going for a new one.

## Check list

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [x] Breaking change
- [ ] Test
- [ ] Documentation change

## Test environment

- Shell
    - [x] bash
    - [x] zsh
    - [ ] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -855,12 +855,6 @@ _forgit_git_checkout_branch() {
     git checkout "${_forgit_checkout_branch_git_opts[@]}" "$@"
 }
 
-_forgit_git_switch_branch() {
-    _forgit_switch_branch_git_opts=()
-    _forgit_parse_array _forgit_switch_branch_git_opts "$FORGIT_SWITCH_BRANCH_GIT_OPTS"
-    git switch "${_forgit_switch_branch_git_opts[@]}" "$@"
-}
-
 # git checkout-branch selector
 _forgit_checkout_branch() {
     _forgit_inside_work_tree || return 1
@@ -900,6 +894,12 @@ _forgit_checkout_branch() {
     else
         _forgit_git_checkout_branch "$branch"
     fi
+}
+
+_forgit_git_switch_branch() {
+    _forgit_switch_branch_git_opts=()
+    _forgit_parse_array _forgit_switch_branch_git_opts "$FORGIT_SWITCH_BRANCH_GIT_OPTS"
+    git switch "${_forgit_switch_branch_git_opts[@]}" "$@"
 }
 
 _forgit_switch_branch() {


bin/git-forgit
```diff
@@ -849,14 +849,18 @@ _forgit_checkout_file() {
...
+_forgit_git_switch_branch() {
+    _forgit_switch_branch_git_opts=()
+    _forgit_parse_array _forgit_switch_branch_git_opts "$FORGIT_SWITCH_BRANCH_GIT_OPTS"
+    git switch "${_forgit_switch_branch_git_opts[@]}" "$@"
 }
 
-# git checkout-branch selector
 _forgit_checkout_branch() {
+    _forgit_switch_branch "$@"
+    echo "forgit: 'gcb' is deprecated and will be removed in a future release. Please use 'gsw' instead." >&2
```
cjappl: A couple thoughts here

1. What is the burden of just keeping this as a wrapper/alias for a while (or forever?). It seems like a relatively low maintenance burden on our end.
2. Could we minimally add an environment var to mute this if it is bugging people? `FORGIT_SILENCE_DEPRECATION_WARNINGS` and document that?


I don't necessarily want to hold onto everything forever, but if the maintenance burden is low we can prevent breaking backwards compatibility/user muscle memory.

bin/git-forgit
```diff
@@ -849,14 +849,18 @@ _forgit_checkout_file() {
...
 }
 
-_forgit_git_checkout_branch() {
-    _forgit_checkout_branch_git_opts=()
-    _forgit_parse_array _forgit_checkout_branch_git_opts "$FORGIT_CHECKOUT_BRANCH_GIT_OPTS"
-    git checkout "${_forgit_checkout_branch_git_opts[@]}" "$@"
+_forgit_git_switch_branch() {
+    _forgit_switch_branch_git_opts=()
+    _forgit_parse_array _forgit_switch_branch_git_opts "$FORGIT_SWITCH_BRANCH_GIT_OPTS"
+    git switch "${_forgit_switch_branch_git_opts[@]}" "$@"
```
cjappl: Do we have a fallback for older versions of git? If this was released in 2.23, presumably this won't work before that. (that seems like 2019)

How concerned are we with supporting as many git versions as possible? I would expect we have at least one bug report from someone who's forgit "breaks" after this.

Do we have other features that are git version dependent? @sandr01d @carlfriedrich 


I think this goes with my question down the review: we could also just keep gcb and add gsw and nudge people in documentation, or check their git version and if they have it print a deprecation warning or similar.

bin/git-forgit
LittleHaku: > 1. What is the burden of just keeping this as a wrapper/alias for a while (or forever?). It seems like a relatively low maintenance burden on our end.

It was to follow what was discussed with @sandr01d on #447:

> I agree, deprecating gcb and gcf in favor of gsw and grs makes sense and so do the steps you proposed for transitioning. We might want to start print a deprecation warning from the old functions one or two releases before we finally remove them. I wouldn't expect most people to read our release notes...

---

> 2. Could we minimally add an environment var to mute this if it is bugging people? FORGIT_SILENCE_DEPRECATION_WARNINGS and document that?

Yes, I think this could be a nice idea.

cjappl: Sounds good. Sorry to be late to the party and re-open an old can of worms. 

I think we should minimally do number 2, but interested in any other thoughts from people on number 1. it's not something that is the end of the world either way, but generally I'm a fan of keeping old behavior unless it severely impedes new progress. Curious on other maintainers thoughts

sandr01d: What I meant with my comment regarding the removal of `gcb` is that I do not want to entire `_forgit_checkout_branch` function to stick around forever, which would result in two separate implementations for the same thing. Keeping the old alias is fine with me.

bin/git-forgit
```diff
@@ -849,14 +849,18 @@ _forgit_checkout_file() {
...
 }
 
-# git checkout-branch selector
```
sandr01d, line 858: This comment was probably removed by accident.

bin/git-forgit
```diff
@@ -875,27 +879,28 @@ _forgit_checkout_branch() {
...
     [[ -z "$branch" ]] && return 1
 
     # track the remote branch if possible
     if [[ "$branch" == "remotes/"* ]]; then
         if git branch | grep -qw "${branch#remotes/*/}"; then
             # hack to force creating a new branch which tracks the remote if a local branch already exists
-            _forgit_git_checkout_branch -b "track/${branch#remotes/*/}" --track "$branch"
-        elif ! _forgit_git_checkout_branch --track "$branch" 2>/dev/null; then
-            _forgit_git_checkout_branch "$branch"
+            _forgit_git_switch_branch -ct "track/${branch#remotes/*/}" --track "$branch"
```
sandr01d: This does currently not work as intended. You're passing in `-t` and `--track` which is the same flag. With the current implementation I get: "fatal: only one reference expected. " Also, I'd prefer to use the long variant for `--create/-c`. This makes things easier to read.

```suggestion
            _forgit_git_switch_branch --create "track/${branch#remotes/*/}" --track "$branch"
```

sandr01d (changes requested):
Thanks for the PR @LittleHaku! Looks mostly good, just a few things to address (see my other comments). I noticed that the completions for fish are currently missing.

bin/git-forgit
sandr01d: Yes, I was under the impression that we'd keep the `checkout_branch` implementation as is for now and remove it later. That being said I wouldn't be too concerned about people using git versions older than 2.23. Even the oldest version of Debian that is still supported (Bullseye LTS) already ships with git 2.30.

LittleHaku: So do keep it as is? Or do we take back the git checkout branch?

bin/git-forgit
LittleHaku: yes, just re-added it

bin/git-forgit
```diff
@@ -875,27 +880,28 @@ _forgit_checkout_branch() {
...
+        elif ! _forgit_git_switch_branch --track "$branch" 2>/dev/null; then
+            _forgit_git_switch_branch "$branch"
         fi
     else
-        _forgit_git_checkout_branch "$branch"
+        _forgit_git_switch_branch "$branch"
     fi
 }
 
+
```
sandr01d: Please remove the extra blank line here.

bin/git-forgit
sandr01d: I think keeping git checkout branch for now would be best, but would be interested to hear what the other maintainers think before we move forward @cjappl @carlfriedrich @wfxr

cjappl: Yes, I am generally in favor of a long deprecation process. God knows someone will complain about it even if we rip it out in 2035 :)

bin/git-forgit
LittleHaku: done

LittleHaku:
> Thanks for the PR @LittleHaku! Looks mostly good, just a few things to address (see my other comments). I noticed that the completions for fish are currently missing.

I already added them :)

LittleHaku:
> > > Thanks for the PR @LittleHaku! Looks mostly good, just a few things to address (see my other comments). I noticed that the completions for fish are currently missing.
> > 
> > 
> > I already added them :)
> 
> Thanks for keeping things going @LittleHaku. Since nobody complained so far, I'd say we bring back the old `gcb` for now and merge this PR soon. Removing it later is easy.

Just added the original `gcb` back @sandr01d

bin/git-forgit
carlfriedrich: Thanks for pushing this forward everybody! I would also vote for a long deprecation process. We should not remove the function before `git` itself removes it (and that can be long). So being able to silence these functions via a variable gets a thumbs up from me!

carlfriedrich: And I would not be concerned about old git versions as well. Of course we could add some checking, but as long as no one complains, i would say YAGNI.

bin/git-forgit
```diff
@@ -855,6 +855,12 @@ _forgit_git_checkout_branch() {
     git checkout "${_forgit_checkout_branch_git_opts[@]}" "$@"
 }
 
+_forgit_git_switch_branch() {
```
sandr01d: Just one final touch from my side: could you please move this function down below `_forgit_checkout_branch`, so that it is next to `_forgit_switch_branch`?

LittleHaku: done

Issue #399, opened 2024-09-16
Feature Request: Add git worktree switcher like gcb 


deekshithanand:
## Check list

- [x] I have read through the [README](https://github.com/wfxr/forgit/blob/master/README.md)
- [x] I have the latest version of forgit
- [x] I have searched through the existing issues

## Environment info

- OS
    - [ ] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Others:
- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish

## Feature Request:

Add a work tree switcher like gcb , leveraging fzf.

carlfriedrich:
Thanks for your idea. I am not sure how the interface is supposed to look like. In `forgit` we only have wrappers for existing git functionality up to now, and there is no `git worktree switch` command or anything similar. Can you describe how your usecase looks like and what exactly you would like to achieve? Is it only about `cd`ing into a different work tree?

suft:
@deekshithanand what do you think of the following abbreviations?

All Possible Worktree Operations (https://git-scm.com/docs/git-worktree):

- `git worktree add`
- `git worktree list`
  + `cd` + interactive as `gwj` (`git worktree jump` - not a native git operation) 
- `git worktree lock`
  + interactive as `gwl`
- `git worktree move`
- `git worktree prune`
- `git worktree remove`
  + interactive `as gwr`
- `git worktree repair`
- `git worktree unlock`
  + interactive `as gwu`
  
I have only implemented `gwj`, `gwl`, `gwr`, and `gwu`.

deekshithanand:
Awesome!

iloveitaly:
Right now `gcb` shows worktrees, but fails to switch to them

![Image](https://github.com/user-attachments/assets/bd384163-c083-4698-bb49-3c272d576da8)

suft:
@iloveitaly Can you post a screenshot of what you get when you run `git worktree list` in the same repo as the gif above?

PR #474, opened 2025-11-26
exclude worktree branches

iloveitaly:
## Check list

## Description

If a branch is used by a git worktree it cannot be switched to and is 
prefixed with a `+`. This excludes these branches from `gcb` to avoid
erroring out when they are selected.

Related to #399

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Test
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish
- OS
    - [ ] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Others:

### DIFF
--- a/tests/worktree_exclude.test.sh
+++ b/tests/worktree_exclude.test.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+
+function set_up() {
+    source bin/git-forgit
+}
+
+function test_forgit_worktree_filter_excludes_plus_lines_by_default() {
+    local input="+ worktree-branch
+* main-branch
+  other-branch"
+    
+    # Default behavior (true)
+    local actual
+    actual=$(echo "$input" | _forgit_worktree_filter)
+    local expected="* main-branch
+  other-branch"
+
+    assert_same "$expected" "$actual"
+}
+
+function test_forgit_worktree_filter_excludes_plus_lines_when_explicitly_enabled() {
+    export FORGIT_CHECKOUT_BRANCH_EXCLUDE_WORKTREES="true"
+    local input="+ worktree-branch
+* main-branch"
+    
+    local actual
+    actual=$(echo "$input" | _forgit_worktree_filter)
+    local expected="* main-branch"
+
+    assert_same "$expected" "$actual"
+    unset FORGIT_CHECKOUT_BRANCH_EXCLUDE_WORKTREES
+}
+
+function test_forgit_worktree_filter_includes_plus_lines_when_disabled() {
+    export FORGIT_CHECKOUT_BRANCH_EXCLUDE_WORKTREES="false"
+    local input="+ worktree-branch
+* main-branch"
+    
+    local actual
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -384,9 +384,10 @@ export FORGIT_LOG_FZF_OPTS='
...
-| `FORGIT_LOG_FORMAT`         | git log format                            | `%C(auto)%h%d %s %C(black)%C(bold)%cr%Creset` |
+| Option                                     | Description                                   | Default                                       |
+|--------------------------------------------|-----------------------------------------------|-----------------------------------------------|
+| `FORGIT_CHECKOUT_BRANCH_EXCLUDE_WORKTREES` | exclude worktree branches in `gcb`            | `true`                                        |
+| `FORGIT_LOG_FORMAT`                        | git log format                                | `%C(auto)%h%d %s %C(black)%C(bold)%cr%Creset` |
 | `FORGIT_PREVIEW_CONTEXT`    | lines of diff context in preview mode     | 3                                             |
 | `FORGIT_FULLSCREEN_CONTEXT` | lines of diff context in full-screen mode | 10                                            |
 | `FORGIT_DIR_VIEW`           | command used to preview directories       | `tree` if available, otherwise `find`         |
```
carlfriedrich, line 393: I think the rest of the table should be widened as well to keep the formatting consistent.

bin/git-forgit
```diff
@@ -65,6 +65,15 @@ _forgit_emojify() {
     fi
 }
 
+_forgit_worktree_filter() {
+    if [[ "${FORGIT_CHECKOUT_BRANCH_EXCLUDE_WORKTREES:-true}" == "true" ]]; then
```
carlfriedrich, line 69: Ah, one more thing: we're evaluating this kind of env variables at a central place in the code (search for `FORGIT_LOG_GRAPH_ENABLE` for example). I think we should do it like that with this variable, too.

bin/git-forgit
```diff
@@ -65,6 +65,15 @@ _forgit_emojify() {
...
+        # Worktree branches are prefixed with '+' in 'git branch' output
+        grep -v '^[+]'
```
sandr01d, line 71: From my understanding, the bracket expression isn't necessary or is there any reason we can't simply use:

```suggestion
        grep -v '^+'
```

bin/git-forgit
sandr01d, line 69: Is there actually a case where somebody would not want to exclude worktrees? If they don't work with `checkout` we might want to always exclude them...

sandr01d (commented):
Looks clean, just two minor questions/comments from my side.

Chemaclass:
As @carlfriedrich says, kudos for adding the unit test! I love seen [bashunit](https://bashunit.typeddevs.com/) being use and getting adoption! It feels so good helping the community. Cheers!

PR #489, opened 2026-02-02
Add interactive git restore command

### LINKED ISSUE #447: Add support for git restore

jetm:
### Summary

Add an interactive `git restore` command (`grs`) that provides an fzf-based
selector for discarding unstaged changes to working tree files. Closes #447.

### Problem

There is no interactive way to selectively discard modifications to working
tree files. Users must either manually type file paths with `git restore` or
use `git checkout`, which is the older discouraged syntax.

### Changes

- Add `_forgit_restore` function with fzf-based file selector and diff preview
- Expose as `grs` alias, consistent with existing forgit commands
- Integrate with shell completion systems across bash/zsh/fish
- Add comprehensive test coverage

## Check list

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Test
- [ ] Documentation change

## Test environment

- Shell
    - [ ] bash
    - [x] zsh
    - [ ] fish
- OS
    - [x] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:


bin/git-forgit
```diff
@@ -522,6 +522,52 @@ _forgit_reset_head() {
...
+    _forgit_discard_git_opts=()
+    _forgit_parse_array _forgit_discard_git_opts "$FORGIT_DISCARD_GIT_OPTS"
+    # Unstage the file first
+    git restore -S "${_forgit_discard_git_opts[@]}" -- "$@"
+    # Only restore tracked files (new files become untracked after unstaging)
+    for file in "$@"; do
+        if _forgit_is_file_tracked "$file"; then
+            git restore "${_forgit_discard_git_opts[@]}" -- "$file"
+        fi
+    done
```
wfxr: I'm wondering if we should limit this command to just unstaging rather than combining unstage + discard?

Sometimes we accidentally stage a file and simply want to move it back to the unstaged state — without discarding the actual changes. If someone truly needs to discard modifications, the existing forgit_checkout_file (gcf) already handles that use case well.

What do you think about keeping the two operations separate?

sandr01d: I agree that unstaging and discarding should be two separate options. One of the main concepts of git is the separation between file states (tracked, untracked, staged) and I don't think we should break with this concept.
However, we already have a command for unstaging too (`_forgit_reset_head`/`grh`) and there are plans to add a new command for `git restore` (#447). So is introducing yet another unstage command necessary? Maybe this PR could be refactored to solve #447 instead.

wfxr: Thanks for pointing this out, @sandr01d. I have to admit I had been overlooking the existing grh command, which already covers the unstaging use case quite well.

@sandr01d's suggestion sounds good to me: it would be more valuable to pivot this PR toward implementing #447 (interactive git restore) instead. @jetm What do you think? Would you be interested in refactoring this PR to tackle #447?

jetm: I have added restore to do #447 and kept discard as I still found it useful.

sandr01d: Gotta say, I'm still against the discard command because of the reasons I outlined above. @wfxr WDYT?

wfxr: > Gotta say, I'm still against the discard command because of the reasons I outlined above. @wfxr WDYT?

My opinion is the same as yours.

sandr01d: Please don't just close our comments, this has not been addressed. You'll have to remove the discard, otherwise there won't be a chance of getting this merged.

completions/git-forgit.fish
```diff
@@ -7,8 +7,7 @@
 
 function __fish_forgit_needs_subcommand
     for subcmd in add blame branch_delete checkout_branch checkout_commit checkout_file checkout_tag \
-        cherry_pick cherry_pick_from_branch clean diff fixup ignore log reflog rebase reset_head \
-        revert_commit reword squash stash_show stash_push switch_branch
```
sandr01d: There have been some commands removed here, probably by mistake.

README.md
```diff
@@ -56,6 +56,9 @@ omf install https://github.com/wfxr/forgit
 # for zinit
 zinit load wfxr/forgit
 
+# for antidote (in .zsh_plugins.txt)
+wfxr/forgit kind:defer
+
```
sandr01d, line 61: This change does not belong in this PR. Feel free to open a separate one.

README.md
```diff
@@ -144,6 +147,10 @@ If you're having issues after updating, and commands such as `forgit::add` or al
 
 - **Interactive `git reset HEAD <file>` selector** (`grh`)
 
+- **Interactive `git restore <file>` selector** (`grs`)
+
+
+
```
sandr01d, line 153: The empty lines should not be here.

tests/restore.test.sh
```diff
@@ -0,0 +1,67 @@
...
+    assert_same "untracked content" "$(cat untracked.txt)"
+}
+
+function test_restore_only_reverts_unstaged_changes_when_file_has_staged_and_unstaged() {
+    echo "committed" >mixed.txt
+    git add mixed.txt
+    git commit -q -m "Add mixed file"
+    echo "staged change" >mixed.txt
+    git add mixed.txt
+    echo "unstaged change" >mixed.txt
```
sandr01d, line 44: You need double `>>` here to not overwrite the file.

```suggestion
    echo "staged change" >> mixed.txt
    git add mixed.txt
    echo "unstaged change" >> mixed.txt
```

tests/restore.test.sh
```diff
@@ -0,0 +1,67 @@
...
+function test_restore_with_renamed_file() {
+    echo "rename content" >before-rename.txt
+    git add before-rename.txt
+    git commit -q -m "Add file for rename test"
+    git mv before-rename.txt after-rename.txt
+    git commit -q -m "Rename file"
+    echo "modified after rename" >after-rename.txt
+    _forgit_git_restore after-rename.txt
+    assert_same "rename content" "$(cat after-rename.txt)"
+}
```
sandr01d, line 58: Because you commit the file after renaming, the renaming does not have any effect anymore in this case.

tests/restore.test.sh
```diff
@@ -0,0 +1,67 @@
...
+
+function test_restore_passes_through_arguments_when_non_flags_provided() {
+    echo "original" >passthrough.txt
+    git add passthrough.txt
+    git commit -q -m "Add passthrough file"
+    echo "modified" >passthrough.txt
+    _forgit_restore passthrough.txt
+    assert_same "original" "$(cat passthrough.txt)"
+}
```
sandr01d, line 67: This test has nothing to do with what it's name implies.

completions/_git-forgit
```diff
@@ -97,6 +98,7 @@ _git-forgit() {
     reflog) _git-forgit-reflog ;;
     rebase) _git-rebase ;;
     reset_head) _git-staged ;;
+    restore) _git-checkout-file ;;
```
sandr01d, line 101: Git has  completions for git restore (`_git-restore` for zsh and `_git_restore` for bash). Any reason not to use those?

bin/git-forgit
```diff
@@ -522,6 +522,39 @@ _forgit_reset_head() {
...
+}
+
+_forgit_git_restore() {
+    _forgit_restore_git_opts=()
+    _forgit_parse_array _forgit_restore_git_opts "$FORGIT_RESTORE_GIT_OPTS"
+    git restore "${_forgit_restore_git_opts[@]}" "$@"
+}
+
+# git restore selector
+_forgit_restore() {
```
sandr01d, line 536: One very neat thing about `git restore` IMO is that it allows to not only discard modified files, but also allows unstaging files with `--staged` and discarding from the working tree and the index (staged files) with `--staged --worktree`. I think it would be worth adding  support for this, which would require us to do a bit of argument parsing to determine which files to show in the selection:

* no `--staged` -> modified
* `--staged` -> staged files
* `--staged --worktree` -> modified and staged files

Would you be willing to implement this with this PR @jetm?

sandr01d (changes requested):
Thanks for the update @jetm! Looks mostly good, but needs a few changes here and there. The pipeline is failing because we require the usage of [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/). I'll do a full functional review, once the code is finalized.

tests/restore.test.sh
```diff
@@ -0,0 +1,67 @@
...
+    source bin/git-forgit
+    export GIT_CONFIG_SYSTEM=/dev/null
+    export GIT_CONFIG_GLOBAL=/dev/null
+    cd "$(bashunit::temp_dir)" || return 1
+    git init -q
+    git config user.email "test@example.com"
+    git config user.name "Test User"
+    echo "initial" >README.md
+    git add README.md
+    git commit -q -m "Initial commit"
```
sandr01d, line 13: This file is added and committed here, but not actually used in any of the tests, so this section should be removed.

tests/restore.test.sh
```diff
@@ -0,0 +1,67 @@
...
+}
+
+function test_restore_shows_message_when_no_modified_files() {
+    output=$(_forgit_restore 2>&1)
+    assert_general_error
+    assert_same "Nothing to restore." "$output"
+}
+
+function test_restore_reverts_modified_file_to_committed_state() {
+    echo "original content" >tracked.txt
```
sandr01d, line 23: Very minor nitpick, but could you please add an extra space after the redirects? This applies to all redirects in this file.

```suggestion
    echo "original content" > tracked.txt
```

sandr01d:
Hey @jetm, do you still plan to work on this PR? If not, I'd offer to take it over and push it over the finish line.

jetm:
> Hey @jetm, do you still plan to work on this PR? If not, I'd offer to take it over and push it over the finish line.

@sandr01d, please, go ahead. Thanks!

sandr01d:
#537 has been merged, so I'm closing this one. Thanks again @jetm.

PR #513, opened 2026-04-09
ci: do not install recommended packages when using apt

sandr01d:
## Check list

## Description

Not installing recommended packages should speed up the CI. The speedup is probably very small, but because there is no downside to this I think we should take it.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Test
- [ ] Documentation change
- [X] CI

## Summary by CodeRabbit

* **Chores**
  * Optimized CI/CD pipeline by reducing unnecessary package dependencies, improving build efficiency.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -37,7 +37,7 @@ jobs:
         elif [[ "${{ matrix.os }}" == ubuntu* ]]; then
           sudo apt-add-repository -y ppa:fish-shell/release-3
           sudo apt -y update
-          sudo apt -y install zsh fish shellcheck
+          sudo apt -y install --no-install-recommends zsh fish shellcheck
         fi
         curl -s https://bashunit.typeddevs.com/install.sh | bash -s 0.31.0
 


carlfriedrich (commented):
@sandr01d Good idea!

I think we could even improve speed more if we run the workflow [in a prebuilt container](https://docs.github.com/de/actions/how-tos/write-workflows/choose-where-workflows-run/run-jobs-in-a-container). This would require building and storing the container image somewhere, though, which might make the build process a little less transparent. WDYT?

sandr01d:
> I think we could even improve speed more if we run the workflow [in a prebuilt container](https://docs.github.com/de/actions/how-tos/write-workflows/choose-where-workflows-run/run-jobs-in-a-container). This would require building and storing the container image somewhere, though, which might make the build process a little less transparent. WDYT?

That's definitely interesting. The question to me is if the extra work having to maintain the containers is already worth it yet. After all our pipeline is not **that** slow at the moment. Maybe just [caching](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching) the dependencies is enough?

wfxr (approved):

carlfriedrich:
@sandr01d You're right, it's probably not worth it. So we should apply YAGNI here and just merge this.

carlfriedrich (approved):

PR #543, opened 2026-07-07
perf(fish): migrate FORGIT_ vars without spawning subprocesses

AllanZyne:
## What

The backwards-compatibility block in `conf.d/forgit.plugin.fish` that exports unexported `FORGIT_` config variables used a shell pipeline:

```fish
set | awk -F ' ' '{ print $1 }' | grep FORGIT_ | while read var
    if not set -x | grep -q "^$var\b"
```

This forks `awk` and `grep` (once for the list, then a `grep` per variable) on **every interactive fish startup**. I replaced it with fish's built-in `set -n` / `set -xn` + `string match`, which run in-process:

```fish
for var in (set -n | string match -er '^FORGIT_')
    if not set -xn | string match -qr "^$var\$"
```

## Why

Pure startup-speed win — no external processes spawned for this block anymore. On my machine the block's cost dropped from ~2.5ms to ~1.5ms; the real benefit is eliminating the per-variable subprocess fork, which scales with how many `FORGIT_` vars a user has set.

## Behavior

Unchanged. Verified:
- Unexported `FORGIT_` variables are still detected, warned about, and exported.
- Nothing is emitted when all `FORGIT_` variables are already exported.

The anchored `^FORGIT_` match on `set -n` output (one name per line) is equivalent to the previous `grep FORGIT_`, just without the substring ambiguity.

## Testing

- `fish conf.d/forgit.plugin.fish` — parses clean
- Manual equivalence tests for both the "all exported" (silent) and "unexported var" (warns + migrates) paths
- OS: Linux; shell: fish 4.6.0

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of user-defined `FORGIT_` environment settings so they are exported more reliably.
  * Streamlined the check for already-exported settings, which should make startup behavior more consistent.

### DIFF
--- a/conf.d/forgit.plugin.fish
+++ b/conf.d/forgit.plugin.fish
@@ -15,8 +15,8 @@ end
 # backwards compatibility:
 # export all user-defined FORGIT variables to make them available in git-forgit
 set unexported_vars 0
-set | awk -F ' ' '{ print $1 }' | grep FORGIT_ | while read var
-    if not set -x | grep -q "^$var\b"
+for var in (set -n | string match -er '^FORGIT_')
+    if not set -xn | string match -qr "^$var\$"
         if test $unexported_vars = 0
             forgit::warn "Config options have to be exported in future versions of forgit."
             forgit::warn "Please update your config accordingly:"


sandr01d (approved):
Nice improvement, thanks for your contribution! I'll leave the PR open over the weekend so the other maintainers have a chance to take a look and will merge it next week should there be no objections.

PR #544, opened 2026-07-09
perf: speed up modified file lookup in large repositories

cjappl:
## Check list

## Description

Use a modified-file-only path for `git-forgit restore` (`grs`) and `git-forgit checkout_file` (`gcf`) so these interactive selectors do not scan large sets of untracked files. This keeps both commands responsive in repositories with many untracked files.

On a repo I use for work, comparing the raw `git diff` with the raw `git ls-files` for listing modified files is ~8x faster, the difference between ~1.5s and ~10s. The quality of life improvement is huge. 

To get this change working when in sibling directories, I pulled out an existing `perl` script used in one of the other functions. 

## Type of change

- [ ] Performance improvement
- [x] Refactor

## Test environment

- Shell
    - [x] bash
    - [x] zsh
    - [x] fish
- OS
    - [ ] Linux
    - [x] Mac OS X
    - [x] Windows
    - [ ] Others:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved modified-file listings when running from subdirectories, with correct sibling-directory relative display paths.
  * Updated restore and file checkout to use the new modified-file selection logic for more accurate results.
  * Ensured untracked files are excluded from “modified-only” outputs.
  * Standardized status row formatting and path display via shared cwd-relative normalization.

* **Tests**
  * Added coverage for modified-file parity, cwd-relative path rewriting, status-entry formatting, and untracked-file filtering behavior.

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -222,6 +222,61 @@ _forgit_list_files() {
     git ls-files -z "$@" "$rootdir" | tr '\0' '\n' | uniq
 }
 
+_forgit_rewrite_repo_paths_for_cwd() {
+    local rootdir mode separator cwd
+    rootdir=$1
+    mode=${2:-paths}
+    separator=${3:-}
+    cwd=$(pwd -P)
+
+    perl -MCwd=realpath -MFile::Spec -e '
+        use strict;
+        use warnings;
+
+        my ($rootdir, $cwd, $mode, $separator) = @ARGV;
+        my $normalized_rootdir = realpath($rootdir);
+        my $normalized_cwd = realpath($cwd);
+
+        while (my $line = <STDIN>) {
+            chomp $line;
+
+            my ($status, $repo_path);
+            if ($mode eq q{status_entries}) {
+                next unless $line =~ /^(..[^[:space:]]*)( )(.*)$/;
+                ($status, $repo_path) = ($1, $3);
+            } else {
+                next if $line eq q{};
+                $repo_path = $line;
+            }
+
+            my $absolute_path = "$normalized_rootdir/$repo_path";
+            my $display_path = File::Spec->abs2rel(realpath($absolute_path) // $absolute_path, $normalized_cwd);
+            $display_path = q{.} if $display_path eq q{};
+
+            if ($mode eq q{status_entries}) {
+                print "[$status]  ${display_path}${separator}${normalized_rootdir}/${repo_path}\n";
+            } else {
+                print "$display_path\n";
+            }
+        }
+    ' "$rootdir" "$cwd" "$mode" "$separator"
+}
+
+_forgit_list_modified_files() {
+    # on large repos with many untracked files, git ls-files
+    # is substantially slower than git diff
+    # Wherever possible (when we don't care about untracked files)
+    # use this command over _forgit_list_files
+
+    # See note on _forgit_list_files() about -z and other options
+    local rootdir
+    rootdir=$(git rev-parse --show-toplevel)
+    git -C "$rootdir" diff --name-only -z |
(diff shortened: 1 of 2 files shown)


bin/git-forgit
```diff
@@ -222,6 +222,61 @@ _forgit_list_files() {
...
+            $display_path = q{.} if $display_path eq q{};
+
+            if ($mode eq q{status_entries}) {
+                print "[$status]  ${display_path}${separator}${normalized_rootdir}/${repo_path}\n";
+            } else {
+                print "$display_path\n";
+            }
+        }
+    ' "$rootdir" "$cwd" "$mode" "$separator"
+}
```
cjappl, line 263: This change was largely just pulled from `_forgit_build_status_entries` below, with a few ergonomic tweaks to make it work for both situations. The new tests should hopefully make sure this is stable

cjappl:
Also, hi! Long time no see. I see from my inbox you all have been super hard at work in this repo.

Please let me know if I've run afoul of any new norms or if I've missed anything.

cjappl:
I like the coderabbit nitpick review, I think that's also a good improvement. I can either apply it here or do it in a follow up commit. Let me know what you think (human reviewers)

bin/git-forgit
```diff
@@ -222,6 +222,61 @@ _forgit_list_files() {
     git ls-files -z "$@" "$rootdir" | tr '\0' '\n' | uniq
 }
 
+_forgit_rewrite_repo_paths_for_cwd() {
+    local cwd mode rootdir separator
+    rootdir=$1
+    mode=${2:-paths}
+    separator=${3-}
+    cwd=$(pwd -P)
```
sandr01d: Nit: I'd put the definition and assignment of the local variables in the same order. I find it easier to read this way.

```suggestion
    local rootdir mode separator cwd
    rootdir=$1
    mode=${2:-paths}
    separator=${3-}
    cwd=$(pwd -P)
```

sandr01d (approved):
Good to see you here again!
I like this very much. I've run into `git ls-files` being slow on large repositories in the past too and this should give us a nice boost. Approved with two minor nitpicks.

sandr01d:
> I like the coderabbit nitpick review, I think that's also a good improvement. I can either apply it here or do it in a follow up commit. Let me know what you think (human reviewers)

I agree. As to whether it should go in a follow up commit or not, I'm fine with both. Maybe leaning a bit more towards having a separate commit because it is a different kind of optimization, but fine either way.

bin/git-forgit
```diff
@@ -222,6 +222,61 @@ _forgit_list_files() {
     git ls-files -z "$@" "$rootdir" | tr '\0' '\n' | uniq
 }
 
+_forgit_rewrite_repo_paths_for_cwd() {
+    local cwd mode rootdir separator
+    rootdir=$1
+    mode=${2:-paths}
+    separator=${3-}
```
sandr01d: I think this is the more common version to do this because it also expands to an empty string if the variable is declared but is null. At least that's what I found when looking it up [here](https://tldp.org/LDP/abs/html/parameter-substitution.html).

```suggestion
    separator=${3:-}
```

cjappl:
> > I like the coderabbit nitpick review, I think that's also a good improvement. I can either apply it here or do it in a follow up commit. Let me know what you think (human reviewers)
> 
> I agree. As to whether it should go in a follow up commit or not, I'm fine with both. Maybe leaning a bit more towards having a separate commit because it is a different kind of optimization, but fine either way.

Cool, I lean to doing a follow on, so I will do that! 

Thanks for the review, applied the two comment suggestions.

PR #545, opened 2026-07-11
ci: update bashunit GitHub action using dependabot

sandr01d:
## Description

Now that we use the bashunit GitHub Action and get updates with dependabot, we can always assume to
be on the latest version. This way we don't have to update the docs constantly.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Test
- [X] Documentation change
- [x] CI change

## Summary by CodeRabbit

* **Documentation**
  * Updated development setup instructions to use the testing utility’s upstream installer without pinning a specific version.

* **Chores**
  * Updated continuous integration to invoke the testing utility at version 0.40.0 with checksum verification enabled.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,9 +27,8 @@ jobs:
     steps:
     # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
     - uses: actions/checkout@v7
-    - uses: TypedDevs/bashunit@v0
+    - uses: TypedDevs/bashunit@0.40.0
       with:
-          version: '0.40.0'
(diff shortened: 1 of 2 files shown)


sandr01d:
I noticed that dependabot actually not update the bashunit GitHub Action, so I'll draft this until I fixed that.

carlfriedrich (approved):

sandr01d:
FYI: I kept the old 0.40.0 of bashunit version on purpose. This way we should have dependabot create a PR right away and we can verify everything is working as intended.

wfxr (approved):

PR #546, opened 2026-07-13
perf: avoid listing modified files twice in checkout-file

cjappl:
_forgit_checkout_file called _forgit_list_modified_files once to check for an empty result and again to feed fzf, running git diff twice on every invocation. Cache the result and reuse it for both checks.

## Check list

## Type of change

- [x] Performance

## Test environment

- Shell
    - [ ] bash
    - [x] zsh
    - [x] fish
- OS
    - [ ] Linux
    - [x] Mac OS X
    - [ ] Windows
    - [ ] Others:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the modified-file checkout picker by reusing a single, consistent list of modified files for both the empty-state check and the selection display.
  * Prevented cases where the candidates shown in the picker could differ from the availability logic used to decide whether files were selectable.

### DIFF
--- a/bin/git-forgit
+++ b/bin/git-forgit
@@ -1123,12 +1123,13 @@ _forgit_git_checkout_file() {
 # git checkout-file selector
 _forgit_checkout_file() {
     _forgit_inside_work_tree || return 1
-    local files opts
+    local files opts modified_files
     _forgit_contains_non_flags "$@" && {
         _forgit_git_checkout_file "$@"
         return $?
     }
-    [[ $(_forgit_list_modified_files | wc -l) -eq 0 ]] && echo 'Nothing to checkout.' && return 1
+    modified_files="$(_forgit_list_modified_files)"
+    [[ -z $modified_files ]] && echo 'Nothing to checkout.' && return 1
     opts="
         $FORGIT_FZF_DEFAULT_OPTS
         -m -0
@@ -1138,7 +1139,7 @@ _forgit_checkout_file() {
     files=()
     while IFS='' read -r file; do
         files+=("$file")
-    done < <(_forgit_list_modified_files |
+    done < <(printf '%s\n' "$modified_files" |
         FZF_DEFAULT_OPTS="$opts" fzf)
     [[ ${#files[@]} -gt 0 ]] && _forgit_git_checkout_file "$@" "${files[@]}"
 }


cjappl:
This originated from a coderabbit review here:
https://github.com/wfxr/forgit/pull/544#pullrequestreview-4667305488

bin/git-forgit
```diff
@@ -1123,12 +1123,13 @@ _forgit_git_checkout_file() {
...
     _forgit_inside_work_tree || return 1
-    local files opts
+    local files opts modified_files
     _forgit_contains_non_flags "$@" && {
         _forgit_git_checkout_file "$@"
         return $?
     }
-    [[ $(_forgit_list_modified_files | wc -l) -eq 0 ]] && echo 'Nothing to checkout.' && return 1
+    modified_files="$(_forgit_list_modified_files)"
+    [[ -z $modified_files ]] && echo 'Nothing to checkout.' && return 1
```
sandr01d: I'd treat this as an array here because it makes the intent clearer IMO and we do so in other places.

```suggestion
    [[ ${#modified_files[@]} -eq 0 ]] && echo 'Nothing to checkout.' && return 1
```

sandr01d (changes requested):
Looks good, just one comment.

bin/git-forgit
cjappl: Ok, this version fails a unit test. You can reproduce this by running bashunit, and we get:

```

There was 1 failure:

|1) tests/empty-state.test.sh:21
|✗ Failed: Checkout file shows message when no modified files
|    Expected 'Nothing to checkout.'
|    but got  ''
|    at tests/empty-state.test.sh:21

Tests:      72 passed, 1 failed, 73 total
Assertions: 120 passed, 1 failed, 121 total

 Some tests failed
Time taken: 9.22s
```


Copilot explanation (as I'm a little out of my depth on the nuances of bash syntax):
```
modified_files="$(_forgit_list_modified_files)"
[[ ${#modified_files[@]} -eq 0 ]]

 modified_files  is a string, not an array. In Bash,  ${#modified_files[@]}  evaluates to  1  even when that string is empty, so the empty-state branch never prints  Nothing to checkout. . The function then passes an empty line into  fzf , which exits without producing output.

```

OK if  I revert to my original version? Or do you have something else you'd like me to try?

sandr01d: Ah sorry, my bad. I somehow had it in my head that it's an array but it isn't. In this case reverting back to the original version is correct.

cjappl: Thanks for the help on this. Reverted!

Let me know if you see anything else.

sandr01d (approved):
LGTM

carlfriedrich (approved):

PR #556, opened 2026-09-08
fix(fish): completions fail to load with fish >= 4.8.0

### LINKED ISSUE #548: Plugin gives error with fish >=4.8.0
With the release of fish 4.8.0 this plugin is giving error when pressing `<tab>` to auto-complete
```sh
git-forgit add source: Error encountered while sourcing file '/usr/share/fish/completions/git.fish':
source: No such file or directory
```
According to changelog https://github.com/fish-shell/fish-shell/releases#release-4.8.0

> With the exception of the $CMAKE_INSTALL_PREFIX/share/fish/man directory, fish no longer installs files to $CMAKE_INSTALL_PREFIX/share/fish. In particular, this means that both $CMAKE_INSTALL_PREFIX/share/fish/completions and $CMAKE_INSTALL_PREFIX/share/fish/functions should no longer exist. These directories have been ignored since fish 4.2. If another package installs fish scripts there, they should be corrected to install to extra_completionsdir (typically $CMAKE_INSTALL_PREFIX/share/fish/vendor_completions.d), extra_functionsdir (typically $CMAKE_INSTALL_PREFIX/share/fish/vendor_functions.d) or extra_confdir (typically $CMAKE_INSTALL_PREFIX/share/fish/vendor_functions.d) instead.

It seems that the asset files that were present in those directories are part of the binary now and those dirs are ignored.

I also looked through the issues (because I'm not familiar with fish codebase) and found this https://github.com/fish-shell/fish-shell/issues/12769 and I tried one solution mentioned there, namely
```fish
# not functions -q __fish_git && source $__fish_data_dir/completions/git.fish
# WARN: see https://github.com/fish-shell/fish-shell/issues/12769

not functions -q __fish_git && status get-file completions/git.fish | source
```
and seems to be working. Not confident if that is the best solution, so reluctant to create a PR myself.

sandr01d:
## Check list

## Description

Fish started shipping git completions in the binary with 4.2.0 and deprecated the old completion files with 4.8.0.

Update forgit’s fish completion loader to support both models:

* prefer `status get-file completions/git.fish` (binary-embedded completions)
* fall back to `$__fish_data_dir/completions/git.fish` (completion files)

Also refactors loading into a helper function to allow early returns and reduce nesting.

closes #548

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Test
- [ ] Documentation change
- [ ] CI change

## Test environment

- Shell
    - [ ] bash
    - [ ] zsh
    - [X] fish
- OS
    - [X] Linux
    - [ ] Mac OS X
    - [ ] Windows
    - [ ] Others:

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Git completion loading across supported Fish versions.
  - Added a fallback for environments where Git completions are not bundled with Fish.
  - Prevents further completion setup when Git completions cannot be loaded, avoiding incomplete or misleading command suggestions.
  - Preserves existing Git completion behavior when completions are already available.

### DIFF
--- a/completions/git-forgit.fish
+++ b/completions/git-forgit.fish
@@ -20,8 +20,29 @@ function __fish_forgit_worktrees
     git worktree list --porcelain 2>/dev/null | string match -r '^worktree .+' | string replace 'worktree ' ''
 end
 
-# Load helper functions in git completion file
-not functions -q __fish_git && source $__fish_data_dir/completions/git.fish
+# Load git completion functions
+function __fish_forgit_load_git_completions
+    if functions -q __fish_git
+        return 0
+    end
+
+    # Fish >= 4.2.0 ships git completions in its binary
+    if set --local git_completions (status get-file completions/git.fish)
+        printf '%s\n' $git_completions | source
+        return 0
+    end
+
+    # Older fish versions ship completions in a file
+    set --local git_completions_file "$__fish_data_dir/completions/git.fish"
+    if test -f "$git_completions_file"
+        source "$git_completions_file"
+        return 0
+    end
+
+    return 1
+end
+
+__fish_forgit_load_git_completions || exit
 
 # No file completion by default
 complete -c git-forgit -x


bingocaller:
Thanks for picking this up, @sandr01d! 🫶

Changes **LGTM** 👍 but I don’t have the power to approve it, unfortunately. 😅

wfxr (approved):

